Issue #32, opened 2017-03-11
[Feature Request] Add romaji input to search


dannyperson:
Searching dictionaries is very useful, but less useful since switching keyboard layouts is required.  It would be nice if it was not necessary.

![screencapture](https://cloud.githubusercontent.com/assets/1476684/23820618/2bfa9af0-05e2-11e7-81e6-8773243e2083.png)

dannyperson:
Closed by #33

PR #33, opened 2017-03-11
Add romaji input to search

dannyperson:
Adds #32 by using wanakana IME on the search query field


dannyperson:
Fix weird encoding issue with wanakana

FooSoft:
Neat! Merged into `dev`, thanks for the code.

PR #65, opened 2017-07-16
Add glossary-brief anki field, like glossary but without tags

### LINKED ISSUE #64: {glossary} without tags?
Hey, thanks for yomichan. Is there any tag to have the definitions for a word without adding the tags like "(P, ichi, news, v1, vt, JMDICT (English))" before it?

My anki card template has a huge font so it's a bit uncomfortable to have both in the same field.

screen

If there's no such thing, I'd like to try implementing it myself (sounds reasonably easy), and would appreciate some directions on how you would prefer to have it done.

dequis:
Closes #64 

Pretty much what you said, just needed to change `ext/bg/js/util.js` too.

I'm not entirely convinced about having `{glossary}` with tags as the main one for terms, when `{glossary}` for kanji doesn't have tags. It's a thing about internal consistency vs backwards compatibility. I think I value backwards compatibility more than internal consistency, so this is fine. But it's not like much would be lost with that behavior change, either. Up to you.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,7 @@ Flashcard fields can be configured with the following steps:
     `{expression}` | Term expressed as Kanji (will be displayed in Kana if Kanji is not available).
     `{furigana}` | Term expressed as Kanji with Furigana displayed above it (e.g. <ruby>日本語<rt>にほんご</rt></ruby>).
     `{glossary}` | List of definitions for the term (output format depends on whether running in *grouped* mode).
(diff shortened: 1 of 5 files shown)


FooSoft:
Good work. I think for the time being, backwards compatibility wins. Going forward, there is an options versioning system in Yomichan, so we can rewrite user settings doing things like remapping tags, etc.

PR #66, opened 2017-07-27
Import terms with transactional adds instead of bulkAdd

lae:
I think I was running into some race or performance issue when attempting to import jmnedict.zip into Yomichan on Firefox 56 - having reliably run into the following failure more than 20 times (including on clean profiles):

dfo6xwyu0aarqix

(where it fails is sometimes the same but usually different)

From what I could see in the Browser Console, `UnknownErr`s were being thrown from the IndexedDB library in Firefox.

With this change, this no longer occurs for me.

I tried doing this on the dev branch with all the async changes but while the database imports worked, I'm not seeing any dropdown when I actually go to hover above words (they get highlighted but no dictionary entries show up).

### DIFF
--- a/ext/bg/js/database.js
+++ b/ext/bg/js/database.js
@@ -189,20 +189,19 @@ class Database {
         };
 
         const termsLoaded = (title, entries, total, current) => {
-            const rows = [];
-            for (const [expression, reading, tags, rules, score, ...glossary] of entries) {
-                rows.push({
-                    expression,
-                    reading,
-                    tags,
-                    rules,
-                    score,
-                    glossary,
-                    dictionary: title
-                });
-            }
-
-            return this.db.terms.bulkAdd(rows).then(() => {
+            return this.db.transaction('rw', this.db.terms, function() {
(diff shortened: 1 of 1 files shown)


FooSoft:
Thank you for reporting this problem and preparing a fix. I was not able to reproduce this problem on the nightly version of Firefox 56 (which platform are you running on, by the way?) but IndexedDB is quite fragile from my experience, so I am honestly not too surprised... I vaguely remember `DOMError` popping up before on Firefox (and Firefox only) when the browser invalidates objects that it believes reference DOM elements (or anything that belonged to DOM). I wonder if this could be somehow related...

It's also interesting that this works fine with `async`... The `dev` branch is in the middle of a large refactor so it may be partially broken. 

Anyway, thank you for the code; it looks like a safe way to work around this library and/or browser bug.

lae:
I should probably also mention that I wasn't able to reproduce this on nightly on my laptop at all - which is what led me to assume it's a performance related issue (though I did also try putting my Firefox profile on a ramdisk and still ran into the issue...).

Thanks!

PR #70, opened 2017-08-08
Allow glossary lists to have centered alignment

### LINKED ISSUE #69: Anki cards with text-align: left in {glossary} look weird
Hi.

Text in anki cards is centered by default. For some reason glossary was sent to the left. I think it's easier to read the flashcards if they are centered. Is it okay to just change this behavior?

A card aligned to the left

screen1

Removing the alignment:

screen2

Adding `list-style-position: inside`

screen3

dequis:
"list-style-position: inside" means the number will be together with the centered text instead of moved to the left

----

Fixes #69 

I figured this was simple enough I could just do it directly. Feel free to close if you disagree with this change or if you'd prefer to have it behind a setting.

### DIFF
--- a/ext/bg/js/templates.js
+++ b/ext/bg/js/templates.js
@@ -169,7 +169,7 @@ templates['fields.html'] = template({"1":function(container,depth0,helpers,parti
 },"46":function(container,depth0,helpers,partials,data) {
     var stack1;
 
-  return "<ol>"
+  return "<ol style=\"padding: 0\">"
     + ((stack1 = helpers.each.call(depth0 != null ? depth0 : (container.nullContext || {}),((stack1 = (depth0 != null ? depth0.definition : depth0)) != null ? stack1.glossary : stack1),{"name":"each","hash":{},"fn":container.program(47, data, 0),"inverse":container.noop,"data":data})) != null ? stack1 : "")
     + "</ol>";
 },"47":function(container,depth0,helpers,partials,data) {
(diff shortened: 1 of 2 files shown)


dequis:
Welp, even a simple change like this one can have issues, I guess. Not because of the change but because of the rebase I did before pushing. The dev branch is broken right now, first error is missing `/bg/js/util.js` in popup.html, but that doesn't fix it completely (dead object errors)

dequis:
>Welp, even a simple change like this one can have issues

Turns out it was off-center because of the padding of `<ol>`

FooSoft:
There are a couple of concerns I have with this change:

* Single entry, long definitions (like the ones that come out of EPWING) objectively make sense to left-justify otherwise you are reading paragraphs of centered text which is kind of gross. Having the text left-justified in some cases and center-aligned in others creates undesired inconsistency in cards.
* People generally don't like when styling is changed on them because it makes existing cards look different from new ones produced by Yomichan.
* Putting things behind option flags is definitely a possibility but it will not scale when you think about all of the different parameters that can be tweaked about card creation.

I think the only real solution is to allow people to edit the templates used for card generation within the options dialog. This way everyone wins and I don't have to maintain lots of branching and options management for making things look the way people want.

Regarding dev, it is still a bit of a mess but it's getting closer. FWIW the new file organization and switch to `async` makes things a lot more readable and easy to work with :+1:

dequis:
Huh looks like I forgot to reply here.

>I think the only real solution is to allow people to edit the templates used for card generation within the options dialog. This way everyone wins and I don't have to maintain lots of branching and options management for making things look the way people want.

Yeah, I agree. I'll close this and work on the correct solution later

PR #73, opened 2017-08-20
Change the {furigana} marker for anki export to distribute furigana over kanji

kmltml:
Requires a bit more testing, but seems to work reasonably well


kmltml:
Rebased (manually)

FooSoft:
There are some style differences with the rest of the code I'll fix up later, but this looks very useful. I merged this into the `dev` branch and it will be included in the next release.

Issue #76, opened 2017-09-11
[bug] Cannot load any dictionaries


Piranhaplant:
Upon a new install of the extension, this message is shown on the settings page under "Dictionaries":
> **Error:** OpenFailedError: UnknownError The operation failed for reasons unrelated to the database itself and not covered by any other error code.

Purging the database doesn't do anything and no dictionaries can be imported.

Using version 1.3.5 in Firefox Nightly 57.0a1 on Windows 7.

FooSoft:
Some versions of Firefox appear to have a buggy implementation of IndexedDb ... will investigate. Thank you for the report.

FooSoft:
@Piranhaplant so it looks like I'm unable to reproduce the problem in the 64-bit build of 57.0a1 nightly (2017-09-11). Could you check to see that your nightly is up to date, and if it isn't, what is the date on it. 

Also, did you get the error before you even attempted to import any dictionaries?

Piranhaplant:
@FooSoft I am using the latest 64-bit nightly (2017-09-11). The error shows up on a clean install of the extension before importing any dictionaries.

If you don't have any other ideas, I will try a fresh profile/install of Firefox and see if that helps.

FooSoft:
@Piranhaplant so it appears then that you must be getting the error on the initial database open (before Yomichan even does anything interesting with it). I've seen a similar issue in Firefox 53. Some relevant threads about people running into the same problem:

* https://github.com/dfahlander/Dexie.js/issues/298
* https://bugzilla.mozilla.org/show_bug.cgi?id=1217257 (do you have other installs of Firefox?)
* https://bugzilla.mozilla.org/show_bug.cgi?id=1310616
* https://greasyfork.org/en/forum/discussion/6663/state-of-denial

It just seems like the underlying support for IndexedDb is pretty spotty. If you could try wiping your profile and/or reinstalling FF I would be very interested in seeing if the problem goes away.

Piranhaplant:
I tried it on a new profile and that fixed the issue, then I went back to my original profile and used the "Refresh Firefox" feature, which also fixed it. I've restored my settings and other extensions and Yomichan is still working, so I'm not sure what the problem was. Sorry to waste your time!

FooSoft:
[...]

PR #77, opened 2017-09-12
Import terms/kanji with transactional adds instead of bulkAdd on async

lae:
Re-implementation of #66 for async, since I guess behaviour was reverted. Also applied it to kanji import.

(the underlying, I suppose I/O with IndexedDB, issue reappeared for me when I was importing terms after my database disappeared during 1.3.5 upgrade)

I tried to use `await this.db.terms.add()` here but that was leading to an `InvalidStateError`. I'm not sure whether the 'await` is necessary there or not.
For reference: http://dexie.org/docs/Dexie/Dexie.transaction()#async-and-await

### DIFF
--- a/ext/bg/js/database.js
+++ b/ext/bg/js/database.js
@@ -187,20 +187,19 @@ class Database {
                 callback(total, current);
             }
 
-            const rows = [];
-            for (const [expression, reading, tags, rules, score, ...glossary] of entries) {
-                rows.push({
-                    expression,
-                    reading,
-                    tags,
-                    rules,
-                    score,
-                    glossary,
-                    dictionary: title
-                });
-            }
-
-            await this.db.terms.bulkAdd(rows);
+            await this.db.transaction('rw', this.db.terms, async function() {
+                for (const [expression, reading, tags, rules, score, ...glossary] of entries) {
+                    this.db.terms.add({
+                        expression,
+                        reading,
+                        tags,
+                        rules,
+                        score,
+                        glossary,
+                        dictionary: title
+                    });
+                }
+            });
         };
 
         const kanjiLoaded = async (title, entries, total, current)  => {
@@ -208,19 +207,18 @@ class Database {
                 callback(total, current);
             }
 
-            const rows = [];
-            for (const [character, onyomi, kunyomi, tags, ...meanings] of entries) {
-                rows.push({
-                    character,
-                    onyomi,
-                    kunyomi,
-                    tags,
-                    meanings,
-                    dictionary: title
-                });
-            }
-
-            await this.db.kanji.bulkAdd(rows);
+            await this.db.transaction('rw', this.db.kanji, async function() {
+                for (const [character, onyomi, kunyomi, tags, ...meanings] of entries) {
+                    this.db.kanji.add({
+                        character,
+                        onyomi,
+                        kunyomi,
+                        tags,
+                        meanings,
+                        dictionary: title
+                    });
+                }
+            });
         };
 
         await Database.importDictionaryZip(archive, indexLoaded, termsLoaded, kanjiLoaded);


FooSoft:
Looks like this got lost in the merge, thanks for the fix (again). The `await` is indeed necessary to make sure that if `transaction` throws that the exception will not get lost and will properly be propagated upwards.

FooSoft:
@lae looking at the code again, we probably don't even need to have transactions, just doing the `add`s with `await` should suffice (since dictionaries are not added atomically anyway). I was never able to repro the issue you were having but this seems like it would be safe change. I'll make the change locally, hopefully it does not bring back issues for you.

FooSoft:
Actually it looks like removing the `transaction` causes perf to completely tank; looks like it's needed after all.

FooSoft:
@lae doing some more performance testing -- it looks like even with `transaction` we are about 2x as slow (at least on Chrome) for dictionary imports which really sucks. I think I might have an idea as to what the actual cause of the problem with `bulkAdd`. Can you check to see if the changes I've made in the `dev` branch allow you to import without problems?

lae:
Will do once I'm home in a few hours (since I can only reproduce the issue on my desktop).

FooSoft:
It appeared to me to be an issue with Firefox being stupid about failing to keep alive objects that are referenced from the extension scripts. I've ran into similar problems with storing of options; the solution there was creating a deep copy of the objects that I'm operating on. I do this by serializing the object to JSON and then immediately deserialzing it back out.

lae:
I think I figured out exactly how to reproduce the issue while trying to get a control - and it looks like my issue wasn't actually (I think) a performance/I/O issue but an issue regarding available disk space.

I'm able to import the two jmdicts (english vocab and names) fine with the currently released 1.3.4 if I have >4GB of disk space available - while the database itself only takes up 400MB. I run into bulkAdd errors (looking at the browser console I see that it's `Unhandled rejection: AbortError: Transaction aborted`) if I fill up enough space that it falls below ~4GB. 

I'm guessing Firefox limits IndexedDBs to 10% of the available free space it determines on browser startup. I tried leaving >4GB of space available, starting Firefox, then dropping free space to 2GB and imports worked fine.

I think the changes I contributed just "happened" to work fine in #66 by operating below the margin of failure with regards to my disk usage then - because I actually did get a more relevant QuotaExceeded error in the browser console when testing out this PR in my browser, which prompted me to rsync some folders off.

I'll test the dev branch next but, I believe that the 10% quota is the actual issue here - if you want to try to reproduce it yourself (I was using `dd` to create a large empty file).

lae:
Okay, just tested dev. The issue still persists and the same behaviour as in my previous comment is reproducible.

I think just having a FAQ or something in the documentation about quotas/required disk space in Firefox (actually, dunno if this is reproducible in Chrome too) should be enough.

FooSoft:
Wow, that is some very interesting info, I really appreciate you taking some time to dig into it. Is the error you get always the same on as you posted in the screenshot in #66? Probably the most reasonable course of action (in addition to updating the docs) would be to catch the specific exception and rethrow it with a more meaningful description.

I will have to try running Chrome with limited disk space and see how it handles things; it is possible it will have similar issues, but I'm betting not. So far all of the IndexedDB related issues have been exclusive to Firefox's implementation. The issue I was tracking down just yesterday, #76 is also yet another example of where Firefox does not fail gracefully when it comes to IndexedDB operations.

Thanks again for looking into this -- sporadic database failures that happen just for *some* users are pretty terrible.

lae:
Yes, it's the same error as the screenshot you're referring to. And yeah, I happened to pass by that issue yesterday too. I ended up reading a really long bug report related to IndexedDB on Firefox when I opened that first PR, even...

PR #83, opened 2017-09-25
Go to next and previous result with Alt+wheel

siikamiika:
Analogous to Alt+up/down which does the same thing


FooSoft:
Looks good! Also, thank you for following the coding style, you are seriously the first person to make an effort!

Issue #89, opened 2017-10-12
Alert for IndexError pops up due to weird interaction with CSS scale


RodericDay:
This issue may be just too niche and bizarre and self-inflicted to merit fixing, but I have a little [webapp](https://roderic.ca/apps/jgap/) that I usually zoom into for viewing comfort. If the zoom is set inappropriately, the yomichan widget zooms-in too, but this is fine. 

The problem is that, depending on the cursor motion, sometimes an annoying alert pops up. I narrowed it down to the following setup, with Firefox Nightly and OSX, and I can trigger it consistently.

```
<meta charset="utf-8">

<style>
input {
    display:block;
    transform-origin: 0 0;
}
input:nth-child(2) {
    transform: translateX(100px);
}
input:nth-child(3) {
    transform: scale(2);
}
.yomichan-imposter {
    opacity: 0.5 !important;
}
</style>

<input value="犬か猫か猿か馬か亀か豚か鬼"/>
<input value="犬か猫か猿か馬か亀か豚か鬼"/>
<input value="犬か猫か猿か馬か亀か豚か鬼"/>
```

![screen shot 2017-10-12 at 00 24 53](https://user-images.githubusercontent.com/6867226/31479256-e274227e-aee3-11e7-9578-fd0cd911b680.png)

The alert happens when I interact with the lower border of the input while pressing shift, where I can see Yomichan overlaying its own detector box if I open the DOM inspector panel. The left, top, and right borders seem to be fine.

I would love to fix this myself and thus become a contributor to this project. If you want me to spelunk and fix this, I would be happy too.

FooSoft:
Thanks for the report -- if you want to dive in and figure fix this up, I would appreciate the help! A good place to look is the `docRangeFromPoint` function in `document.js`.

RodericDay:
So, removing the opacity on the imposter, the problem becomes clear.

Under a CSS transformation, it's not overlaid over the original input element like it should be. In fact, under very large scales, there's no overlap whatsoever, and the alert can be triggered reliably.

I'm looking into various ways of ensuring that the offset is correct.

One thing I'm doing very stupidly right now, though, is reloading the code for testing. I just change a file, recreate the zip file, reinstall it in firefox, and test. It's not clear to me (first extension I work on) if I can make the reload automatic somehow.

FooSoft:
[...]

PR #91, opened 2017-10-13
Place `nodeType` checks inside `caretRangeFromPoint` function

### LINKED ISSUE #80: [bug] Error: IndexSizeError: Index or size is negative or greater than the allowed amount
**Summary:**
On some sites I get the alert `Error: IndexSizeError: Index or size is negative or greater than the allowed amount` when scanning.

**Steps to reproduce:**
1. Open this test case (Firefox 55, Linux or Windows):
```html
<!DOCTYPE html>
<html>
    <head>
        <meta charset="utf-8">
        <title>IndexSizeError</title>
    </head>
    <body>
        <form method="get" action="https://test"><div><input name="test" value="123" type="hidden"><input value="test" type="submit"></div></form>
    </body>
</html>
```
2. Scan anywhere on the page.

**Expected results:**
Nothing should happen because there is no text under the mouse cursor.

**Actual results:**
You get an alert with the text  `Error: IndexSizeError: Index or size is negative or greater than the allowed amount`.

---

**More info:**
- The error in this case has something to do with the `<form>` being an one-liner. If I add newlines between `<form>` and `<div>`, there are no errors.
- On dev branch, the error appears to originate from https://github.com/FooSoft/yomichan/blob/5476c100467e31d7dac02a4b1c14e62a4e9b34d0/ext/fg/js/frontend.js#L176 or a corresponding line on current master https://github.com/FooSoft/yomichan/blob/0c650dac828b7ab9641396268a66d3a7410d4000/ext/fg/js/frontend.js#L180 (`textSource.containsPoint(point)`)
- I don't think that there are bad consequences from silently ignoring this error unless it can be reproduced with something that actually contains text.
- I cannot reproduce it on Chrome/Chromium.

RodericDay:
Checked that it does not un-fix #80
Squelches alerts (but does not fix underlying issue) for #89

### DIFF
--- a/ext/fg/js/document.js
+++ b/ext/fg/js/document.js
@@ -76,7 +76,7 @@ function docRangeFromPoint(point) {
     if (!document.caretRangeFromPoint) {
         document.caretRangeFromPoint = (x, y) => {
             const position = document.caretPositionFromPoint(x,y);
-            if (position && position.offsetNode) {
+            if (position && position.offsetNode && position.offsetNode.nodeType === Node.TEXT_NODE) {
                 const range = document.createRange();
                 range.setStart(position.offsetNode, position.offset);
                 range.setEnd(position.offsetNode, position.offset);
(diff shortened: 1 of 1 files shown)


RodericDay:
I just realized I wasn't working from `dev` branch. Will close and reopen, sorry.

PR #95, opened 2017-10-19
Feature: result grouping by main dictionary sequence (along with some other changes)

siikamiika:
I can't find anything major to change or implement, so I'm opening this pull request. Some minor things below:

- I'm not sure if `compactTags` and `compactGlossaries` should be on by default for [new users](https://github.com/siikamiika/yomichan/blob/feature-merge-similar-results/ext/bg/js/options.js#L196). They will be off for [existing users](https://github.com/siikamiika/yomichan/blob/feature-merge-similar-results/ext/bg/js/options.js#L270).

- I tested on a Windows Firefox that `utilStringHashCode` gives the same hash for the first revision of the field templates (it could have messed up with `\r\n` after `formRead`). Can we trust this or should the string be normalized before hashing?

### DIFF
--- a/ext/bg/js/dictionary.js
+++ b/ext/bg/js/dictionary.js
@@ -368,7 +368,8 @@ async function dictFieldFormat(field, definition, mode, options) {
             merge: options.general.resultOutputMode === 'merge',
             modeTermKanji: mode === 'term-kanji',
             modeTermKana: mode === 'term-kana',
-            modeKanji: mode === 'kanji'
+            modeKanji: mode === 'kanji',
+            compactGlossaries: options.general.compactGlossaries
         };
 
         const html = await apiTemplateRender(options.anki.fieldTemplates, data, true);
--- a/ext/bg/js/dictionary.js
+++ b/ext/bg/js/dictionary.js
@@ -124,6 +124,7 @@ function dictTermsCompressTags(definitions) {
             filterOutCategories.push('dictionary');
         } else {
             lastDictionary = dictionary;
+            lastPos = '';
         }
 
         if (lastPos === pos) {
--- a/ext/bg/js/dictionary.js
+++ b/ext/bg/js/dictionary.js
@@ -112,11 +112,11 @@ function dictTermsUndupe(definitions) {
 
 function dictTermsCompressTags(definitions) {
     let lastDictionary = '';
-    let lastPos = '';
+    let lastPartOfSpeech = '';
 
     for (const definition of definitions) {
         const dictionary = JSON.stringify(definition.definitionTags.filter(tag => tag.category === 'dictionary').map(tag => tag.name).sort());
-        const pos = JSON.stringify(definition.definitionTags.filter(tag => tag.category === 'pos').map(tag => tag.name).sort());
+        const partOfSpeech = JSON.stringify(definition.definitionTags.filter(tag => tag.category === 'partOfSpeech').map(tag => tag.name).sort());
 
         const filterOutCategories = [];
 
@@ -124,13 +124,13 @@ function dictTermsCompressTags(definitions) {
             filterOutCategories.push('dictionary');
         } else {
             lastDictionary = dictionary;
-            lastPos = '';
+            lastPartOfSpeech = '';
         }
 
-        if (lastPos === pos) {
-            filterOutCategories.push('pos');
+        if (lastPartOfSpeech === partOfSpeech) {
+            filterOutCategories.push('partOfSpeech');
         } else {
-            lastPos = pos;
+            lastPartOfSpeech = partOfSpeech;
         }
 
         definition.definitionTags = definition.definitionTags.filter(tag => !filterOutCategories.includes(tag.category));
--- a/ext/bg/js/dictionary.js
+++ b/ext/bg/js/dictionary.js
@@ -227,7 +227,6 @@ function dictTermsMergeByGloss(result, definitions, appendTo, mergedIndices) {
 
         for (const tag of definition.tags) {
             if (dictIsJmdictTermTag(tag)) {
-                // TODO: expand tags
                 result.expressions.get(definition.expression).get(definition.reading).add(tag);
             } else {
                 definitionsByGloss[gloss].tags.add(tag);
--- a/ext/bg/js/dictionary.js
+++ b/ext/bg/js/dictionary.js
@@ -89,7 +89,7 @@ function dictTermsSort(definitions, dictionaries=null) {
             return 1;
         }
 
-        return v2.expression.localeCompare(v1.expression);
+        return v2.expression.toString().localeCompare(v1.expression.toString());
     });
 }
 
--- a/ext/bg/js/dictionary.js
+++ b/ext/bg/js/dictionary.js
@@ -144,10 +144,10 @@ function dictTermsGroup(definitions, dictionaries) {
     return dictTermsSort(results);
 }
 
-function dictTermsMergeBySequence(definitions) {
+function dictTermsMergeBySequence(definitions, mainDictionary) {
     const definitionsBySequence = {'-1': []};
     for (const definition of definitions) {
-        if (definition.sequence > 0) {
+        if (mainDictionary === definition.dictionary && definition.sequence > 0) {
             if (!definitionsBySequence[definition.sequence]) {
                 definitionsBySequence[definition.sequence] = {
                     reasons: definition.reasons,
--- a/ext/bg/js/dictionary.js
+++ b/ext/bg/js/dictionary.js
@@ -196,7 +196,7 @@ function dictTermsMergeByGloss(result, definitions, appendTo, mergedIndices) {
             }
         }
 
-        const gloss = JSON.stringify(definition.glossary);
+        const gloss = JSON.stringify(definition.glossary.concat(definition.dictionary));
         if (!definitionsByGloss[gloss]) {
             definitionsByGloss[gloss] = {
                 expression: new Set(),
--- a/ext/bg/js/dictionary.js
+++ b/ext/bg/js/dictionary.js
@@ -229,7 +229,7 @@ function dictTermsMergeByGloss(result, definitions, appendTo, mergedIndices) {
             definitionsByGloss[gloss] = {
                 expression: new Set(),
                 reading: new Set(),
-                definitionTags: new Set(),
+                definitionTags: [],
                 glossary: definition.glossary,
                 source: result.source,
                 reasons: [],
@@ -254,10 +254,8 @@ function dictTermsMergeByGloss(result, definitions, appendTo, mergedIndices) {
         }
 
         for (const tag of definition.definitionTags) {
-            if (typeof tag === 'string') {
-                definitionsByGloss[gloss].definitionTags.add(tag);
-            } else if (tag.category && tag.category !== 'dictionary') {
-                definitionsByGloss[gloss].definitionTags.add(tag.name);
+            if (!definitionsByGloss[gloss].definitionTags.find(existingTag => existingTag.name === tag.name)) {
+                definitionsByGloss[gloss].definitionTags.push(tag);
             }
         }
 
--- a/ext/bg/js/dictionary.js
+++ b/ext/bg/js/dictionary.js
@@ -207,7 +207,8 @@ async function dictFieldFormat(field, definition, mode, options) {
         const data = {
             marker,
             definition,
-            group: options.general.groupResults,
+            group: options.general.resultOutputMode === 'group',
+            merge: options.general.resultOutputMode === 'merge',
             modeTermKanji: mode === 'term-kanji',
             modeTermKana: mode === 'term-kana',
             modeKanji: mode === 'kanji'
--- a/ext/bg/js/dictionary.js
+++ b/ext/bg/js/dictionary.js
@@ -175,7 +175,7 @@ function dictTermsGroup(definitions, dictionaries) {
 function dictTermsMergeBySequence(definitions, mainDictionary) {
     const definitionsBySequence = {'-1': []};
     for (const definition of definitions) {
-        if (mainDictionary === definition.dictionary && definition.sequence > 0) {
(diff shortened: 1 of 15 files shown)


FooSoft:
Thanks for the PR! This is pretty large, so I will be looking this over the next couple of days (probably only get to running my normal tests on it this weekend).

ext/bg/js/util.js
```diff
@@ -26,6 +26,47 @@ function utilIsolate(data) {
...
+function utilSetDifference(setA, setB) {
+    return new Set(
+        [...setA].filter(value => !setB.has(value))
+    );
+}
+
+function utilStringHashCode(string) {
+    let hashCode = 0;
+
+    if (string.length === 0) {
```
FooSoft: This conditional is unnecessary since if `string.length == 0` the for loop will not execute and you will end up return `0` anyway.

ext/bg/js/translator.js
```diff
@@ -158,21 +266,33 @@ class Translator {
...
+                    });
+                }
             }
         }
     }
 
     async expandTags(names, title) {
         const tags = [];
         for (const name of names) {
+            if (typeof name !== 'string') {
```
FooSoft: What else could `name` be?

ext/bg/js/translator.js
```diff
@@ -45,9 +46,113 @@ class Translator {
...
+        const secondarySearchTitles = Object.keys(options.dictionaries).filter(dict => options.dictionaries[dict].allowSecondarySearches);
+        const titles = Object.keys(dictionaries);
+        const {length, definitions} = await this.findTerms(text, dictionaries, alphanumeric);
+
+        const definitionsBySequence = dictTermsMergeBySequence(definitions, mainDictionary);
+
+        const definitionsMerged = [];
+        const mergedByTermIndices = new Set();
+        for (const sequence in definitionsBySequence) {
+            if (!(sequence >= 0)) {
```
FooSoft: That's a pretty strange way of saying `if (sequence < 0)`...

ext/bg/js/translator.js
```diff
@@ -45,9 +46,113 @@ class Translator {
...
+                continue;
+            }
+
+            const result = definitionsBySequence[sequence];
+
+            const rawDefinitionsBySequence = await this.database.findTermsBySequence(Number(sequence), mainDictionary);
+            const definitionsByGloss = dictTermsMergeByGloss(result, rawDefinitionsBySequence);
+
+            const secondarySearchResults = [];
+            if (secondarySearchTitles.length) {
```
FooSoft: Minor: prefer explicit boolean expressions for things that aren't checking whether or not something is defined, meaning this should check `secondarySearchTtiles.length > 0`. In the coding style of Yomichan it makes it look like you are testing if `length` is `undefined` or `null`.

ext/bg/js/options.js
```diff
@@ -205,6 +260,24 @@ function optionsVersion(options) {
...
+        () => {
+            if (options.general.groupResults) {
+                options.general.resultOutputMode = 'group';
+            } else {
+                options.general.resultOutputMode = 'split';
+            }
+            options.general.compactTags = false;
+            options.general.compactGlossaries = false;
+            if (utilStringHashCode(options.anki.fieldTemplates) !== -805327496) { // a3c8508031a1073629803d0616a2ee416cd3cccc
+                options.anki.fieldTemplates = '{{#if merge}}\n' +
```
FooSoft: Prefer template string as it does not require explicit line breaks and you can just use all of the expressions inline without having to combine strings with `+` operator.

ext/bg/js/options.js
```diff
@@ -205,6 +260,24 @@ function optionsVersion(options) {
...
                 options.scanning.modifier = 'none';
             }
+        },
+        () => {
+            if (options.general.groupResults) {
+                options.general.resultOutputMode = 'group';
+            } else {
+                options.general.resultOutputMode = 'split';
+            }
+            options.general.compactTags = false;
```
FooSoft: You don't need to set `compactTags` and `compcatGlossaries` defaults here since they are already specified in the default options object. The first thing that happens when options are loaded is that the default options are merged onto the user options, meaning any missing fields are filled in. The only time you have to manually do things in version blocks is if you want to base settings on a converted version settings used in a prior options version.

ext/bg/js/handlebars.js
```diff
@@ -75,6 +75,18 @@ function handlebarsMultiLine(options) {
     return options.fn(this).split('\n').join('<br>');
 }
 
+function handlebarsTermFrequencyColor(options) {
+    const termFrequency = options.fn(this);
+
+    if (termFrequency === 'popular') {
```
FooSoft: Probably better to do this with classes. Having to update `js` files to change color scheme is odd.

ext/bg/js/dictionary.js
```diff
@@ -178,6 +318,26 @@ function dictTagsSort(tags) {
     });
 }
 
+function dictTermTagScore(tags) {
+    let score = 0;
+
+    const tagScores = {
```
FooSoft: Maybe I'm missing something, but aren't these actually stored in the metadata in the dictionary JSON?

ext/bg/js/dictionary.js
```diff
@@ -144,6 +172,118 @@ function dictTermsGroup(definitions, dictionaries) {
...
+        // result->expressions[ Expression1[ Reading1[ Tag1, Tag2 ] ], Expression2, ... ]
+        if (!result.expressions.has(definition.expression)) {
+            result.expressions.set(definition.expression, new Map());
+        }
+        if (!result.expressions.get(definition.expression).has(definition.reading)) {
+            result.expressions.get(definition.expression).set(definition.reading, new Set());
+        }
+
+        for (const tag of definition.definitionTags) {
+            if (typeof tag === 'string') {
```
FooSoft: Why again is it that `definitionTags` has tags that are either strings or objects? Is there a reason why they aren't objects always?

ext/bg/js/settings.js
```diff
@@ -237,6 +243,18 @@ function dictionaryGroupsSort() {
     dictGroups.append(dictGroupChildren);
 }
 
+function dictionarySetMain(e) {
+    const mainDictionary = $(e.target).closest('.dict-group');
+    const mainDictionaryTitle = mainDictionary.data('title');
+
+    $('.dict-group').each((index, element) => {
+        const dictionary = $(element);
+        if (dictionary.data('title') !== mainDictionaryTitle) {
```
FooSoft: I don't think that this should be necessary if all of the radio buttons are in the same group (set by `name` attribute). Setting one should unset the others.

FooSoft:
I'm not sure I like the UX of the radio buttons for setting the "main" dictionary, but I don't yet know what would be better. Radio buttons are usually shown grouped closely together, and the fact that they are so far apart looks odd. Also I notice that after clearing dictionaries (or perhaps even importing dictionaries for the first time), none of the dictionaries are set as "main". If we want to support not having a "main" dictionary, then radio button is not the correct control to use. It should be a dropdown list with the names of all the dictionaries with the first option being set to "None" or something similar.

ext/bg/js/dictionary.js
siikamiika: `dictTermsMergeByGloss` is used for definitions from `Database.findTermsBySequence` first (tags aren't expanded) and a mix of definitions from `Translator.findTerms` (tags expanded) and `Database.findTermsExact` (not expanded) after that. The tag metadata isn't needed in `dictTermsMergeByGloss` currently, so I didn't think it was necessary to expand the tags in the rest of the definitions at that point yet.

ext/bg/js/dictionary.js
siikamiika: The score for terms in `term_bank_n.json` is currently the result of `termTags` like `P` or `news` and `definitionTags` like `arch`. Maybe yomichan-import should be updated to have a score separately for both of these. But that would require changes to the sorting. `tag_bank_n.json` only has `[Name, Category, Order, Notes]`.

ext/bg/js/options.js
siikamiika: Yeah, seems like they are useless at the moment. I wasn't sure if they should be set true by default for new users but false for existing users that haven't had the features in the first place.

ext/bg/js/translator.js
siikamiika: Huh, I wonder what I was thinking there :D I even modified that afterwards due to an off-by-one error caused by confusion over whether the sentinel value should be `0` or `-1`.

ext/bg/js/translator.js
siikamiika: This is related to the issue with `dictTermsMergeByGloss`. Some of the definitions come from `Translator.findTerms` and have their tags expanded while others don't. A better way to handle this would probably be to check the type before even calling `expandTags`.

tmpl/terms.html
```diff
@@ -25,13 +34,50 @@
...
+            </div>
+            {{/if}}
+            {{~#if frequencies~}}
+            <div class="frequencies">
+                {{~#each frequencies~}}
+                <span class="label label-default tag-frequency">{{dictionary}}:{{frequency}}</span>
+                {{/each}}
+            </div>
+            {{/if}}<!--
+     --></div><!--
```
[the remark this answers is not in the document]
siikamiika: These are probably result of me trying to eliminate some whitespace between the term and the preceding comma. Seems like I wasn't aware of `~` at that moment.

siikamiika:
I remember that I planned setting the first imported dictionary that has `sequence` to the main dictionary but failed to find a good way to do it without changing `format` in `index.json`. Even in that case there could be 0 dictionaries that have `sequence` so radio buttons would still be the wrong approach. A dropdown list (that appears when merged mode is enabled?) sounds like a good solution. Also, it would be nice if the dropdown only contained dictionaries that actually have `sequence` but I can't think of a good way to do this with the current dictionary format.

ext/bg/js/dictionary.js
FooSoft: I think tags should be expanded from the start for everything for clarity sake. It's much harder to reason about code when the type is not explicit. Of course, in languages like JavaScript, explicit type declaration is not an option (unless you use Typescript or something), so we have to rely on convention. By making tags be the same thing everywhere it's a lot easier to write bug-free code. Using `typeof` should be reserved only for checking if something that could have a false-like value is `undefined`.

ext/bg/js/dictionary.js
FooSoft: In that case, I think it would be fine to introduce `Score` to the tag array (at the end, as usual). As you've probably noticed, a lot of effort is made to make sure that all data comes from data files only. The only exception is for styling (that's why it is OK that tag colors are in the style sheets).

ext/bg/js/options.js
FooSoft: Yeah the idea is that you only have to write versions that transform prior data *or* you are fixing up broken options data that you introduced with a bug, which I have totally never had to do before.

ext/bg/js/translator.js
FooSoft: Same comment as the one about the tags above. You should always know what tags are, and it should be possible to infer what type the variable is just from conventions in the code.

FooSoft:
> Also, it would be nice if the dropdown only contained dictionaries that actually have sequence but I can't think of a good way to do this with the current dictionary format.

This can be achieved by expanding on the data contained in the database:
```
dictionaries: '++,title,version'
```

We already store meta information about the dictionaries; all you would have to do is add a column like "hasSequences" and set it to true if all of the sequence data coming out of the dictionary is valid (`!= -1`). Versioning the `dictionaries` table is not going to be required since `hasSequences`  does not have to be a database key. You should be able to just grab all the dictionaries and iterate over them and check for the truthiness of `hasSequences`. Dictionaries imported with older versions of Yomichan will have a value of `undefined` for the column, which will will work just fine instead of `false`.

siikamiika:
The problem areas should be better now. The `<!-- comment -->` issue in `terms.html` actually required some more of them to get fixed completely (but I think I also replaced some with `~`). Not sure if Handlebars would have some cleaner way to do it but this is how I've done it in the past because it lets you keep the markup indented.

ext/bg/js/settings.js
```diff
@@ -93,13 +104,37 @@ function formUpdateVisibility(options) {
...
 }
 
+async function formMainDictionaryOptionsPopulate(options) {
+    const select = $('#main-dictionary').empty();
+
+    let titles = await utilDatabaseGetTitlesWithSequences();
+    titles = titles.filter(title => options.dictionaries[title].enabled);
+    const formOptionsHtml = [];
+    let mainDictionarySelected = false;
+    for (title of titles) {
```
FooSoft: You probably mean `for (const title of titles)`; not having `const` here makes it this look kind of wonky.

ext/bg/settings.html
```diff
@@ -48,6 +52,20 @@
...
                 <div class="form-group">
+                    <label for="result-output-mode">Result grouping</label>
+                    <select class="form-control" id="result-output-mode">
+                        <option value="group">Group results by term-reading pairs</option>
+                        <option value="merge">Group results by main dictionary entry ID (experimental)</option>
+                        <option value="split">Split definitions to their own results</option>
+                    </select>
+                </div>
+
+                <div class="form-group options-merge">
```
FooSoft: Please move this under the "Dictionaries" section.

FooSoft:
Looking good :+1: 

One last thing: I think it would be more visually appealing to have the `(Not selected)` text appear in a muted color (add the `text-muted` class to the `option`, drop the parens).

siikamiika:
Nice! I hope it doesn't break things when released and that users find it useful. It must be confusing that the mode doesn't work with old dictionaries but it should help that they don't show up in the dropdown. Feel free to rename things to make them easier to understand and remove the warning (experimental) if you think this feature is ready.

FooSoft:
🎉 🎈

siikamiika:
👍

PR #97, opened 2017-11-09
merged mode: expand termTags earlier

siikamiika:
Fixes errors with main dictionary other than JMdict. Forgot to do this together with `definitionTags`


FooSoft:
Thanks for the fix 👍

PR #102, opened 2017-12-16
Add option to play audio automatically

guidocella:
Adds an option to play the audio of the first term automatically when a popup opens and when a search is performed. This was the only feature I missed from Rikaisama other than the pitch accent information. Personally I always want to play the audio of the word I'm checking, and ALT+P is awkward to press with only your left hand.


FooSoft:
Looks good. Only request I have is that you reopen this against the `dev` branch since `master` is meant to reflect the extension in its current state on the chrome and firefox marketplaces.

PR #103, opened 2017-12-16
Add option to play audio automatically

guidocella:
Same as #102 but against the `dev` branch. Sorry for the mistake, this is actually the first time I mess with a browser extension. :)


FooSoft:
Thanks for contributing!

PR #140, opened 2018-10-02
Add horizontal popup offset

### LINKED ISSUE #113: Popup horizontal offset
There is `Popup offset` in settings but it changes vertical offset.

Horizontal offset would be useful in case of horizontal text, which is often used in Japanese (manga, ranobe, etc).

KarboniteKream:
This PR implements an option for a horizontal popup offset.

Fixes #113.


FooSoft:

PR #156, opened 2019-02-17
Fix alternate audio URLs

toasted-nutbread:
* HTML format for ```'jpod101-alternate'``` seems to have changed at some point, so the HTML node with the URL had to be updated.
* jisho.org now uses https by default.
* Audio URLs are now normalized. This fixes an issue where audio from jisho.org could not be played back on file:/// URLs since the audio URLs look like: ```"//d1vjc5dkcd3yh2.cloudfront.net/audio/21df5faaebe19fd4355d40fbe2fd97aa.mp3"```

### DIFF
--- a/ext/bg/js/audio.js
+++ b/ext/bg/js/audio.js
@@ -72,7 +72,7 @@ async function audioBuildUrl(definition, mode, cache={}) {
                     const url = row.querySelector('audio>source[src]').getAttribute('src');
                     const reading = row.getElementsByClassName('dc-vocab_kana').item(0).innerText;
                     if (url && reading && (!definition.reading || definition.reading === reading)) {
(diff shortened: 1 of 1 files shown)


ext/bg/js/audio.js
```diff
@@ -112,6 +115,24 @@ async function audioBuildUrl(definition, mode, cache={}) {
     }
 }
 
+function normalizeAudioUrl(url, baseUrl, basePath) {
```
FooSoft: Please rename to `audioUrlNormalize` to be consistent with all the other functions in the file, which have the `audio` prefix.

FooSoft:
Thanks for the change!

FooSoft:

PR #157, opened 2019-02-17
Add support for touch input

toasted-nutbread:
* Text can now be touched to bring up the dictionary popup.
* Feature can be enabled or disabled through settings (enabled by default).
* When touch input is disabled, touch events will not be hooked.
* An option to allow the dictionary popup be fixed to the bottom or top of the page has also been added. This mode may be less intrusive when physically touching a screen and moving along the text.
* Touch inputs can be tested without a touchscreen in Chrome by using the touchscreen device simulation mode.

### DIFF
--- a/ext/fg/js/document.js
+++ b/ext/fg/js/document.js
@@ -92,7 +92,19 @@ function docRangeFromPoint(point) {
 
     if(imposter !== null) imposter.style.zIndex = -2147483646;
 
-    const rect = range.getClientRects()[0];
+    let rects;
+    try {
+        rects = range.getClientRects();
+    }
+    catch (e) {
+        return;
+    }
+
+    if (rects.length === 0) {
+        return;
+    }
+
+    const rect = rects[0];
     if (point.y > rect.bottom + 2) {
         return;
     }
--- a/ext/fg/js/document.js
+++ b/ext/fg/js/document.js
@@ -85,20 +85,18 @@ function docRangeFromPoint(point) {
                 range.setEnd(position.offsetNode, position.offset);
                 return range;
             }
+            return null;
         };
     }
 
     const range = document.caretRangeFromPoint(point.x, point.y);
+    if (range === null) {
+        return;
+    }
 
     if(imposter !== null) imposter.style.zIndex = -2147483646;
 
-    let rects;
-    try {
-        rects = range.getClientRects();
-    }
-    catch (e) {
-        return;
-    }
+    const rects = range.getClientRects();
 
     if (rects.length === 0) {
         return;
--- a/ext/fg/js/frontend.js
+++ b/ext/fg/js/frontend.js
@@ -96,7 +96,7 @@ class Frontend {
 
         const search = async () => {
             try {
-                await this.searchAt({x: e.clientX, y: e.clientY}, Frontend.SearchType.Mouse);
+                await this.searchAt({x: e.clientX, y: e.clientY}, 'mouse');
             } catch (e) {
                 this.onError(e);
             }
@@ -203,7 +203,7 @@ class Frontend {
         }
 
         const touch = touches[index];
-        this.searchFromTouch(touch.clientX, touch.clientY, Frontend.SearchType.TouchMove);
+        this.searchFromTouch(touch.clientX, touch.clientY, 'touchMove');
 
         e.preventDefault(); // Disable scroll
     }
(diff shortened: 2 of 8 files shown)


ext/fg/js/document.js
```diff
@@ -92,7 +92,19 @@ function docRangeFromPoint(point) {
 
     if(imposter !== null) imposter.style.zIndex = -2147483646;
 
-    const rect = range.getClientRects()[0];
+    let rects;
+    try {
+        rects = range.getClientRects();
+    }
+    catch (e) {
+        return;
```
FooSoft: What could cause `getClientRects` to throw here? Is it something that could be tested for explicitly as opposed to doing a blanket catch-all?

ext/fg/js/frontend.js
```diff
@@ -262,7 +371,97 @@ class Frontend {
...
+    }
+
+    searchFromTouch(x, y, type) {
+        this.popupTimerClear();
+
+        if (!this.options.general.enable || this.pendingLookup) {
+            return;
+        }
+
+        const pos = {x: x, y: y};
```
FooSoft: This can actually just be `{x, y}`, and at the point where it's that short, you don't even have to have separate constant for it : )

ext/fg/js/frontend.js
```diff
@@ -262,7 +371,97 @@ class Frontend {
...
+            } catch (e) {
+                this.onError(e);
+            }
+        };
+
+        search();
+    }
 }
 
+Frontend.SearchType = {
```
FooSoft: Prefer to use strings instead of explicit type enumerations. I get what you are going for, but it's JS so you don't get any additional type safety and this just requires more typing than `'mouse'`, `'touchStart'`, etc. Strings also have the advantage of being easier to debug than constants and don't introduce code dependencies (if you wanted to access the values).

ext/fg/js/frontend.js
```diff
@@ -262,7 +371,97 @@ class Frontend {
...
+                return i;
+            }
+        }
+        return -1;
+    }
+
+    excludeTouches(touchList, excludeTouchList) {
+        const result = [];
+        for (let i = 0, ii = touchList.length; i < ii; ++i) {
+            let r = result[i];
```
FooSoft: Is this supposed to be `let r = touchList[i]`? You'd be getting `undefined` here. If that's the case, you should also use the `for` `of` loop instead of manually doing indexing.

ext/fg/js/frontend.js
```diff
@@ -262,7 +371,97 @@ class Frontend {
 
         this.textSourceLast = null;
     }
+
+    getPrimaryTouch(touchList) {
+        return touchList.length > 0 ? touchList[0] : null;
+    }
+
+    getIndexOfTouch(touchList, identifier) {
+        for (let i = 0, ii = touchList.length; i < ii; ++i) {
```
FooSoft: Use `for` `of` loop here instead of manually indexing.

ext/fg/js/document.js
toasted-nutbread: Looks like it's ```TypeError: "range is undefined"``` coming from
```js
const range = document.caretRangeFromPoint(point.x, point.y);
```

I thought there was a different exception being thrown that was coming from ```getClientRects``` internally, but I guess not. Looks like it stems from the [fallback implementation](https://github.com/FooSoft/yomichan/pull/157/commits/33ec369eff87ace3e8b49c6461fa1b9a5d2a1202#diff-8dcb9a0ea4375b3bf7ed569620281840L79) of document.caretRangeFromPoint. Somehow that is being triggered in the extension on Firefox Developer despite the standard browser context having it defined.

ext/fg/js/frontend.js
toasted-nutbread: You are correct; updated.

ext/fg/js/frontend.js
toasted-nutbread: ```for``` ```in``` since the index what is returned.

ext/fg/js/frontend.js
toasted-nutbread: Not sure why I overcomplicated that. Fixed.

ext/fg/js/frontend.js
toasted-nutbread: I was probably thinking number comparisons are faster than string comparisons, but after the property accesses it probably doesn't matter. Furthermore:
> it's JS

FooSoft:
Finally got some time to get back to this! Looking good :+1:

FooSoft:

PR #160, opened 2019-05-04
Add support for mobile Firefox

toasted-nutbread:
Mobile Firefox supports addons, and I had added touch support in a previous pull request, so it seemed logical to add this.

Changes:
* HTML pages now look correct for devices with different display scaling values.
* Added a search link to the settings page since it cannot be accessed otherwise on mobile Firefox.
* Fixed some extension commands which are not supported on mobile Firefox.
  * Extension badges and custom commands.
* Added storage usage information to the settings page.
  * Interestingly, Firefox for Android's databases seem to take up about 3x the storage space as desktop Firefox, so adding a dictionary takes 3x as much space.
* Slightly changed how database dictionary insertion is handled in order to not leave the database in a completely unusable state.
  * If adding a dictionary fails while importing, the database will remain full but none of the added content will be usable. This change allows the added content to be used.

### DIFF
--- a/ext/bg/background.html
+++ b/ext/bg/background.html
@@ -2,6 +2,7 @@
 <html lang="en">
     <head>
         <meta charset="UTF-8">
+        <meta name="viewport" content="width=device-width,initial-scale=1" />
     </head>
     <body>
         <script src="/mixed/lib/dexie.min.js"></script>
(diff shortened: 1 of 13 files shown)


FooSoft:
Good work, thanks for the PR :smile:

FooSoft:

PR #166, opened 2019-07-09
Add support for custom popup CSS

toasted-nutbread:
Allows user-defined CSS to be injected into the popup window.


FooSoft:
Looks great, thanks!

PR #171, opened 2019-07-31
return focus to site after hiding popup

siikamiika:
When "Automatically hide results" is checked and empty part of the website is scanned, the popup hides. However, iframe#yomichan-float still has focus and prevents the site from being used normally with keyboard.

This PR fixes the issue by blurring the popup iframe each time its `hide` method is called.


FooSoft:
Looks good!

PR #172, opened 2019-07-31
allow looking up terms from within terms

siikamiika:
<strike>I'd like to have `onTermLookup` inside the Display class so that it would work in search page too, but I couldn't figure out how to import the dependencies from `fg/js/document.js` there. Maybe some files have to be moved around?</strike>

<strike>Also, this doesn't remember history more than the previous page, so that should probably be fixed too.</strike>


siikamiika:
I'm also thinking if there should be a possibility to set a modifier for this to avoid accidentally jumping to other definitions

ext/mixed/js/display.js
```diff
@@ -265,6 +265,7 @@ class Display {
             $('.action-view-note').click(this.onNoteView.bind(this));
             $('.action-play-audio').click(this.onAudioPlay.bind(this));
             $('.kanji-link').click(this.onKanjiLookup.bind(this));
+            $('.glossary li').click(this.onTermLookup.bind(this));
```
siikamiika: This doesn't work when there's only one glossary item, use `.glossary-item` instead

siikamiika:
One thing that becomes more apparent when clicking the term definitions without visual feedback (text selection) is that scanning actually starts half a character left of the actual starting point of the word

siikamiika:
> I'd like to have `onTermLookup` inside the Display class so that it would work in search page too, but I couldn't figure out how to import the dependencies from `fg/js/document.js` there. Maybe some files have to be moved around?
> 
> Also, this doesn't remember history more than the previous page, so that should probably be fixed too.

Importing the dependencies without referencing the script in a HTML page seems possible using the new ES6 [import](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) and [export](https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export) statements provided that the enforced strict mode and module imports clashing with traditional script tags doesn't cause too much problems. For now I took copies of `fg/js/document.js` and `fg/js/source.js` that are used with the new syntax only because otherwise Yomichan wouldn't work.

FooSoft:
I'm all for using ES6 modules, I think it's a pretty elegant solution. This probably means that we have to update the minimum browser versions in the manifest file though, right?

siikamiika:
Great! :+1: Do you have usage statistics about the older versions of Chrome and Firefox? At least Firefox 52 ESR is EOL and the oldest supported ESR is based on 60 which supports modules out of the box (but also drops support for legacy extensions if someone depends on them in addition to Yomichan). I'll update the minimum versions to the manifest file, but there's also the option to use Babel to transpile and support older versions. That could come with a performance penalty.

siikamiika:
Actually it turned out to be very difficult to switch to using ES6 modules, because content scripts cannot be added as modules. I reverted the changes and switched to dependency injection.

siikamiika:
I think the core feature is ready now and I'm open for any suggestions

siikamiika:
A few usability things that come to my mind are the lack of a back button and the popup view jumping to the top of the entry when the back hotkey is pressed. If the popup is small, this could mean that the source word that was clicked is not visible any more when returning to the entry. The problem could probably be solved by storing the absolute position to context before navigating to another entry.

siikamiika:
The above usability improvements are now done. It also feels that the source term button could have absolute position instead of showing it together with every entry, but that's a minor thing

FooSoft:
Excellent work, this is something that I've wanted to have for a really long time, but never gotten around to. I'll publish this to the testing extension, and assuming that there are no issues, I'll move all of the changes there (including this one) to the mainline :+1:

odehjoseph:
Using the testing version and I have some suggestions:

1) This might be too much of a request but do you think it's possible to have another smaller popup show up when searching for terms within terms? I'd like to see the original definition while also hovering over words I don't know and quickly seeing them in another popup instead of clicking the back button repeatedly.

2) For the second search page, can you add the option to assign it its own dictionaries to use? Example usage would be: I'm reading something online and I look up a word with my Japanese dictionary (like 新明解) only to see another word in the popup that I don't understand. Instead of looking up that word with the same Japanese dictionary (using the new search within terms functionality you guys added), I think it would be better to look up those secondary words with a Japanese-English dictionary instead for easy monolingual branching.

siikamiika:
@FooSoft It's also been on my backlog but studies, work and life kept me too busy. :smile: Speaking of releases, I saw some issues that were probably related to users having an old version of Yomichan. Would be nice to see a new release on addons.mozilla.org and chrome web store after some testing feedback (and possible bug fixes or additions) so that everyone can enjoy the new features.

@odehjoseph I understand that it would be more intuitive to get another popup inside the popup, but it has some problems. Should it be possible to have another popup inside the popup inside the popup? How different should the popup be? You mentioned custom dictionaries, but someone could also want to adjust the scanning modifier or grouping mode. One thing I thought is making options a recursive structure where each popup configuration has an optional link to a new popup configuration that the users can define as deep as they wish to have popups, but that's probably not trivial to implement. I'm not saying that the nested popups and term clicking can't coexist, but at least it needs some time.

By the way, have you tried using Alt+B to go back to the source term? At least to me it feels like a very smooth way to navigate because you can keep your mouse cursor where the source term was and you will find the position instantly after returning from the other definitions. If the problem is needing both hands, I could try to make the mouse back button go back to the source term like browser history.

odehjoseph:
@siikamiika Is it possible then to make another extension that contains some of the same code as yomichan but only works with yomichan's popups? Just an idea, but it seems more plausible in my mind, because it would consist mostly of just copying code and slightly editing for functionality in a limited area (yomichan's popups). I don't know about Chrome's rules for extensions working with other extensions so maybe that might be an issue.

I have tried Alt+B and exactly as you mentioned, it gets annoying using two hands. Also, it feels much nicer in my mind if it works through popups so that I can see the original definitions.

siikamiika:
@odehjoseph Unfortunately, as you guessed, the extension API is quite limited when it comes to interacting with other add-ons. You could try adding `"matches": [[...other urls], "chrome-extension://*/*"],` to a custom Yomichan loaded in developer mode, but I don't know if that would work at all.

I was able to make Yomichan open popups inside a popup on Firefox by adding this
```diff
--- a/ext/fg/float.html
+++ b/ext/fg/float.html
@@ -7,6 +7,7 @@
         <link rel="stylesheet" href="/mixed/lib/bootstrap/css/bootstrap.min.css">
         <link rel="stylesheet" href="/mixed/lib/bootstrap/css/bootstrap-theme.min.css">
         <link rel="stylesheet" href="/mixed/css/display.css">
+        <link rel="stylesheet" href="/fg/css/client.css">
         <style type="text/css">
             .entry, .note {
                 padding-left: 10px;
@@ -37,8 +38,10 @@
         <script src="/fg/js/api.js"></script>
         <script src="/fg/js/util.js"></script>
         <script src="/fg/js/document.js"></script>
+        <script src="/fg/js/popup.js"></script>
         <script src="/fg/js/source.js"></script>
         <script src="/mixed/js/display.js"></script>
+        <script src="/fg/js/frontend.js"></script>
 
         <script src="/fg/js/float.js"></script>
     </body>
```
to float.html, but only one layer deep. On Chrome it would show the error `Unexpected response of type object` in console after the first lookup.

In the search mode opened using Alt+Insert I could open a popup inside the search after adding the same scripts to search.html on Chrome but not on Firefox.

odehjoseph:
@siikamiika It's unfortunate to hear that it won't work in Chrome. I tried replacing the matches part in the manifest with what you suggested (``"matches": [["http://*/*", "https://*/*", "file://*/*"], "chrome-extension://*/*"],``) and this was the result: ``Invalid value for 'content_scripts[0].matches[0]': Expect string value.``

siikamiika:
@odehjoseph Actually there's an error in that content script match field, I used [...] for representing the continuation from other matching sites but that's also valid syntax for an array so it was confusing. When you add chrome-extension to the end of the array of matches, you get the error `Invalid value for 'content_scripts[0].matches[3]': Invalid scheme.
Could not load manifest.` meaning that it doesn't support matching chrome extension urls.

If you want to help with the development of this feature, it would be great if you found out why Chrome stops delivering messages between the Yomichan background page and the popup after the first two word lookups when you edit float.html as shown above. We should probably also move this discussion to a new issue that highlights the problems so that it's easier for other contributors to find it.

PR #176, opened 2019-08-10
Don't scan when touching the current selection

toasted-nutbread:
It is sometimes desirable to try to act on the current text selection without performing another scan. For example: a long-press on the selected word to open the context menu.

### DIFF
--- a/ext/fg/js/frontend.js
+++ b/ext/fg/js/frontend.js
@@ -172,7 +172,12 @@ class Frontend {
             return;
         }
 
-        this.setPrimaryTouch(this.getPrimaryTouch(e.changedTouches));
+        let touch = this.getPrimaryTouch(e.changedTouches);
+        if (this.selectionContainsPoint(window.getSelection(), touch.clientX, touch.clientY)) {
+            touch = null;
+        }
+
+        this.setPrimaryTouch(touch);
     }
 
     onTouchEnd(e) {
@@ -451,6 +456,18 @@ class Frontend {
         };
 
         search();
+    }
+
+    selectionContainsPoint(selection, x, y) {
+        for (let i = 0; i < selection.rangeCount; ++i) {
+            const range = selection.getRangeAt(i);
+            for (const rect of range.getClientRects()) {
+                if (x >= rect.left && x <= rect.right && y >= rect.top && y <= rect.bottom) {
+                    return true;
+                }
+            }
+        }
+        return false;
     }
 }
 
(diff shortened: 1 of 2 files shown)


ext/fg/js/frontend.js
```diff
@@ -285,7 +290,8 @@ class Frontend {
             if (!hideResults && (!this.textSourceLast || !this.textSourceLast.equals(textSource))) {
                 searched = true;
                 this.pendingLookup = true;
-                hideResults = !await this.searchTerms(textSource) && !await this.searchKanji(textSource);
+                const focus = (type === 'mouse');
```
siikamiika: I think it would make more sense to pass the type as-is and check if it's mouse later

ext/mixed/js/display.js
```diff
@@ -269,7 +269,9 @@ class Display {
 
     async termsShow(definitions, options, context) {
         try {
-            window.focus();
+            if (context && context.focus) {
```
siikamiika: This could cause a bug when context is null or undefined. Maybe replace with `if (!context || context.focus) {`, or `if (!context || context.type === 'mouse') {` if you do the search type check later

ext/mixed/js/display.js
```diff
@@ -321,7 +323,9 @@ class Display {
 
     async kanjiShow(definitions, options, context) {
         try {
-            window.focus();
+            if (context && context.focus) {
```
siikamiika: same as above

ext/fg/js/frontend.js
toasted-nutbread: My reasoning was that it would be better to keep all reference to ```type```'s values (```'mouse'```, ```'touchMove'```, ```'touchStart'```) to the same file.

ext/mixed/js/display.js
toasted-nutbread: Fixed.

ext/mixed/js/display.js
toasted-nutbread: Fixed.

ext/fg/js/frontend.js
siikamiika: Ah, I wasn't familiar with the code that much. Just in general I think that transforming a value for a single purpose in a state object can obscure its origin and it's also less useful if for some reason there must be special handling for touchMove and touchStart events in the future. Anyway I trust your reasoning here because you authored this feature (and it's not like I decide what's accepted anyway :laughing:)

siikamiika:
I'll test this later when I get my phone ready, but the code looks good to me

siikamiika:
I tested the code on my phone, and was able to long-press the currently scan-selected text and scroll the viewport by dragging or flicking from the selected text. I noticed that it's still possible to deselect the currently selected text by touching text that's not yet selected, but I guess that's out of the scope of this pull request. I approve these changes :+1:

FooSoft:
Thanks for testing it on mobile, @siikamiika :+1:

PR #181, opened 2019-08-14
ignore zero-width non-joiner

### LINKED ISSUE #179: No longer scanning full words in Google Docs
Up until yesterday I could use Yomichan to scan full words in a Google Docs document. Today, however, it only wants to scan one character at a time, which isn't helpful at all. It happens even after a browser restart. It seems to be working as normal on other sites, including Google Sheets.

siikamiika:
fixes #179


siikamiika:
Not sure if the string should be cleared in `text()` or `setEndOffset()` and if `strippedIndices` should be returned in some other way. Now the scanning length in settings includes also the stripped characters

siikamiika:
The above error has been fixed

PR #182, opened 2019-08-16
Anki screenshot

### LINKED ISSUE #169: [Feature Request] Screenshot export to Anki (Netflix, etc. support)
You can currently use Yomichan with subtitled videos by using it in combination with extensions like Subadub (actually, not sure if that's even necessary). However, you can't create Subs2srs-style Anki cards. For people who don't pirate videos (e.g. companies, people without access to torrents, etc., etc.), this is a huge issue, because it means they can't easily make cards with pictures to accompany their sentences.

A simple feature like this would basically turn Yomichan into Voracious (without the audio, I guess...)*, but would let it work for things besides just video (i.e., anything in the browser). It's of course possible to take screenshots manually by yourself, but then you have to go through all the trouble of locating the newly created Anki card, etc.

*and btw there's a Chrome extension that lets you automatically pause Netflix videos at each subtitle if that's your thing, so combined with that it'd REALLY be like Voracious.

toasted-nutbread:
Add support for exporting a screenshot of the browser tab to Anki. Resolves #169

* Screenshots can now be exported to Anki using the {screenshot} tag.
* Popup is hidden while the screenshot is taken.
* Advanced settings included for screenshot format (png, jpeg) and quality (0-100).

Tested on Chrome Dev 78, Firefox Developer Edition 69, Firefox for Android Beta 68.

### DIFF
--- a/ext/fg/js/api.js
+++ b/ext/fg/js/api.js
@@ -33,8 +33,8 @@ function apiKanjiFind(text) {
     return utilInvoke('kanjiFind', {text});
 }
 
-function apiDefinitionAdd(definition, mode) {
-    return utilInvoke('definitionAdd', {definition, mode});
+function apiDefinitionAdd(definition, mode, context) {
+    return utilInvoke('definitionAdd', {definition, mode, context});
 }
 
 function apiDefinitionsAddable(definitions, modes) {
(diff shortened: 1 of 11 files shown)


FooSoft:
Sweet!

ext/fg/js/api.js
```diff
@@ -53,6 +53,10 @@ function apiCommandExec(command) {
     return utilInvoke('commandExec', {command});
 }
 
-function apiAudioGetUrl(definition, source) {
```
siikamiika: I guess this was accidentally removed when adding `apiScreenshotGet` and `apiForward`?

FooSoft: Yup... just re-added that.

toasted-nutbread: #183

PR #183, opened 2019-08-17
Fix function which was inadvertently removed

toasted-nutbread:
Noted in PR #182. Thanks @siikamiika for noticing.


toasted-nutbread:
Closing; fixed by @FooSoft in 317a7a1538ee6bf0f16077af01d3cc309d15fa57.

PR #185, opened 2019-08-19
Recursive popups

toasted-nutbread:
Allows content of popups to be scanned in the same way that standard webpages are, creating multiple popups to be created in the parent frame.

The concept here is not all that complicated, although it did require some infrastructure to overcome some of the restrictions of browser extensions.
* A PopupProxy class is used inside of ```/fg/float.html```. The public API of this class is the same as the standard Popup class.
* Messages must be passed from the PopupProxy class to the parent frame, and this frame is ultimately where the real Popup is created. Since the popup is created here and not recursively inside of ```/fg/float.html```, there is no need to worry the popup not fitting or getting too small.
* PopupProxyHost is in charge of forwarding API calls to the real Popup.
* New settings to control whether recursive popups and popups on the search page are enabled. The Javascript code is only loaded on-demand if these options are enabled.

Browser limitations:
* Firefox has a more limited subset of APIs available to [content scripts](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Content_scripts) than Chrome. This includes not having access to [chrome.tabs](https://developer.chrome.com/extensions/tabs) APIs.
* Messages must be able to be passed from the ```/fg/float.html``` frame to the parent frame, and this only seems to be possible by forwarding through the backend as a middle-man.
* The callback feature of the message passing systems ([runtime.sendMessage](https://developer.chrome.com/apps/runtime#method-sendMessage) and [tabs.sendMessage](https://developer.chrome.com/extensions/tabs#method-sendMessage)) does not work if sending another message before invoking ```sendResponse``` because the channel will be closed. Therefore, the [runtime.connect](https://developer.chrome.com/apps/runtime#method-connect) and [tabs.connect](https://developer.chrome.com/extensions/tabs#method-connect) APIs had to be used instead. This is why BackendApiForwarder, FrontendApiReceiver, and FrontendApiSender exist.
* The ```.connect``` calls need to be handled by exactly one window, so some unique IDs/```frameId```s are used to filter this. ```<iframe>```s will otherwise cause issues.

Other:
* Seems to work reliably on Chrome Dev 78 and Firefox Developer Edition 69. Also works on Firefox for Android Beta 68, but the message passing system seems to get flooded a lot easier, which can lead to some timeouts.
* There are some new files added for this, and for the most part, the copyright notice was was just copied from existing files.
* Inspired from a [comment](https://github.com/FooSoft/yomichan/pull/172#issuecomment-517975972) by @odehjoseph in #172 (and a few other issues), I mostly made this to see if it was possible. There's a non-trivial amount of complexity added, so feel free to decline the PR if it doesn't seem beneficial.

Example image:
image

### DIFF
--- a/ext/bg/js/options.js
+++ b/ext/bg/js/options.js
@@ -221,6 +221,7 @@ function optionsSetDefaults(options) {
             modifier: 'shift',
             deepDomScan: false,
             popupNestingMaxDepth: 0,
+            enableOnPopupExpressions: false,
             enableOnSearchPage: true
         },
 
--- a/ext/bg/js/options.js
+++ b/ext/bg/js/options.js
@@ -219,7 +219,8 @@ function optionsSetDefaults(options) {
             delay: 20,
             length: 10,
             modifier: 'shift',
-            deepDomScan: false
+            deepDomScan: false,
+            popupNestingMaxDepth: 0
         },
 
         dictionaries: {},
--- a/ext/bg/js/options.js
+++ b/ext/bg/js/options.js
@@ -220,7 +220,8 @@ function optionsSetDefaults(options) {
             length: 10,
             modifier: 'shift',
             deepDomScan: false,
-            popupNestingMaxDepth: 0
+            popupNestingMaxDepth: 0,
+            enableOnSearchPage: true
         },
 
         dictionaries: {},
--- a/ext/fg/js/frontend-api-sender.js
+++ b/ext/fg/js/frontend-api-sender.js
@@ -1,0 +1,125 @@
+/*
+ * Copyright (C) 2019 Alex Yatskov <alex@foosoft.net>
+ * Author: Alex Yatskov <alex@foosoft.net>
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+
+class FrontendApiSender {
+    constructor() {
+        this.ackTimeout = 3000; // 3 seconds
+        this.responseTimeout = 10000; // 10 seconds
+        this.callbacks = {};
+        this.disconnected = false;
+        this.nextId = 0;
+
+        this.port = chrome.runtime.connect(null, {name: 'backend-api-forwarder'});
+        this.port.onDisconnect.addListener(this.onDisconnect.bind(this));
+        this.port.onMessage.addListener(this.onMessage.bind(this));
+    }
+
+    invoke(action, params, target) {
+        if (this.disconnected) {
+            return Promise.reject('Disconnected');
+        }
+
+        const id = `${this.nextId}`;
+        ++this.nextId;
+
+        return new Promise((resolve, reject) => {
+            const info = {id, resolve, reject, ack: false, timer: null};
+            this.callbacks[id] = info;
+            info.timer = setTimeout(() => this.onError(id, 'Timeout (ack)'), this.ackTimeout);
+
+            this.port.postMessage({id, action, params, target});
+        });
+    }
+
+    onMessage({type, id, data}) {
+        switch (type) {
+            case 'ack':
+                this.onAck(id);
+                break;
+            case 'result':
+                this.onResult(id, data);
+                break;
+        }
+    }
+
+    onDisconnect() {
+        this.disconnected = true;
+
+        const ids = Object.keys(this.callbacks);
+        for (const id of ids) {
+            this.onError(id, 'Disconnected');
+        }
+    }
+
+    onAck(id) {
+        if (!this.callbacks.hasOwnProperty(id)) {
+            console.warn(`ID ${id} not found`);
+            return;
+        }
+
+        const info = this.callbacks[id];
+        if (info.ack) {
+            console.warn(`Request ${id} already ack'd`);
+            return;
+        }
+
+        info.ack = true;
+        clearTimeout(info.timer);
+        info.timer = setTimeout(() => this.onError(id, 'Timeout (response)'), this.responseTimeout);
+    }
+
+    onResult(id, data) {
+        if (!this.callbacks.hasOwnProperty(id)) {
+            console.warn(`ID ${id} not found`);
+            return;
+        }
+
+        const info = this.callbacks[id];
+        if (!info.ack) {
+            console.warn(`Request ${id} not ack'd`);
+            return;
+        }
+
+        delete this.callbacks[id];
+        clearTimeout(info.timer);
+        info.timer = null;
+
+        if (typeof data.error === 'string') {
+            info.reject(data.error);
+        } else {
+            info.resolve(data.result);
+        }
+    }
+
+    onError(id, reason) {
+        if (!this.callbacks.hasOwnProperty(id)) { return; }
+        const info = this.callbacks[id];
+        delete this.callbacks[id];
+        info.timer = null;
+        info.reject(reason);
+    }
+
+    static generateId(length) {
+        let id = '';
+        for (let i = 0; i < length; ++i) {
+            id += Math.floor(Math.random() * 256).toString(16).padStart(2, '0');
+        }
+        return id;
+    }
+}
--- a/ext/fg/js/frontend-api-sender.js
+++ b/ext/fg/js/frontend-api-sender.js
@@ -19,6 +19,7 @@
 
 class FrontendApiSender {
     constructor() {
+        this.senderId = FrontendApiSender.generateId(16);
         this.ackTimeout = 3000; // 3 seconds
         this.responseTimeout = 10000; // 10 seconds
         this.callbacks = {};
@@ -43,11 +44,12 @@ class FrontendApiSender {
             this.callbacks[id] = info;
             info.timer = setTimeout(() => this.onError(id, 'Timeout (ack)'), this.ackTimeout);
 
-            this.port.postMessage({id, action, params, target});
+            this.port.postMessage({id, action, params, target, senderId: this.senderId});
         });
     }
 
-    onMessage({type, id, data}) {
+    onMessage({type, id, data, senderId}) {
+        if (senderId !== this.senderId) { return; }
         switch (type) {
             case 'ack':
                 this.onAck(id);
@@ -69,7 +71,7 @@ class FrontendApiSender {
 
     onAck(id) {
         if (!this.callbacks.hasOwnProperty(id)) {
-            console.warn(`ID ${id} not found`);
+            console.warn(`ID ${id} not found for ack`);
             return;
         }
 
--- a/ext/bg/background.html
+++ b/ext/bg/background.html
@@ -15,6 +15,7 @@
         <script src="/bg/js/anki.js"></script>
         <script src="/bg/js/api.js"></script>
         <script src="/bg/js/audio.js"></script>
+        <script src="/bg/js/backend-api-forwarder.js"></script>
         <script src="/bg/js/database.js"></script>
         <script src="/bg/js/deinflector.js"></script>
         <script src="/bg/js/dictionary.js"></script>
--- a/ext/bg/js/api.js
+++ b/ext/bg/js/api.js
@@ -205,3 +205,8 @@ function apiForward(action, params, sender) {
         chrome.tabs.sendMessage(tabId, {action, params}, (response) => resolve(response));
     });
 }
+
+function apiFrameInformationGet(sender) {
+    const frameId = sender.frameId;
+    return Promise.resolve({frameId});
+}
--- a/ext/bg/js/backend-api-forwarder.js
+++ b/ext/bg/js/backend-api-forwarder.js
@@ -1,0 +1,45 @@
+/*
+ * Copyright (C) 2019 Alex Yatskov <alex@foosoft.net>
+ * Author: Alex Yatskov <alex@foosoft.net>
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+
+class BackendApiForwarder {
+    constructor() {
+        chrome.runtime.onConnect.addListener(this.onConnect.bind(this));
+    }
+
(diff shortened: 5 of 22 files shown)


siikamiika:
This looks really interesting! I'll see about the code later today and this week. Do you think it would also be possible to make popups that appear when scanning a non-Yomichan iframe on the page appear on the parent page instead of inside the iframe?

toasted-nutbread:
I actually did look into that while implementing this. It is possible to change the behavior to support this, although the main issue is being able to determine the correct positioning of the iframe(s) in order to offset the coordinates of the popup on the root frame correctly. Keep in mind that there could be multiple nested iframes.

So it's possible, but I'm not sure how much additional work would be needed beyond what was added in this PR.

siikamiika:
Yeah, better keep this as simple as possible for now. I'm surprised it even works like this. Great work finding out what was preventing the messages from being delivered when having more than one popups! I glanced at the code and it looks good overall, although it's unfortunate that so much extra code is needed to make it work.

Can you try if Android works better after merging the `seekBackwardHelper` hotfix (or master, really) to this branch? I noticed that on a very large page the scanning would get quite slow at the bottom of the page without it, even on desktop.

I have a few comments about the code for now, but I'll try to review the PR more closely next week because I'm busy until the end of the week.

ext/bg/js/options.js
```diff
@@ -214,7 +214,9 @@ function optionsSetDefaults(options) {
             autoHideResults: false,
             delay: 20,
             length: 10,
-            modifier: 'shift'
+            modifier: 'shift',
+            popupNestingMaxDepth: 2,
+            enableOnSearchPage: true
```
siikamiika: I think this should be off by default, because users who don't have Japanese dictionaries installed could find it very confusing if another popup appeared when moving their mouse on top of the expressions above the glossary. Maybe there should also be an option for disabling clicking the terms if the user likes to close the recursive popup by clicking on the page (popup one level below in this case). That has to be kept on by default, though, because there is already a stable release where it's always on. I might have preferred to keep it disabled by default, but too late now.

ext/fg/js/frontend-api-sender.js
```diff
@@ -0,0 +1,128 @@
...
+        const info = this.callbacks[id];
+        delete this.callbacks[id];
+        info.timer = null;
+        info.reject(reason);
+    }
+
+    static generateId(length) {
+        let id = '';
+        for (let i = 0; i < length; ++i) {
+            id += Math.floor(Math.random() * 256).toString(16).padStart(2, '0');
```
siikamiika: This really feels hazardous, but if you can't connect to the backend any other way, it should be fine if the random string is long enough. Do you think this could be sequential instead or would it cause the responses to be delivered to wrong popups?

siikamiika:
One more thing, it could be good to exclude the headwords on top of each entry from nested popups, or take the whitelist approach and only allow scanning `.glossary-item`s like clicking. I can see some use cases for scanning a part of the headword like in your screenshot, but especially users who don't have a scanning modifier could find it annoying if they had to move their mouse cursor past a minefield every time before reaching the glossary.

ext/fg/js/frontend-api-sender.js
toasted-nutbread: There are 3 places a pseudorandom 32-character ID is used:
1. [PopupProxyHost L#53](https://github.com/FooSoft/yomichan/pull/185/files/b64b2df1cc0de3af4629df95a55647443e1e71c1#diff-0c0192ac230f017dd32f41afd4194c98R53) in ```generateUniqueId```
  This could (and likely should) be replaced with a counter which is incremented for each popup created.
2. [FrontendApiSender L#38](https://github.com/FooSoft/yomichan/pull/185/files/b64b2df1cc0de3af4629df95a55647443e1e71c1#diff-4ee0bdacc6e5034db6a70faadee030a9R38) in ```invoke```
  This could also be replaced with a counter which is incremented for each message sent. This ID is used to find the reference to the callback.
3. [FrontendApiSender L#22](https://github.com/FooSoft/yomichan/pull/185/files/b64b2df1cc0de3af4629df95a55647443e1e71c1#diff-4ee0bdacc6e5034db6a70faadee030a9R22) in ```constructor```
  This is the only unique identifier which is truly necessary. I believe this was the unique ID which was used to determine the correct destination when receiving callback messages when there are iframes on the page. The [```Port```](https://developer.chrome.com/extensions/runtime#type-Port) type can have multiple receivers on either end, so this is why this was necessary.

The uniqueness of the psuedorandom IDs should be adequate for the 3rd case, as there only needs to be enough unique IDs as the amount of frames on the page. Something like [```Crypto.getRandomValues```](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues) could be used instead of ```Math.random```, but this wouldn't affect anything in practice. The intent was to generate GUID-like strings using random input, which seems to be a valid way of generating GUIDs (https://www.ietf.org/rfc/rfc4122.txt section 4.4).

toasted-nutbread: Updated to use incrementing index for IDs for 1 and 2.

ext/bg/js/options.js
toasted-nutbread: Disabled by default.

toasted-nutbread:
>  I glanced at the code and it looks good overall, although it's unfortunate that so much extra code is needed to make it work.

I agree, it's unfortunate that this change adds 7 new files; and like I said, it's fine if this is declined. As far as what it would take to get the popups to only spawn on the root frame for non-yomichan iframes, I don't think that would require nearly as much added code (assuming based off of this change's functionality, i.e. PopupProxy).

> Can you try if Android works better after merging the `seekBackwardHelper` hotfix (or master, really) to this branch? I noticed that on a very large page the scanning would get quite slow at the bottom of the page without it, even on desktop.

I did a rebase off of master and the result looks about the same. Everything works well on more simple pages, but the URL I was testing on is fairly slow/resource intensive on Android. I think the page has about 11 frames, mostly due to not having adblock enabled. I think a lot of the messages are a result of ```containsPoint``` calls also.

I think scanning on mobile is slower in general, mostly due to the processors being slower than a PC. For example, importing the basic jmdict_english dictionary on mobile takes about 4-5x longer than it does on PC. (This will obviously vary depending on specs.) I have looked a little bit into seeing if anything can be optimized for the scanning/lookup process, but it's hard to say how much extra performance can be achieved.

>Maybe there should also be an option for disabling clicking the terms if the user likes to close the recursive popup by clicking on the page (popup one level below in this case).

<span>+</span>

>One more thing, it could be good to exclude the headwords on top of each entry from nested popups, or take the whitelist approach and only allow scanning .glossary-items like clicking. I can see some use cases for scanning a part of the headword like in your screenshot, but especially users who don't have a scanning modifier could find it annoying if they had to move their mouse cursor past a minefield every time before reaching the glossary.

I agree, there are probably some usability issues that need to be ironed out. I haven't yet put much time into seeing how this feature fares with different settings/dictionary combinations.

toasted-nutbread:
> One more thing, it could be good to exclude the headwords on top of each entry from nested popups

Implemented in ba1297c381846ff8830ae4e5871936318a71c902.

siikamiika:
Hey! Sorry for the long break. I did some testing on this PR yesterday, took some notes, and forked your branch. I'm still in the process of understanding how everything works, but I'd say I'm about 60% done. Do you think it's best if I create a PR to your branch when I'm ready or do you prefer to do the changes yourself?

I also noticed that you usually rebase or otherwise force-push to this branch, and I'm not proficient enough with git to have my own version of the changes and then update my branch with your changes. I appreciate the effort trying to keep the commit history clean, though. There's a conflict with master now and I don't mind if you force push after rebase because I only have a single commit in my fork, but I mainly mentioned this because I'd like to learn how to do it. :joy: 

My branch: https://github.com/siikamiika/yomichan/tree/recursive-popups

Notes:
# Usability

* When using "Automatically hide results"
    * Return focus to the parent popup instead of the page when a nested popup is closed, because otherwise key bindings won't work.
    Could be caused by: https://github.com/FooSoft/yomichan/pull/171/files
* When using "Automatically hide results" together with "Scan modifier key" = None
    * Don't close the nested popup after mouse is hovered over it.
    Probably also related to https://github.com/FooSoft/yomichan/pull/171/files
* Settings per nesting level (maybe not in the minimal version, though, but this has been requested)
    * Especially enabled (not installed) dictionaries and scanning modifier,
    but maybe even the entire settings including Anki templates
    * Would provide a workaround for the issue with nested popups and no modifier
* Possibility to disable the popup proxy completely to avoid performance issues on low end devices. Could already work, but I haven't investigated enough. This could also be a non-issue after adding performance improvements


# Code

* make containsPoint async everywhere instead of checking (simpler code). implemented on my branch
* move ignoreNodes implementation to source.js as much as possible because it's for low level DOM stuff
* security: use unique addon id as base for message passing source identifier?

siikamiika:
And regarding your comments about having this PR declined, I think it would be a waste because this is such a requested feature. Even if this would never reach mainline because of maintainability concerns, I think there should be a testing version or a separate release for people who are interested in the feature.

toasted-nutbread:
> Do you think it's best if I create a PR to your branch when I'm ready or do you prefer to do the changes yourself?

That's fine if you want to PR to it.

> I also noticed that you usually rebase or otherwise force-push to this branch, and I'm not proficient enough with git to have my own version of the changes and then update my branch with your changes. I appreciate the effort trying to keep the commit history clean, though. There's a conflict with master now and I don't mind if you force push after rebase because I only have a single commit in my fork, but I mainly mentioned this because I'd like to learn how to do it.

I've just been doing ```git rebase master``` or ```git rebase -i master``` via command line, and yeah my intent has been to have a cleaner history, although I do wish that Github would show diffs of rebases.

I don't know if you're using any sort of client for git, but you should just be able to do <code>git rebase --onto <i>upstream</i> HEAD~<b>#</b></code>, where *```upstream```* would be my branch and **```#```** is how many commits you've made. Doing a quick test, it looks like this works.

> make containsPoint async everywhere instead of checking (simpler code). implemented on my branch
That probably makes more sense.

> Possibility to disable the popup proxy completely to avoid performance issues on low end devices. Could already work, but I haven't investigated enough. This could also be a non-issue after adding performance improvements

Setting the options to 0 recursion level should completely preclude the scripts from loading. The PopupProxyHost will still exist, but it's only used to generate the first real popup. Could potentially turn off its runtime message handlers as well if desirable, but I don't think that incurs that much overhead (extra inactive Pipe maybe).

> security: use unique addon id as base for message passing source identifier?

AFAIK there shouldn't be any security issues, since all of the message passing is done via the chrome.runtime.* APIs, and I don't believe these messages are visible to other extensions or the host webpage.

FooSoft:
I agree that this is an awesone feature that would be a shame to miss out on. I will leave the details in @siikamiikas capable hands :grinning:

siikamiika:
@toasted-nutbread Ok, I'll continue with my branch and send a pull request to you so that the code is reviewed by both of us. Apologies for the slow progress, I've been a bit busy again.

It's not that I don't trust you doing the changes, quite the contrary, but rather I don't always know if it makes sense to actually do the changes and after testing I already have the changes I wanted.

Thanks for the git instructions, I'm also using git on the command line and familiar with many of the features but I've been using merge commits instead of rebase so far.

@FooSoft Thanks for approving the feature! I'll do my best but I have to give credit to @toasted-nutbread for finding out a way to do this because I tried a few times without success. I learned a lot about the extension API diving into the PR

siikamiika:
There are now fixes for the focus and child popup closing issues on my branch, and I also managed to rebase onto your branch. I'll look at the other issues next and create a PR to your branch probably some day next week. If everything is ok after that, I think I'm ready to accept this PR

toasted-nutbread:
Are there any parts of the Code/Usability sections in https://github.com/FooSoft/yomichan/pull/185#issuecomment-525914906 that you'd like me to handle? Not sure what all you've done so far/what you're planning on doing.

siikamiika:
You can do the rest of the things on the list if you have time, so refactor
ignoreNodes and see if it's possible to easily make configuration for each
popup (but skip if it's complicated because it's not needed for the
workaround anymore). I was also thinking if I could make Popup.children
point to a single child only and rename it to Popup.child instead, but I
can look into that myself

siikamiika:
I'm ready with my branch except for the refactoring and settings, so I'll open a PR to your branch. I can still do either or both of the changes if you want, but it could take a few days. Let me know if you disagree with any of the changes and I'll fix or revert

siikamiika:
Excellent work! I approve this PR as it is, but I'll leave it open for a day if you want to add something.

toasted-nutbread:
I think I have found the issue you mentioned in https://github.com/toasted-nutbread/yomichan/pull/1#issuecomment-527247855 and fixed it with 9028b55774f788f0b61acadb8d3ba85b2bfab34a.

At least on my end, if you spawn several cascading popups and move the cursor out of the topmost one, the full sequence will often close. I believe it's due to the mouse not being inside the root level popup, caused by [this check](https://github.com/FooSoft/yomichan/blob/9028b55774f788f0b61acadb8d3ba85b2bfab34a/ext/fg/js/frontend.js#L300). The workaround is to check if the cursor is with the bounds of the popup or any of its descendants.

Illustration:
image
Moving cursor along the blue dashed line into the parent popup will frequently cause all popups to close on Firefox. Root level popup outline emphasized with red border overlay.

siikamiika:
@toasted-nutbread I can see why that would fix it, but do you think that's the root cause? My guess is that the mousemove events are leaking through the 1px border to the site due to a bug or a feature in Firefox.

I tried replacing the border styles with `outline` and setting `outline-offset` to `-1px` to the same effect. However, outline doesn't support separate styles for each border, so the full width popups can't set border to none with it.

I still approve the PR because your fix works just as well, but I will hear your thoughts on this before merge

toasted-nutbread:
> @toasted-nutbread I can see why that would fix it, but do you think that's the root cause? My guess is that the mousemove events are leaking through the 1px border to the site due to a bug or a feature in Firefox.

The root cause is that the root page is receiving ```mousemove``` event's. I don't know if I would call it "leaking", because maybe it's intended to do so, but when the cursor moves from one iframe directly into another, the base page will receive a ```mousemove``` event. When this ```mousemove``` event occurs, the scan is not prevented by the ```this.popup.containsPoint(point)``` call, and no text is found underneath the cursor, therefore the root popup closes, subsequently causing its children to close.

This behaviour is mostly consistent on Firefox and Chrome. On Firefox, my testing indicated that only one ```mousemove``` is fired, but Chrome receives two. This difference does not affect the end behaviour. The size of the border should not affect anything, so if the border pixels were actually triggering this behaviour on your end, that might have been something else.

siikamiika:
I tested again on Chrome and realized I've been thinking that this issue didn't exist on Chrome at all. Actually it seems that the CSS workaround only helps on Firefox (68) and not on Chrome. Thanks for explaining how it works!

The PR has been waiting long enough and I think we will get better feedback by letting others test this on the testing branch. I might create another PR soon for disabling term clicking because the features don't work perfectly together.

odehjoseph:
Absolutely fantastic work to everyone who's made this possible! One more thing though to make this commit even better, which may not be considered of high priority, is possibly adding dictionary preferences for different popup levels? I think it would be preferable to have a Japanese to Japanese dictionary for the first level and maybe for the next levels have only Japanese to English. I guess I could just ignore certain dictionaries within the levels that pertain but my English accustomed eyes will not let me do so as well.

Otherwise once again, absolutely outstanding work, especially to @toasted-nutbread!

toasted-nutbread:
@odehjoseph I would consider this as roughly the same as @siikamiika's request for "Settings per nesting level" in https://github.com/FooSoft/yomichan/pull/185#issuecomment-525914906. I think it would have been out of scope to add to this PR, but I'll look into what can be done with settings to provide more flexibility as a follow-up.

FooSoft:
Yup, big thanks to @toasted-nutbread and @siikamiika for making this happen :+1: 
I've pushed this change to testing, and assuming that there are no blocking issues, I'll take it to stable next week.

PR #188, opened 2019-08-23
Add support for Edge browser

### LINKED ISSUE #124: Issue after porting Yomichan to MS Edge
Hello,

I'm aware that this extension is not primarily made for use with MS Edge. However if you could help me out I'd be very happy. The thing is that after loading the extension I get such an error in the options screen.

screenshot_1
screenshot_2

toasted-nutbread:
Resolves #124, but there's a catch.

Here is the [source code](https://github.com/FooSoft/yomichan/blob/168285097004bcc99ecc1dae43e1429c05443d40/ext/bg/js/translator.js#L42) of a basic performance test:
test-code

Tested on [a dictionary site](https://jisho.org/search/kyou) with only the common jmdict_english.zip dictionary installed:
test

Behold, the performance of scanning the word *exactly once* in various different browsers. The result is the number of milliseconds it took to run ```findTerms```.
**Chrome:**
test-chrome

**Firefox:**
test-firefox

**Edge:**
test-edge

---

These numbers can vary a bit, but the gist is that Edge is *very* slow. More often than not, a basic scan will take in excess of 1 second (the test case was actually one of the better timings). The test was not performed on an exceptionally powerful device, but the difference is still quite significant.

The origin of the slowdown seems to be anything involving querying the IndexedDB, I was a single query take ~12ms frequently, sometimes more, sometimes less. Tried with the most recent version of dexie also with the same results. Could also be something to do with Edge's method of scheduling callbacks, as it [wouldn't be the first time](https://danyow.net/edge-promise-perf/) [that has happened](https://wpdev.uservoice.com/forums/257854-microsoft-edge-developer/suggestions/11329845-improve-promise-implementation-performance).

Oh well.

### DIFF
--- a/ext/bg/background.html
+++ b/ext/bg/background.html
@@ -10,6 +10,8 @@
         <script src="/mixed/lib/jszip.min.js"></script>
         <script src="/mixed/lib/wanakana.min.js"></script>
 
+        <script src="/mixed/js/extension.js"></script>
+
         <script src="/bg/js/anki.js"></script>
         <script src="/bg/js/api.js"></script>
         <script src="/bg/js/audio.js"></script>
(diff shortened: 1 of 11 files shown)


FooSoft:
Thanks!

PR #193, opened 2019-08-30
Update isPointInRange to be more accurate

### LINKED ISSUE #173: Scanning a word starts from the middle of the previous character
One thing that becomes more apparent when clicking the term definitions without visual feedback (text selection) is that scanning actually starts half a character left of the actual starting point of the word

_Originally posted by @siikamiika in https://github.com/FooSoft/yomichan/pull/172#issuecomment-517399168_

### LINKED ISSUE #191: No longer recognises vertical text
Since the latest update, it seems Yomichan has been unable to detect any vertical text (even single characters). For example, the text on http://tategaki.github.io/awards2015/

toasted-nutbread:
See https://github.com/FooSoft/yomichan/issues/191#issuecomment-526400339 for more details.

This fixes the issue by increasing the range to include 1 character in forward direction, and then performing a point-in-rect check. If this fails, the range is changed to include 1 character in the backward direction and the point-in-rect check is retried.

Tested on Firefox Developer Edition 70, Chrome Dev 78, primarily using https://tategaki.github.io/awards2015/

Fixes #173
Fixes #191

### DIFF
--- a/ext/fg/js/document.js
+++ b/ext/fg/js/document.js
@@ -17,8 +17,6 @@
  */
 
 
-const IS_FIREFOX = /Firefox/.test(navigator.userAgent);
-
 function docOffsetCalc(element) {
     const scrollTop = window.pageYOffset || document.documentElement.scrollTop || document.body.scrollTop;
     const scrollLeft = window.pageXOffset || document.documentElement.scrollLeft || document.body.scrollLeft;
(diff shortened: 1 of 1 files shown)


FooSoft:
Nice, this change is probably warrants a release soon

toasted-nutbread:
For reference, if anyone ever needs to visualize what these rects look like here is some helper code.

```js

const rects = [];
function createRectVisual(i) {
    if (i < rects.length) { return rects[i]; }
    const r = document.createElement('div');
    r.style.cssText = 'position: absolute; pointer-events: none; background: rgba(255,0,0,0.25); z-index: 100000; outline: 1px solid red;';
    document.body.appendChild(r);
    rects.push(r);
    return r;
}
function showRects(rects) {
    let i = 0;
    const docRect = document.documentElement.getBoundingClientRect();
    for (const rect of rects) {
        const e = createRectVisual(i++);
        e.style.left = `${rect.left - docRect.left}px`;
        e.style.top = `${rect.top - docRect.top}px`;
        e.style.width = `${rect.width}px`;
        e.style.height = `${rect.height}px`;
        e.style.display = 'block';
    }
    for (; i < rects.length; ++i) {
        rects[i].style.display = 'none';
    }
}
// Then call showRects(range.getClientRects());
```

siikamiika:
I checked and this doesn't cause a regression for https://github.com/FooSoft/yomichan/pull/174. Great job fixing multiple issues at once! :smile:

Issue #125, opened 2018-06-07
Issues with imposter document


Garethp:
Hey. So this isn't an issue with Yomichan directly, but rather with how imposter documents are handled if you happen to use the `document.js` in a project that doesn't require holding shift to get definitions. So, the main issues I've found (and solved in my own repo) are:

1. Sometimes the imposter document stays around longer than desired, and can get in the way scrolling through an input box. So I hover over a textarea to get the definition, but then when I try to scroll through that text area, the scrolling has no effect since the imposter document overlays the textarea. Solution? [Right after getting the text ranges, set the z-index be behind the original element](https://github.com/Garethp/RikaiRebuilt/blob/master/src/content/document.js#L86). Since the imposter is only really needed to get the text range, there's no point in keeping in front of the original element any longer than necessary

2. When you get the caret position for an element (easiest seen in page-sized textareas like [Editpad](https://www.editpad.org/)), it can ignore the y-axis. That is to say that if you have some Japanese text at the top of the textarea, and nothing below it, you can put your mouse under (say, 20-30 lines below) the text and it will still create a range for that text. Not so much of an issue when you need to hold down a button to enable Yomi-chan, but for me this was an issue. Solution: [Fetch the rect shape of the text range and check that the mouse is actually near that text range](https://github.com/Garethp/RikaiRebuilt/blob/master/src/content/document.js#L89)

I don't know if you see a need to bring these in to Yomichan, but I figured I may as well report the issues I found and how I solved them anyway

Garethp:
Or, at least, the second fix works in Firefox. I suppose I need to do more testing to say it works in Chrome

FooSoft:
These fixes make sense to me. If you could check that everything is fine in Chrome and send me a pull request, I would be happy to merge it in.

Garethp:
I'ved moved the check for the range outside of the declaration of `document.createRangeFromPoint`, which has it working in Chrome as it does in Firefox. PR made.

Had a quick question, how do you feel your implementation of reading EPWING databases compares to what Rikaisama had? To implement EPWING support, I was planning on just wrapping the existing lookup for Rikaisama in a JSON request/response and just use it through Native Messaging, but if your implementation is much better I'm wondering how difficult it would be to have your project do lookups of a definition, as opposed to just exporting them

FooSoft:
[...]

PR #197, opened 2019-08-31
Imposter element improvements

### LINKED ISSUE #148: Yomichan doesn't detect words at the correct position on Google Translate web site
Google Translate's web page (https://translate.google.com/) was updated recently to have a new look and feel.  Previously, I was able to use Yomichan with Japanese text entered into the "translate from" field on the left side without any issues.  Since the change, Yomichan seems to be detecting the text at an offset from where it actually appears on the screen.  In other words, if I hover over a word, Yomichan doesn't detect it, and therefore doesn't show a pop-up.  But if I move my mouse a bit above the word, Yomichan will then find it and show the pop-up.

For example, if I'm trying to use Yomichan on a word in the first line of "translate from" text, I have to hover over the border between the Translate menu ("Detect Language", etc.) and the text.

If this isn't clear, please let me know and I'll try to take a screenshot.

toasted-nutbread:
This change fixes and improves the use of the imposter element:

* CSS ```margin``` property not being zeroed out. This was causing the imposter element to appear in the wrong location in #148, since the ```<textarea>``` had margins and [```getBoundingClientRect```](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect) does not include margins in the returned ```DOMRect```.
* Fixed element size sometimes not being correct. Due to rendering differences between ```<textarea>``` and ```<div>```, the size of the imposter element would sometimes not match the original textarea. This is usually due to how the size of the scrollbar(s) affect the size of the textarea. The size of the imposter element is now adjusted to match the textarea if there is a difference.
* Removed ```Math.round``` from ```docOffsetCalc``` results, since it reduces accuracy; CSS ```left``` and ```top``` can take decimals values. Rounding could previously cause the imposter position to be off by ~1px. (Usually not noticeable, but I saw no reason to keep it that way.)
* Sets [```pointer-events```](https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events) to ```'none'``` when moving the imposter to the lowest ```z-index``` to further prevent any unwanted interactions.
* Fixes some styling issues with ```<input type="text">```.
  * CSS ```overflow``` set to ```'hidden'``` and ```white-space``` to ```'nowrap'``` to prevent wrapped text. Text inputs are single line.
  * CSS ```line-height``` set to the height of the element in order to get the text to middle align vertically.
* Force all style overrides to use CSS priority ```important```, preventing styles on the underlying webpage from being able to affect it.
* Finally, I also made a change to how the imposter elements are tracked and ultimately removed. The imposter elements are now tracked by the corresponding ```TextSourceRange``` which is created. ```TextSourceRange``` now has a ```cleanup``` function to remove the imposter from the DOM. This should make the lifetime of the imposter more clear, and removes the need to search for nodes to remove using ```document.getElementsByClassName('yomichan-imposter')```.

Fixes #148
Maybe related: #125?

### DIFF
--- a/ext/fg/js/document.js
+++ b/ext/fg/js/document.js
@@ -30,7 +30,7 @@ function docOffsetCalc(elementRect) {
     return {top, left};
 }
 
-function docImposterCreate(element) {
+function docImposterCreate(element, isTextarea) {
     const styleProps = window.getComputedStyle(element);
     const stylePairs = [];
     for (const key of styleProps) {
@@ -49,8 +49,14 @@ function docImposterCreate(element) {
     imposter.style.opacity = 0;
     imposter.style.zIndex = 2147483646;
     imposter.style.margin = '0';
-    if (element.nodeName === 'TEXTAREA' && styleProps.overflow === 'visible') {
-        imposter.style.overflow = 'auto';
+    if (isTextarea) {
+        if (styleProps.overflow === 'visible') {
+            imposter.style.overflow = 'auto';
+        }
+    } else {
+        imposter.style.overflow = 'hidden';
+        imposter.style.whiteSpace = 'nowrap';
+        imposter.style.lineHeight = styleProps.height;
     }
 
     document.body.appendChild(imposter);
@@ -79,8 +85,10 @@ function docRangeFromPoint(point) {
             case 'BUTTON':
                 return new TextSourceElement(element);
             case 'INPUT':
+                imposter = docImposterCreate(element, false);
+                break;
             case 'TEXTAREA':
-                imposter = docImposterCreate(element);
+                imposter = docImposterCreate(element, true);
                 break;
         }
     }
--- a/ext/fg/js/document.js
+++ b/ext/fg/js/document.js
@@ -70,12 +70,6 @@ function docImposterCreate(element) {
     return imposter;
 }
 
-function docImposterDestroy() {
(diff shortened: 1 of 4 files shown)


siikamiika:
Nice work! This indeed fixes #148. While testing on Google Translate with a huge block of text that I noticed another bug that's probably related to the imposter but was also present in master. If the textarea height exceeds the viewport height, the imposter will cause another scrollbar to appear and disappear (both Chrome and Firefox) as the text is scanned, and that causes everything to shake left and right. Does this always happen when appending the imposter to DOM or is it dependent on some CSS property? Another workaround could be to remove the imposter at a later stage when text scanning the textarea is completely over, for example when mouse leaves the textarea.

toasted-nutbread:
> Does this always happen when appending the imposter to DOM or is it dependent on some CSS property? 

This would occur because the imposter element would go past the bounds of the document, causing the size of the document scrollbars to change. It's rather tricky how browsers calculate the bounding size of the document content, since it takes ```position: absolute``` elements into account.

The solution I went with was to place the imposter into a container which has ```overflow: hidden```, thus preventing the underlying page from being resized. There are other ways to solve this that don't involve a container element, but I believe the container element method is the most robust.

Fixed with e3d7ec8db7a86d475fba00d48f9cbe150feb36ff.

Ceiyne:
I was the person who entered #148, and I just got the update that includes this fix.  Thanks!

Issue #189, opened 2019-08-26
Performance optimizations


toasted-nutbread:
I have recently been looking into some optimizations in order to get Yomichan's word scanning to perform faster, particularly on lower-powered mobile devices where scanning is inevitably going to be somewhat slower. Before I jump in and start making unsolicited pull requests, I would like to gauge if this is interesting or worthwhile.

Some preliminary profiling seems to indicate that it's possible to speed up the typical scan by roughly 3x on Chrome, Firefox, and Firefox Mobile. It also looks like the Edge support PR (#188) can be improved to run at nearly-usable speeds, in case that feature is desirable (using the same optimizations, nothing per-browser).

Would it be worthwhile to pursue these optimizations? If so, I can provide further details on what all could be done.

erasedff:
Some optimizations would be great. Reading long-ish articles in Japanese always makes me want to open Firefox and use Rikaichamp instead. Rikai's scan is pretty much instant while Yomichan is quite sluggish, particularly if you set it to mimic Rikai behavior (no hotkey scan, autohide results, low delay).

toasted-nutbread:
I am closing this issue as I now largely consider it to be completed. There may be some more opportunities for optimization, but they are beyond the scope of what I had in mind for this issue. While it's difficult to quantify exactly, the speed improvements feel quite noticeable. Doing a screen recording with a pre-optimized Yomichan and post-optimized Yomichan running simultaneously, post-optimized Yomichan consistently shows the updated popup several frames before pre-optimized Yomichan.

With all of these changes, Yomichan is in a fairly usable state as an extension for Edge. It's certainly not as fast as on Firefox or Chrome, but it's tolerable. Further, some Edge-specific differences may be irrelevant after Edge switches to be Chromium-based. So it's probably viable to release on the Edge extension store if @FooSoft has any interest in doing that.

PR #202, opened 2019-09-04
Temporary object optimization

toasted-nutbread:
First of several optimizations; this change should slightly reduce memory footprint and slightly increase speed. This change largely optimizes several calls to ```Array.concat``` and ```String.split```, which can be replaced with calls which don't generate intermediate arrays. Also makes a change to ```dictTermsGroup``` to only convert the key array to a string once.

Timing information won't be as relevant for this change, especially since the precision of [```performance.now()```](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now) was reduced in most browsers in light of Spectre and similar exploits. However, I did create several tests for the changes in isolation, which can be run in node or browser.

https://gist.github.com/toasted-nutbread/1b8bab884e95e56483af7d494909baef
For all 4 files, ```test1()``` mimics behavior before this change, ```test2()``` mimics the behavior after this change. All changes except ```perf-lookup.js``` are universally faster; ```perf-lookup.js``` is faster in Firefox, but marginally slower in Chrome. This difference should not be noticeable in practice.

#189

### DIFF
--- a/ext/bg/js/dictionary.js
+++ b/ext/bg/js/dictionary.js
@@ -140,16 +140,17 @@ function dictTermsCompressTags(definitions) {
 function dictTermsGroup(definitions, dictionaries) {
     const groups = {};
     for (const definition of definitions) {
-        const key = [definition.source, definition.expression].concat(definition.reasons);
+        const key = [definition.source, definition.expression];
+        key.push(...definition.reasons);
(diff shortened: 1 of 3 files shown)


FooSoft:
Good work -- any optimization (especially those that don't hurt code readability) are a big win.

PR #203, opened 2019-09-06
Remove unsafe-eval from content_security_policy

toasted-nutbread:
As far as I can tell, this is not necessary, nor should it be.

https://developer.chrome.com/extensions/contentSecurityPolicy#relaxing-eval

### DIFF
--- a/ext/manifest.json
+++ b/ext/manifest.json
@@ -57,7 +57,7 @@
         }
     },
     "web_accessible_resources": ["fg/float.html"],
-    "content_security_policy": "script-src 'self' 'unsafe-eval'; object-src 'self'",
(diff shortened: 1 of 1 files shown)


FooSoft:
Pretty sure that this was necessary for handlebar templates to work, since they use `eval` internally (more specifically using `new Function` which is effectively the same thing). Maybe something has changed but it used to be that Chrome store would not even accept Yomichan without this permission.

toasted-nutbread:
Ah, that explains why I couldn't find it. I searched for the standard suspects (```eval```, ```new Function```, etc.) but handlebars was using ```Function.apply```. Thanks for the clarification, will close.

PR #204, opened 2019-09-07
Settings improvements

toasted-nutbread:
Following the commit history may tell a better story than the total diff, but here's a high-level overview:

* First 5 commits are HTML and styling changes.
* Next 2 commits add some validation to the options loading process.
* Next 2 commits refactor the options update process.
* Last 4 commits are intended to clean up and simplify the flow of how options updates are propagated.

I'm trying to keep this description somewhat brief since the commits may speak for themselves, but let me know if I can clarify or expand upon anything.

The goal of this change is to provide a better starting point for some changes to support multiple settings as discussed in #185. It seemed like a better idea to put these changes up for review separately to avoid making everything too convoluted.

### DIFF
--- a/ext/bg/js/api.js
+++ b/ext/bg/js/api.js
@@ -21,12 +21,16 @@ async function apiOptionsSet(options) {
     utilBackend().onOptionsUpdated(options);
 }
 
-async function apiOptionsGet() {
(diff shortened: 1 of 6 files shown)


FooSoft:
Looks good, thanks!

PR #209, opened 2019-09-14
Settings profiles

toasted-nutbread:
This change adds support for multiple profiles; the diff should be a bit more straightforward since all of the prerequisites were taken care of in #204, #205.

### DIFF
--- a/ext/bg/js/api.js
+++ b/ext/bg/js/api.js
@@ -21,9 +21,13 @@ function apiOptionsGet(optionsContext) {
     return utilBackend().getOptions(optionsContext);
 }
 
+function apiOptionsGetFull() {
+    return utilBackend().getFullOptions();
+}
+
 async function apiOptionsSave(source) {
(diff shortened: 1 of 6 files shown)


odehjoseph:
how would you use this to separate settings for different popups? Experimenting with this and I haven't had any luck figuring it out for that purpose.

toasted-nutbread:
> how would you use this to separate settings for different popups? Experimenting with this and I haven't had any luck figuring it out for that purpose.

This feature is the first step towards that end. There will be another feature added after this is ready which will allow you to assign conditions to profiles. These conditions will be the rules which determine how different profiles can be automatically selected based on different contexts, e.g. based on popup level or URL.

toasted-nutbread:
My comment from #213 about isolating objects has been addressed with 9fb89d8f7d77dd70b2493f730d8b224c994a6e98.

FooSoft:
Good work :+1:

PR #218, opened 2019-09-18
Settings profile conditions

toasted-nutbread:
This change adds the ability to assign customizable conditions to profiles, allowing them to be automatically used in different contexts.

There are currently two types of condition:
* **Popup Level**, which tests the level of the popup. This satisfies @siikamiika's request from #185 for "Settings per nesting level".
* **URL**, which tests against the URL of the current page. This type of condition was added for two reasons:
  1) To demonstrate that the conditions can be used for more than just popup level.
  2) As something that is potentially useful for the Anki screenshot functionality added in #182 for #169, since maybe screenshots should only be taken on certain video websites like Netflix, etc.

Each condition has different types of operators, which vary depending on which condition is being used. For example, the **Popup Level** type has standard comparison operators (```= ≠ < > ≤ ≥```), and the **URL** type has one operator for matching domains, and another which takes a regular expression.

Below is the new UI for these conditions, which also has some explanation of how they are used:

---

image

---

FYI @odehjoseph, since you had a question in #209.

### DIFF
--- a/ext/bg/background.html
+++ b/ext/bg/background.html
@@ -16,11 +16,13 @@
         <script src="/bg/js/api.js"></script>
         <script src="/bg/js/audio.js"></script>
         <script src="/bg/js/backend-api-forwarder.js"></script>
+        <script src="/bg/js/conditions.js"></script>
         <script src="/bg/js/database.js"></script>
         <script src="/bg/js/deinflector.js"></script>
         <script src="/bg/js/dictionary.js"></script>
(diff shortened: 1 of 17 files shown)


odehjoseph:
All I can say is: you're a god...

FooSoft:
Looks great to me :+1:

FooSoft:
This change is now in the latest testing release.

Issue #215, opened 2019-09-15
Content of textarea/input not scanned inside XHTML documents


ejls:
The content of `<input/>` and `<textarea/>` is only scanned on HTML pages, not XHTML ones.
You can test it for yourself on [this XHTML page](https://ejls.fr/tmp/textarea.xhtml), note that the text inside `<p/>` is correctly scanned.
This was tested on Firefox 69, no error are reported in the console.

PR #221, opened 2019-09-26
Treat <br /> as line break for extracting sentence.

XiaowenLei:
For example, this is necessary for
https://yarukizero.livejournal.com/166006.html. If <br \> is not treated
as line break, docSentenceExtract finds a really long "sentence".

### DIFF
--- a/ext/fg/js/source.js
+++ b/ext/fg/js/source.js
@@ -140,6 +140,10 @@ class TextSourceRange {
             state.offset = offset + consumed + stripped;
             state.remainder -= consumed;
         } else if (TextSourceRange.shouldEnter(node)) {
+            if (node.tagName === 'BR') {
+                state.content += '\n';
+                state.remainder -= 1;
+            }
             for (let i = 0; i < node.childNodes.length; ++i) {
                 if (!TextSourceRange.seekForwardHelper(node.childNodes[i], state)) {
                     break;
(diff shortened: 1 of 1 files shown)


FooSoft:
The problem here is that this will only be useful for one particular site. What about `<p>` tags? What about `<div>`? Unless there is some sort of generic solution which is able to visually identify logical word groups to be treated as sentences, one-off changes are not going to be globally useful.

toasted-nutbread:
```node.nodeName.toUpperCase()``` should be used instead of ```node.tagName``` in order to work on XHTML pages. See: #215. This is also how ```shouldEnter``` works internally:
https://github.com/FooSoft/yomichan/blob/2fd0c0d9c3c07529c2805563f96bbb0379a2a453/ext/fg/js/source.js#L90-L91

Also, as @FooSoft just mentioned as I was typing this, this operates on the assumption that ```<br>``` tags are used to break sentences. One typical situation where this is not the case (at least in English) is in emails. Some email clients will force line breaks using ```<br>``` where it is not necessarily at the end of a sentence, or some people like doing that manually.

FooSoft:
I'm going to close this for now, as I don't feel comfortable special casing sentence breaks for `<br>` tags. It seems too special-case to me. That said, I like the idea of being able to come up with a way to split text visually, although I don't know the best way to go about this... or if the result of such an operation should still be called a "sentence"...

PR #222, opened 2019-09-26
Add option to disable Alt+B lookup within first popup

odehjoseph:
I liked @toasted-nutbread's rendition of the nested popups and so decided to add an option to enable @siikamiika's first popup search in case anyone got annoyed by having both.

PR #223, opened 2019-09-27
Add option to enable Alt+B lookup

odehjoseph:
I liked @toasted-nutbread's rendition of the nested popups and so decided to add an option to enable @siikamiika's first popup search in case anyone got annoyed by having both.

Also, I'm reposting this because of git issues I had due to inexperience.

Fixed the change request too.

### DIFF
--- a/ext/bg/js/options.js
+++ b/ext/bg/js/options.js
@@ -281,6 +281,7 @@ function profileOptionsCreateDefaults() {
             modifier: 'shift',
             deepDomScan: false,
             popupNestingMaxDepth: 0,
+            enablePopupSearch: false,
             enableOnPopupExpressions: false,
             enableOnSearchPage: true
         },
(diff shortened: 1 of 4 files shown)


odehjoseph:
#222

odehjoseph:
#222 #172

toasted-nutbread:
Note that this change makes "Alt+B" lookup disabled by default, which @siikamiika had a comment about when I was designing nested popups: https://github.com/FooSoft/yomichan/pull/185#discussion_r315891032. Doesn't matter much to me, but now both forms of popup scanning will be disabled by default.

Also, IMO, the label "Enable search within the first popup with Alt+B" is a bit misleading, since from what I understand, Alt+B just goes back _after_ you have already clicked on a glossary term. Alt+B behaviour also applies to clicking ```kanji-link```s. The change also applies to _all_ levels of popups, not just the first level. A more accurate label could be something like "Enable search when clicking glossary entries".

I have plans for a few more fixes/improvements on the settings page, so I could change the label as part of that if everyone thinks that makes sense.

oudajosefu:
Oh yeah, my bad, good catch. I just got really annoyed with the "click glossary entry to search" function so after quickly coding this up, hastily named the button whatever was going on in my head at the time and moved on. Thank you for fixing that mistake.

PR #224, opened 2019-09-28
Remove jQuery usage from Display.js

toasted-nutbread:
There are many places in the ```Display``` class and its subclasses where jQuery is used, but it doesn't (IMO) provide much benefit over native Javascript APIs. The goal of this change is to remove the need to have jQuery loaded in each popup, which would reduce load times and memory footprint. The only thing that didn't really have a counterpart with native Javascript APIs was the smooth scrolling animation, but that is simple enough to recreate (IMO).

Note that I do not think there would be any benefit to try and remove the jQuery dependency from the settings page, as that page actually takes greater advantage of jQuery's features.

This change falls under the purview of #189. Let me know if any of the changes are disagreeable.

---

High level changes:
* Event listeners like ```$(node).on('event', callback)``` has been changed to ```node.addEventListener('event', callback, false)```. jQuery does a lot of extra tracking and wrapping when setting up event callbacks which is largely not used, so eliding this will save some instructions.
* Use direct references to DOM nodes rather than ```$(node)```.
* ```$(node).data('name')``` replaced with ```node.dataset.name```.
* ```$(node).hasClass/addClass/removeClass``` replaced with ```node.classList.contains/add/remove```.

Improvements:
* The slide-up animation on the search page had some issues with jQuery's implementation. Some tweaks to the HTML makes it such that a native CSS transition can be used to accomplish the same animation without any resize jitter.
* ```Display.entryIndexFind``` changed into a non-static. Now does a local search for entries (using ```container.querySelectorAll```), rather than searching in the full DOM (```document.querySelectorAll```). This could be further improved by caching the returned ```NodeList```, but it wasn't a big deal for now.
* ```wheel``` event is now specified to be ```passive:false``` since Chrome's default changed to ```passive:true```. See: [browser compatibility](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Browser_compatibility).
* Current entry now has the ```current-entry``` CSS class applied to it, which also controls visibility of the yellow ✱ indicator. This also provides the ability to use custom popup CSS to apply styles to just the current entry.
* Event listeners now explicitly removed when regenerating HTML. There previously may have been a memory leak here, but it's a bit hard to determine whether the event callbacks were garbage collected or not.

Re-implemented functionality:
* Implemented custom smooth scrolling animation when using Alt+Arrows/Page*/Home/End keys. This was relying on jQuery's ```node.animate({scrollTop: target}, time)``` functionality. This new functionality was built into a new class for API simplicity and reusability.

### DIFF
--- a/ext/bg/js/search.js
+++ b/ext/bg/js/search.js
@@ -28,7 +28,8 @@ class DisplaySearch extends Display {
 
         this.search = $('#search').click(this.onSearch.bind(this));
         this.query = $('#query').on('input', this.onSearchInput.bind(this));
-        this.intro = $('#intro');
+        this.intro = document.querySelector('#intro');
+        this.introHidden = false;
 
         this.dependencies = Object.assign({}, this.dependencies, {docRangeFromPoint, docSentenceExtract});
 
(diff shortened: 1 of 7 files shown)


FooSoft:
Good work -- I agree that nowadays you don't have to work around JS differences in common browsers, so jQuery is unnecessarily heavy. We also have control over what browsers can actually install the extension, so that furthers the point.

PR #225, opened 2019-09-28
TextSourceRange improvements

toasted-nutbread:
Simplified the algorithm used for ```TextSourceRange.seekForward``` and ```TextSourceRange.seekBackward```.

Changes:
* Recursion is no longer used, and a single while loop controls iteration across nodes. This functionality is using the ```TextSourceRange.getNextNode``` function I implemented for getting the nodes in a ```Range```.
* Fixed a bug in ```TextSourceRange.getNextNode``` which was null checking the wrong variable.
* The ```shouldEnter``` function had a few optimizations/simplifications as well, the most notable one being that the ```Array.includes``` check was changed to a switch statement since it's faster and doesn't require a temporary array.

Differences:
* Speed of ```seekForward```/```seekBackward``` can be around 10x faster, but this will depend on the length that it's trying to scan.
* Minor difference in how furigana is handled. In the old version, when the scanning cursor is over ruby text, the content text will start at the character _after_ the ```</rt>``` tag. The updated algorithm will start scanning at the start of the full ```<ruby>``` tag.

In my testing, these were the only differences I found. If you think there may be anything else I may have overlooked, let me know.

#189

### DIFF
--- a/ext/fg/js/source.js
+++ b/ext/fg/js/source.js
@@ -95,12 +95,10 @@ class TextSourceRange {
         }
 
         const style = window.getComputedStyle(node);
-        const hidden =
+        return !(
             style.visibility === 'hidden' ||
             style.display === 'none' ||
-            parseFloat(style.fontSize) === 0;
-
-        return !hidden;
+            parseFloat(style.fontSize) === 0);
     }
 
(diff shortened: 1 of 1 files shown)


FooSoft:
Great work :+1:

PR #229, opened 2019-09-29
Database optimizations

toasted-nutbread:
This change adds support for bulk database lookups and a few other smaller optimizations. A majority of the work came from restructuring some of the code to support bulk lookups, so the changes are listed below.

**Database changes:**
* In database.js, two new bulk functions were added: ```findTermsBulk``` and ```findTermMetaBulk```. These correspond to the old ```findTerms``` and ```findTermMeta```, the difference being that they take an array of terms instead of a single term.
* In order to get the maximum performance, the database lookups are implemented using native IndexedDB API calls. I found that this gave increased performance over Dexie in all browsers, particularly Edge. Edge has some sort of issue with Dexie where ```db.table.getAll()``` will hang before resolving the promise; updating Dexie did not fix that issue.

**Deinflector changes:**
* The way that the deinflector was implemented was not not suitable for bulk database lookups. The reason was that the deinflector runs the database lookups _as_ it was finding deinflections, rather than saving them all and returning them at the end. The API was changed to return an array of all deinflections.
* The Deinflection class is no longer needed. The list of deinflections can be returned using a single function without recursion, which should be overall faster.

**Translator changes:**
* translator.js had some changes to accommodate the new database bulk functions and deinflection changes.
* Fewer redundant lookups should be performed for deinflections. I think this is not an extremely common case though. As a contrived example: ```すル``` would previously look up ```すル, する, す, す```, where す would be looked up twice.
* A few other changes to make sure that any repeated terms would not be looked up more than once in the database. This primarily involved creating some unique sets and mappings to the original arrays in ```buildTermFrequencies```.

**Results:**
* In Firefox, the speeds I tested were about 2x faster. I tested this by building two versions of the addon, one with the current source and one with this new change, and ran them at the same time on the same words.
* I saw similar results in Chrome while testing each individually.
* In Edge, the scan times went from 1000ms+ to about 200ms, which made it tolerably usable. It could potentially fare a bit better on a full powered device also.

If there is interest in testing these timings, I have two branches which output timing information:
* [database-optimizations-timing-old](https://github.com/toasted-nutbread/yomichan/tree/database-optimizations-timing-old)
* [database-optimizations-timing-new](https://github.com/toasted-nutbread/yomichan/tree/database-optimizations-timing-new)

---

Let me know if there's anything I can clarify or test. I know the deinflector was modified pretty significantly in order to accomplish the bulk database actions; I did test it quite a bit and it seemed to function the same, but it's possible I could have overlooked something.

#189

### DIFF
--- a/ext/bg/js/database.js
+++ b/ext/bg/js/database.js
@@ -68,18 +68,7 @@ class Database {
         const results = [];
         await this.db.terms.where('expression').equals(term).or('reading').equals(term).each(row => {
             if (titles.includes(row.dictionary)) {
-                results.push({
-                    expression: row.expression,
-                    reading: row.reading,
-                    definitionTags: dictFieldSplit(row.definitionTags || row.tags || ''),
-                    termTags: dictFieldSplit(row.termTags || ''),
-                    rules: dictFieldSplit(row.rules),
-                    glossary: row.glossary,
-                    score: row.score,
-                    dictionary: row.dictionary,
-                    id: row.id,
-                    sequence: typeof row.sequence === 'undefined' ? -1 : row.sequence
-                });
+                results.push(Database.createTerm(row));
             }
         });
 
@@ -94,18 +83,7 @@ class Database {
         const results = [];
         await this.db.terms.where('expression').equals(term).each(row => {
             if (row.reading === reading && titles.includes(row.dictionary)) {
-                results.push({
-                    expression: row.expression,
-                    reading: row.reading,
-                    definitionTags: dictFieldSplit(row.definitionTags || row.tags || ''),
-                    termTags: dictFieldSplit(row.termTags || ''),
-                    rules: dictFieldSplit(row.rules),
-                    glossary: row.glossary,
-                    score: row.score,
-                    dictionary: row.dictionary,
-                    id: row.id,
-                    sequence: typeof row.sequence === 'undefined' ? -1 : row.sequence
-                });
+                results.push(Database.createTerm(row));
             }
         });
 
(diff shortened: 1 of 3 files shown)


toasted-nutbread:
Doing a bit more testing, I've ran across a few instances where some words are being deinflected differently than they were before. E.g. 涼みな is selecting all three characters, detecting it as "imperative negative". I will look into fixing this.

toasted-nutbread:
Okay so I believe I have found the issue. It looks like I was reducing the deinflections to a unique set too early and was skipping a check to verify that the lookup rules match the deinflection rules.

This check was missing:
https://github.com/FooSoft/yomichan/blob/d96d4b0658de9eeb272a57966a07a7798da09433/ext/bg/js/deinflector.js#L59-L69

This has now been added in 44119eea2c8ad4c8eed38070ef1a3ce27fa9359e:
https://github.com/FooSoft/yomichan/blob/44119eea2c8ad4c8eed38070ef1a3ce27fa9359e/ext/bg/js/translator.js#L240-L246

I will continue to do some more testing over the week.

FooSoft:
This is amazing work, @toasted-nutbread ! 

When I do deinflection testing, I usually try really long conjugation chains (something like 食べられませんでした), and make sure that the deinflection paths I get back from Yomichan makes sense. 食べる is also interesting for testing because JDICT has 食べられる as an unique entry for it. 

I'm going to run some tests of my own on this and see if I can find any issues. Deinflection was generating tons of database queries, so it's awesome that you are reducing the number of queries that have to be made :+1:

FooSoft:
Looks good!

PR #238, opened 2019-10-05
Deinflector optimization

toasted-nutbread:
This change addresses two things:

1. The deinflection rules added in #235 weren't specific enough and were generating false positives. Invalid terms like 着かないで and 着かないでない were being deinflected incorrectly. This change improves the specificity of the deinflection rules, as well as adds support for the とる contraction which I previously thought would be more difficult. This change fixes all the false positives that I was seeing.
2. Improves the performance of ```Deinflector.deinflect``` by somewhere in the range of 2x to 5x. Previously, the deinflect algorithm would do an intersection test on the rule arrays to see if there was any overlap. This has been simplified to use a bitwise ```&``` check instead, after first converting the rules arrays into a single flags integer value.

   This is similar to the algorithm I saw in [rikaichamp](https://github.com/birtles/rikaichamp) when creating #199, except I have not changed the JSON file structure. Instead, the JSON data is converted to an array-based structure, which should also result in faster access than object property lookup (i.e. ```kanaIn = array[0];``` rather than ```kanaIn = object.kanaIn;```, and so on). This structure also converts the rules arrays into flags integer values.

   #189

### DIFF
--- a/ext/bg/js/deinflector.js
+++ b/ext/bg/js/deinflector.js
@@ -19,51 +19,74 @@
 
 class Deinflector {
     constructor(reasons) {
-        this.reasons = reasons;
+        this.reasons = Deinflector.normalizeReasons(reasons);
     }
 
     deinflect(source) {
         const results = [{
             source,
             term: source,
-            rules: [],
+            rules: 0,
             definitions: [],
             reasons: []
         }];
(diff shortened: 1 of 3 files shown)


FooSoft:
Sounds good -- I'll push this out to testing to get some more people to hammer on this change.

Issue #227, opened 2019-09-29
Using shortcuts causes tabs to change


oudajosefu:
I've noticed this problem before when enabling some options in the settings page: after activation, the browser would shift to another tab for no reason.

After configuring the shortcut to enable/disable yomichan, I noticed every time I do so, the same problem happens, the tabs change.

Don't really know if this is just me, but I assume the problem is reproducible. I am using the latest Chrome version on macOS High Sierra, if that helps.

Edit: It seems when I hold down the shortcuts, the tabs just start to spazz out and change very quickly.

oudajosefu:
Reinstalling and refreshing actually seemed to fix it for me. I'll update if it happens again. However, I wonder if there are any telling signs in the code that might be causing this? Just curious.

PR #240, opened 2019-10-06
Fix focus

### LINKED ISSUE #239: Focus issue when switching tabs
Recent report for October 5th release; adding it here for tracking.

A bug occurs when Yomichan is loaded as follows:

* I am a tab
* I open a new tab and go to a website
* I switch back to the first tab before the second new tab finishes loading
* Firefox forcibly switches back to the second tab a few seconds after it finishes loading.

I have disabled all other add-ons and confirmed this behavior only occurs when Yomichan is loaded.  In addition it only started occurring today after the latest update was installed.

toasted-nutbread:
A change which fixes #239 (hopefully). I was not able to reproduce it on my end, but it sounds suspiciously similar to #227.

This PR makes a few changes to hopefully help clarify the how visibility changes work:

* Renamed ```Popup.setVisible``` to ```setVisibleOverride```. ```setVisibleOverride``` better reflects what this function does, since it was only used when hiding all popups when creating an Anki card which takes a screenshot. Its use was to temporarily force popups to be hidden and then revert them afterwards.

* Popup visibility state is now tracked more explicitly with two internal variables: ```Popup.visible``` and ```Popup.visibleOverride```.
  * ```visible``` indicates the preferred popup visibility when no override is specified.
  * ```visibleOverride``` corresponds to the value set by ```setVisibleOverride```. This also takes a value of ```null```, indicating no override.

  All popup visibility changes are now performed by changing ```style.visibility```, rather than using both ```style.visibility``` _and_ ```style.display```.

* The ```Popup.hide``` method has been changed such that it does nothing if the popup is not visible. This should fix the issue which was causing focus to change, as this could be triggered on page load.

* The ```window.onresize``` handler has been changed to only passively hide the popup if it is open. If a ```resize``` event happens on a window/tab without focus, this could potentially cause the focused tab to change.

Overall, I'm still not sure why browsers would let web-extensions change the focused tab using non-web-extension APIs; seems exploitable.

### DIFF
--- a/ext/fg/js/frontend.js
+++ b/ext/fg/js/frontend.js
@@ -145,7 +145,7 @@ class Frontend {
     }
 
     onResize() {
-        this.searchClear(true);
+        this.searchClear(false);
     }
 
     onClick(e) {
--- a/ext/fg/js/frontend.js
+++ b/ext/fg/js/frontend.js
@@ -527,8 +527,8 @@ Frontend.runtimeMessageHandlers = {
         self.updateOptions();
     },
 
-    popupSetVisible: (self, {visible}) => {
-        self.popup.setVisible(visible);
+    popupSetVisibleOverride: (self, {visible}) => {
+        self.popup.setVisibleOverride(visible);
     }
 };
 
--- a/ext/fg/js/popup-proxy-host.js
+++ b/ext/fg/js/popup-proxy-host.js
@@ -41,7 +41,7 @@ class PopupProxyHost {
             show: ({id, elementRect, options}) => this.show(id, elementRect, options),
             showOrphaned: ({id, elementRect, options}) => this.show(id, elementRect, options),
             hide: ({id, changeFocus}) => this.hide(id, changeFocus),
-            setVisible: ({id, visible}) => this.setVisible(id, visible),
+            setVisibleOverride: ({id, visible}) => this.setVisibleOverride(id, visible),
             containsPoint: ({id, x, y}) => this.containsPoint(id, x, y),
             termsShow: ({id, elementRect, writingMode, definitions, options, context}) => this.termsShow(id, elementRect, writingMode, definitions, options, context),
             kanjiShow: ({id, elementRect, writingMode, definitions, options, context}) => this.kanjiShow(id, elementRect, writingMode, definitions, options, context),
@@ -103,9 +103,9 @@ class PopupProxyHost {
         return popup.hide(changeFocus);
     }
 
-    async setVisible(id, visible) {
+    async setVisibleOverride(id, visible) {
         const popup = this.getPopup(id);
-        return popup.setVisible(visible);
+        return popup.setVisibleOverride(visible);
     }
 
     async containsPoint(id, x, y) {
--- a/ext/fg/js/popup-proxy.js
+++ b/ext/fg/js/popup-proxy.js
@@ -65,9 +65,9 @@ class PopupProxy {
         return await this.invokeHostApi('hide', {id: this.id, changeFocus});
     }
 
-    async setVisible(visible) {
+    async setVisibleOverride(visible) {
         const id = await this.getPopupId();
-        return await this.invokeHostApi('setVisible', {id, visible});
+        return await this.invokeHostApi('setVisibleOverride', {id, visible});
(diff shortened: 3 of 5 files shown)


FooSoft:
Thanks for looking into this so quickly!

This is something I've been thinking about for a bit, but seeing as you are very active in the development of this extension, I've sent you an invite to officially become a project collaborator. I think you possess both a great vision for this extension (so far all of the things you've done made a lot of sense), as well as the technical skills to make it happen :smile:

toasted-nutbread:
Cool, I'll accept that invite here soon. A few questions below on what you think the best practices regarding contributions:
* Is there a preferred way you (or other contributors) would like to see contributions made? Creating pull requests, or pushing features to master?
* Thus far I've generally just been addressing some issues/feature requests that I have found interesting, as well as adding features that I use. Would you (or other contributors) like any sort of prior notice about new features I may work on, similar to what I did for the performance optimizations issue?
* Finally, are there other issues/features/ideas that are of interest to you that perhaps you haven't had time to look into?

FooSoft:
@toasted-nutbread probably the best way to contribute is just to push features to master, and create issues to discuss or track large features (like you've done with performance optimization). I think most of the things you've contributes are just objectively good for the project, and you've raised questions where a discussion was needed, or there was some sort of trade-off (for instance complexity cost vs. benefit of having recursive definition lookups). If something is more of a fix, or a workflow improvement, just go for it. If it's a significant feature (especially if it needs to go in over several commits), a tracking issue might be good.

One thing that would be awesome to address/document is all of the wonkiness related to IndexedDB in Firefox. It seems pretty unreliable, and according to some users is sometimes just wiped with seemingly no reason. Is there some sort of manifest tagging we can do to try to make this data more permanent? Is it a disk usage issue? If there are no workarounds, can we at least detect this error state and tell the user about it (so that it doesn't look like Yomichan just stopped working?)

toasted-nutbread:
Cool, I'll start following your recommendation. Is there anything I should do if there are features/fixes which are ready to be included in a release? I could start including/updating a changelog file if that would be helpful.

>One thing that would be awesome to address/document is all of the wonkiness related to IndexedDB in Firefox.

I actually have looked into this briefly, checking to see if the [```unlimitedStorage```](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/permissions#Unlimited_storage) permission would cause the IndexedDB to not be deleted. Unfortunately, it doesn't prevent deletions. And I can also confirm that if Firefox doesn't have enough space, it will just delete your data without warning. I have intentionally caused this to happen on PC, and the issue is also present on Firefox for Android when running low on storage, which is likely to be even more frequent.

I'll see if I can look into this some more at some point, since it is very misleading that there is no indication as to what happened

FooSoft:
For releases, it's generally just a game of trying to keep the risk of breakage to a minimum, while at the same making sure that `master` does not get too far ahead of `stable` (or `testing` for that reason). I try to make things go through testing most of the time, but the focus thing I ended up just pushing to `stable` (with a bunch of other things), as the code has seemed to have diverged enough that just letting it all go through together seemed like a smaller risk of breakage.

Regarding having some sort of roll-out plan -- you can just email me at alex@foosoft.net, if you'd like to see something pushed to `testing` or `stable`, otherwise I'll try to just try to promote branches during weekends on weeks where there have been commits (so `master` will go to `testing` and `testing` will go to `stable`). That should hopefully keep the branches from diverging too much.

Regarding IndexedDb, it might we worthwhile to see if there has been anything interesting going on in <https://dexie.org/> (Yomichan uses a pretty old version), but I kind of doubt it. Deleting dictionaries was always something that seemed like it would be cool to do, but my efforts were kind of disappointing. Deleting IndexedDb records was absurdly slow (much slower than creating new ones), and left Chrome in a weird state where subsequent queries were much slower...

One thing that might be possible to catch the database being nuked by Firefox is just to have a hidden option that is like "have dictionary imported", which will get set when you import a dictionary and cleared when you nuke the database manually. If database gets nuked without user pressing the button, then it probably means that Firefox did it by itself, and we can tell the user about it. Maybe we can also warn people if they don't have a lot of space available?

Issue #249, opened 2019-10-08
[Feature Request] Add option to enable dark mode


oudajosefu:
Just a suggestion: a dark mode for reading at night to be easier on the eyes.

I will be implementing this in the popup only for now, options page will come later. If anyone else wants to give it a shot too, please do because I'm not as skilled as the main developers of Yomichan.

oudajosefu:
Until I add this ability (which might take some time due to my schedule), anyone wanting this now can just add this custom css to the custom css popup input box in the options page:
```css
body {
    color: #fff;
    background: #000;
}
.expression .kanji-link {
    color: rgb(191, 191, 191);
}
.glossary-item {
    color: #fff;
}
```

toasted-nutbread:
Feature added with #253.

Issue #242, opened 2019-10-06
Custom CSS user instructions


oudajosefu:
A while ago, I noticed that you can add custom CSS to Yomichan, but I still can't manage to get it working. I think a little section in the options page that dictates its usage would be insightful enough for anyone trying to add a little flare to their own Yomichan.

With this new feature, I've been wondering if it's possible to achieve an inversion of colors to simulate a dark mode. Please let me know if this possible and how it would be done.

toasted-nutbread:
This is a feature I added to provide a quick and dirty way to make some tweaks to the internals of the popup. Technically you should be able to adjust any of the visual styles, but it doesn't have the best workflow for doing so. I may need to add some sort of live view of a popup to show what the results look like, because currently it requires a full page refresh of any webpage using yomichan before the differences will be shown.

Adding a dark mode using the current settings is entirely possible assuming you are familiar with CSS rules and the underlying HTML structure, it just isn't a great workflow currently. The only things you won't be able to change are styles which are outside of the popup, for example the drop shadow is currently not able to be adjusted. I've been considering experimenting with [shadow DOMs](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM) to potentially provide more options here, but I can't say for sure how well that would work.

oudajosefu:
Figured it out. This is a nice solution to having a dark mode:
```
body {
    color: #fff;
    background: #000;
}
.expression .kanji-link {
    color: rgb(191, 191, 191);
}
.glossary-item {
    color: #fff;
}
```

oudajosefu:
I'm actually gonna try implementing this as a toggle on the options page, so I will keep this open for now until I do so. Or maybe I should open another issue for that?

PR #251, opened 2019-10-11
Dark mode in popup using customPopupCss

oudajosefu:
This seems quick and dirty but I believe it gets the job done reasonably enough. I can see other more logical ways of implementing dark mode such as directly changing the css itself instead of appending darkModeCss to @toasted-nutbread's customPopupCss.

As @toasted-nutbread said in #242, it would be nice to have "some sort of live view of a popup to show what the results look like, because currently it requires a full page refresh of any webpage using yomichan before the differences will be shown."

This satisfies #249, I hope.

### DIFF
--- a/ext/bg/js/options.js
+++ b/ext/bg/js/options.js
@@ -262,6 +262,7 @@ function profileOptionsCreateDefaults() {
             resultOutputMode: 'group',
             debugInfo: false,
             maxResults: 32,
+            darkMode: false,
             showAdvanced: false,
             popupDisplayMode: 'default',
             popupWidth: 400,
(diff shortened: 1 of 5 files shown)


toasted-nutbread:
FYI: I am in the process of implementing a live style preview on the settings page, as mentioned in #242. With this change, I will probably also include a "popup theme" option similar to what you've currently done. It may include some additional styling tweaks which couldn't be easily done without some other architecture changes (#252).

I'll create it as a PR and tag you on it to let you know when it's done. Please wait warmly.

toasted-nutbread:
Closing now that the style editor and its themes has been merged in. Feel free to open another issue or PR if you would like additional changes.

PR #253, opened 2019-10-12
Popup style preview + themes

toasted-nutbread:
This change adds support for previewing some of the styles of the popup from the settings page. This allows custom CSS changes to be visualized in real time.

* Popup preview added to settings page. The background on this page can be switched from dark to light to visualize the popup on different backgrounds.
* Bootstrap styles removed from float.html. Very few were actually used, so this may allow the page to load (very slightly) faster.
* Added support for popup themes. Currently includes the default light theme and a new dark theme.
* Added support for popup shadow themes. Since the box shadow for the popup is outside of the iframe, custom CSS can't be applied in the same way. Three themes are supported: ```default```, ```dark```, and ```auto```. ```auto``` will select the best shadow color based on the page's background color. I.e. if the page is light-ish, the black shadow will be used; if the page is dark-ish, the white shadow will be used.

---

image

---

image

### DIFF
--- a/ext/bg/css/settings.css
+++ b/ext/bg/css/settings.css
@@ -148,6 +148,15 @@ input[type=checkbox]#storage-persist-button-checkbox {
     padding: 0;
 }
 
+#settings-popup-preview-frame {
+    background-color: transparent;
+    border: none;
+    margin: 0;
+    padding: 0;
+    width: 100%;
+    height: 320px;
+}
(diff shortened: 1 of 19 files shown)


toasted-nutbread:
FYI @oudajosefu / #251

FooSoft:
Looks rad :+1:

PR #259, opened 2019-10-19
Database optimizations part 2

toasted-nutbread:
Optimization of the remainder of the database calls. Completes the database read optimizations started in #229.

* Bulk operations are now used for all database reads, all of which use native IndexedDB APIs for maximum speed.
* Reduced complexity of ```Translator.findTermsMerged```'s function body by creating some helper functions which do portions of the work. This should also make it easier for Javascript compilers to optimize (typically they're bad at dealing with very large functions).
* Moved the tag cache out of ```Database``` and into ```Translator```. This makes it so that ```Database``` doesn't have to worry about caching, only requesting and updating data.
* A few other simplifications to ```sort``` calls and other inline functions.

This change mostly affects the speed when the result grouping method is set to "Group results by main dictionary entry".

#189

### DIFF
--- a/ext/bg/js/database.js
+++ b/ext/bg/js/database.js
@@ -182,7 +182,12 @@ class Database {
         this.validate();
 
         let result = null;
-        await this.db.tagMeta.where('name').equals(name).each(row => {
+        const db = this.db.backendDB();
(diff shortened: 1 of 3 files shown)


FooSoft:
Nice work!

PR #261, opened 2019-10-20
Reuse open search tab when clicking search button

### LINKED ISSUE #260: Move focus to already opened search page when hotkey is pressed
Instead of opening a whole new search page, Yomichan could check all tabs to see if the search page was already opened and shift the focus to that tab if it exists. I dislike having all these search pages open when I'm constantly searching words that I hear.

toasted-nutbread:
Resolves #260.

* Middle clicking buttons opens in a new tab, even if a tab already exists.
* Settings button has been updated to have the same type of functionality.


siikamiika:
Nice :+1: I approve this functionality wise, but I would also add a tooltip to the buttons indicating what happens when you middle-click them

Issue #270, opened 2019-10-29
Database improvements


toasted-nutbread:
This is intended to be an overview of various issues and tasks related to the database, as well as a venue for general discussion.

**Database modifying operations**
* [x] Use [```beforeunload```](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event) event to warn if the settings page is being closed while a database change operation is occurring (import, purge, delete, etc.). #272
   _This should help prevent corruptions/lingering data._
* [x] Add ability to update dictionary. #86, #272
   _This can likely be accomplished by implementing a "Delete Dictionary" option._
* [x] Import multiple dictionaries at once. #269, #272

**Database health validation**
* [x] Detect when databases have been automatically purged by the browser. #100
   _This may be preventable with the right permissions (#256, #254), but a check should still be implemented._
* [x] Show information about lingering data in the event of a partial database deletion. #272
   _E.g. the entry from the "dictionaries" object store has been deleted, but the terms/kanji definitions still exist._
* [x] Some tweaks to how the settings page displays dictionary information. #272
   _In order to make it easier to more robustly show the information mentioned above._
* [x] Indicate when the backend / database is upgrading. When the database version is upgrading, the extension won't run. This operation can take multiple seconds, so it can be noticeable. This was the case in the version upgrade in #287, despite the difference being very small.
* [x] Sanitization of dictionary data during the import process.

**Extended database functionality**
* [x] Show entries that match the selected expression plus wildcards. #128, #274
   _This may be difficult, since all IndexedDB has to work with is [```IDBKeyRange.bound```](https://developer.mozilla.org/en-US/docs/Web/API/IDBKeyRange/bound). It may be possible to use this to some extent, but it is very limited. Maybe only enabling on the search page since it's likely to be a more expensive operation._
* [x] Custom dictionary with images. #322
* [x] Documentation of the dictionary formats. #322
* [x] Pitch accents. #61, #131

**Other**
* [x] Use IndexedDB APIs directly for everything, rather than using dexie. #278
   _This one is certainly up for debate, but from what I've seen, dexie largely obscures how IndexedDB works, which causes sub-optimal performance since it's not a 1-to-1 mapping. These performance issues have largely been addressed for real-time lookups, but now the ```Database``` class is in this weird state where half of it uses dexie and half of it uses the IndexedDB API. Might be better to just make everything use IndexedDB APIs for consistency._
[...]

Issue #128, opened 2018-06-16
[Feature Request] Show entries that match the selected expression plus wildcards


rgamici:
When I add word to Anki, I usually also add some related words, as they help grasp the distinction between synonyms, and also to help reinforce that new word in other possible uses.
For example, on Jisho I can add the wildcards '*' before and/or after the word I'm looking for and it will return other expressions that use that (sub)word.

Is it feasible to add this feature on yomichan?
For example, show those 'similar' result on the list of the results.

toasted-nutbread:
Looking into this a bit while working on #270, it does seem that this is partially possible.

For suffix wildcards, this is possible by making clever use of how strings are sorted. The following can be done using [```IDBKeyRange.bound```](https://developer.mozilla.org/en-US/docs/Web/API/IDBKeyRange/bound):
```js
const term = 'あ'; // example
const query = IDBKeyRange.bound(term, `${term}\uffff`, true, false);
```
Using this query will find all terms beginning with ```'あ'``` (non-inclusive in this case).

The same cannot be done for prefix wildcards due to how strings are sorted. However, it would potentially be possible if each dictionary entry also stored a string reversed version of the search term. This would require re-importing all dictionaries, and the database size would be slightly larger.

The other factor in supporting wildcards is how to show them. Wildcards on the search page could be fairly straight forward: perform no deinflection and only search for terms starting with the exact text before the wildcard. In the popup, wildcards are not as clear-cut with regards to how they could be handled. They should be off by default, but I'm not sure what the best way to incorporate the feature would be, or if it would even make sense to include it.

rgamici:
Thank you for taking my request in consideration.

> The same cannot be done for prefix wildcards due to how strings are sorted.

I'm not sure how useful is to have the prefix wildcard compared to the suffix one, but I would guess at least 30 % of the total.
Since I've never experienced yomichan reach the limits of available memory both on HDD or RAM, I think a *slightly* increase would be imperceptible.
Do you have an approximate estimation of how bigger it would be considered the current database size?

If the bottleneck is the time necessary to process the request, maybe another modifier key could be used to activate the wildcard search.
For example, currently the `shift` key is used to search a word, and then `ctrl+shift` could be used to do the wildcard search.
I'm not sure how easy it would be to add this, but it would leave normal users unaffected.

I've currently installed the dictionaries JMDict, KireiCake, and JMnedict.
The first two sometimes miss an entry or have them split, and having both usually gets a slightly better result than have only one of them.
[...]

PR #274, opened 2019-11-05
Lookup wildcards

toasted-nutbread:
Adds support for suffix wildcards on the search page. This adds partial support for #128.

Tracked in #270.

---

image


siikamiika:
Maybe `findTermsWildcard` should use a new database method that has a limit, and skip sorting altogether to enable pagination

toasted-nutbread:
I think imposing a limit of the number of results returned from the search is probably a good idea, especially with how many results can be returned using the ```getAll``` methods. Pagination presents an issue in that you can't really give the indexedDB API a cursor to start searching at. Might be best done as something like "Show more results" instead of something like Page 1, 2, 3, etc. I'll if this works well.

siikamiika:
I was thinking that you could give a new lower and upper bound when going to the next page. For example:
- Look up `あ＊` (`IDBKeyRange.bound('あ', 'あ\uffff')`)
- while `options.general.maxResults` hasn't been reached:
  - Fetch the first `options.general.maxResults * dictionary count etc.` results
  - Process the results based on split/group/merge settings and do additional exact DB searches if needed
  - Store the word (for example あか) that made the amount of results reach `options.general.maxResults` and `break`
- Show results along with a button that does `IDBKeyRange.bound('あか', 'あか\uffff')` but skips the words that were the immediate results (not merge results) on the previous pages

Going back is probably too hard to do in the DB unless you cache the pages. I agree that it would also be difficult to count the amount of pages there will be unless you fake it and don't show the number of the last page.

FooSoft:
Looks good to me :+1:

toasted-nutbread:
So a bit of a follow up to @siikamiika's comments: I haven't completely tested everything, but I think some of the asks would involve significant changes to the code, potentially adding complexity. A few factors below:

* Supporting all three values for ```resultOutputMode``` (```'group', 'merge', 'split'```) for wildcards, some of which do their own sorting/grouping.
* The usage of ```dictTermsUndupe```, which operate on the full set of results.
* ```options.general.maxResults * dictionary count``` is not necessarily accurate, especially with frequency dictionaries. This is compounded by how the lookup has to do 2x the searches: one using reading, and one using expression.

Another thing I noticed: It is possible to sort the results by ordering by the expression/reading string. However, the current method may sort the more "relevant" results first. For example, searching ```aga*``` using the current method yields ```上がる```,  but when sorting using the string, the first result is ```あがぁな```. Not only is the latter potentially less interesting, jisho.org also yields [```上がる```as the first resul](https://jisho.org/search/aga*)t. So while the results aren't identical, it shows that a similar search on a dictionary website doesn't sort results in string-sort order.

PR #277, opened 2019-11-09
Remove Download check when resolving Audio data

sabslikesobs:
There is a bug (seemingly unreported) in Yomichan where an Anki card
will not contain any audio if the JapanesePod101 audio source trumps
a secondary audio source (e.g. JapanesePod101-alternate) where
the jpod101 source can't find the word requested.

For example, そして has an audio entry in the alternate source but not
the standard source, so Anki cards created from そして won't have an
audio field set.

(Alternatively, there may be a bug in the jpod101 audioUrlBuilder,
because I've only noticed this problem with hiragana-only expressions.
JPod101 may not host those on the same url scheme any more.  I'm not
sure how to fix that, though, and the bug I'm addressing here does
still persist).

The reason this happens is that audioGetFromUrl uses downloaded audio
to effectively check for a 404 (by examining the audio duration), but
that check doesn't happen when an Anki card is being created (i.e.
"download" is set, which I've changed to "willDownload" here). So it happily
reports that the first URL it creates is the correct download URL, even if that
URL would have been passed over if listened to normally.

This change removes that check, but retains the will-download intent
information to prevent attempts to download tts data, which AnkiConnect
cannot do. I've also added a short explanation as to why the download
check happens where it does.

I think the unused audio object will get garbage collected since it's
not referenced again, but I've explicitly unset it as well.

### DIFF
--- a/ext/mixed/js/audio.js
+++ b/ext/mixed/js/audio.js
@@ -88,17 +88,13 @@ class TextToSpeechAudio {
 
 }
 
-function audioGetFromUrl(url, download) {
+function audioGetFromUrl(url, willDownload) {
     const tts = TextToSpeechAudio.createFromUri(url);
     if (tts !== null) {
-        if (download) {
-            throw new Error('Download not supported for text-to-speech');
+        if (willDownload) {
+            throw new Error('AnkiConnect does not support downloading text-to-speech audio.');
         }
         return Promise.resolve(tts);
-    }
-
-    if (download) {
-        return Promise.resolve(null);
     }
 
     return new Promise((resolve, reject) => {
@@ -115,7 +111,7 @@ function audioGetFromUrl(url, download) {
     });
 }
 
-async function audioGetFromSources(expression, sources, optionsContext, download, cache=null) {
+async function audioGetFromSources(expression, sources, optionsContext, willDownload, cache=null) {
     const key = `${expression.expression}:${expression.reading}`;
     if (cache !== null && cache.hasOwnProperty(expression)) {
         return cache[key];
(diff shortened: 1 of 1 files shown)


toasted-nutbread:
This behaviour is not exactly a bug, as I purposefully designed it to function this way, but the motivation behind doing so is obsolete. The rationale behind originally only checking the primary audio source for Anki downloads was because Yomichan used to have four audio sources enabled by default. If audio is not available on any source, Yomichan will have to download each audio to verify it is not a 404. This would make the addition of a card take longer, especially for terms without audio from any source.

However, the default audio sources has since been reduced to just one source due to similar delay issues when playing audio. Therefore, download times are no longer really an issue since it's something the user has to opt in to.

All of your changes look good to me, but I'm going to use this as an opportunity to bring something up to @FooSoft:

It seems that there is a redundant download occurring when cards with audio are added using AnkiConnect. Yomichan first has to verify that the audio is not a 404 before instructing AnkiConnect to download it. AnkiConnect then downloads the file again and stores this on the file system.

Do you think it would be beneficial to have Yomichan download and check the audio files (as a byte array) and then instruct Yomichan to create a new data file? The [screenshot](https://github.com/FooSoft/yomichan/pull/182) functionality already does this, so it is possible. I've not done any timings or testing related to this idea, so I'm not sure how useful/beneficial it would be, but it could eliminate an extra network request.

FooSoft:
Yes, I think that would be ideal (especially, as you pointed out, that it is how screenshotting works). It would be ideal to have all multimedia be treated the same way for consistency, and it helps avoid the (unlikely but annoying) case of the audio failing to download for some reason on the second request.

toasted-nutbread:
Cool, then I may look into that at some point.

PR #278, opened 2019-11-10
Dexie removal

toasted-nutbread:
The intent of this change is to have consistency between all of the database APIs, rather than having a mix of dexie and native IndexedDB. Since dexie doesn't reach the maximum lookup speeds that are possible when using IndexedDB, the preference would be for native IndexedDB only.

Overall, it shouldn't be significantly more complicated with regards to maintaining, there should be just the initial time overhead of setting up the functions. Everything is still abstracted using promises. The [database upgrade code](https://github.com/FooSoft/yomichan/blob/aef0bd81cd77a5cdeac1ad67d7feaeb157adfc51/ext/bg/js/database.js#L31-L81) is also abstracted similarly to how dexie handled it, but it's a bit more verbose to avoid having to implement a string parser.

#270


FooSoft:
Looks nice and clean.

I was honestly assuming that moving everything to IndexedDB would add a lot of boilerplate; that does not look to be the case at all :+1:

PR #279, opened 2019-11-10
Search page query parser

siikamiika:
I'm opening a pull request because I need a few comments from other developers before I am ready to merge this.

This pull request adds a text parser to the search page that uses two different methods.

The first method that is enabled by default uses the built-in scanning so that it takes the longest result and matches readings from the best match using the score-based sorting and a new furigana function.

The other method is more problematic, as it uses native messaging to parse the text in an external application, MeCab. Here's a list of concerns:

- Should the protocol between the MeCab native host and Yomichan be versioned, and if it should, how? I'm planning to add part of speech highlighting in the future when I have investigated MeCab IPADIC part of speech tag names. I already have a working implementation in my previous project using my custom version of the Unidic dictionary.
- How should the native application be distributed? Currently I have a link on the settings page to https://github.com/siikamiika/yomichan-mecab-installer where I have installation instructions. I opted to bundle the Windows binary with the repository because I couldn't figure out how to otherwise add it to `%PATH%` safely, and I'm not sure if there's an "official" binary distribution of it for Windows where build environment setup is more difficult.

Few other questions about the code.

- Should `QueryParser` be a subclass of `Frontend`? It works quite dfiferently so I just copied the required functionality from other places of code.
- Are we transitioning away from Handlebars and should the text parsing view be implemented using `<template>` instead?

### DIFF
--- a/ext/bg/js/backend.js
+++ b/ext/bg/js/backend.js
@@ -180,6 +180,7 @@ Backend.messageHandlers = {
     optionsSet: ({changedOptions, optionsContext, source}) => apiOptionsSet(changedOptions, optionsContext, source),
     kanjiFind: ({text, optionsContext}) => apiKanjiFind(text, optionsContext),
     termsFind: ({text, details, optionsContext}) => apiTermsFind(text, details, optionsContext),
+    textParse: ({text, optionsContext}) => apiTextParse(text, optionsContext),
     definitionAdd: ({definition, mode, context, optionsContext}) => apiDefinitionAdd(definition, mode, context, optionsContext),
     definitionsAddable: ({definitions, modes, optionsContext}) => apiDefinitionsAddable(definitions, modes, optionsContext),
     noteView: ({noteId}) => apiNoteView(noteId),
--- a/ext/bg/js/backend.js
+++ b/ext/bg/js/backend.js
@@ -101,6 +101,8 @@ class Backend {
 
         if (options.parsing.enableMecabParser) {
             this.mecab.startListener();
+        } else {
+            this.mecab.stopListener();
         }
     }
 
--- a/ext/bg/js/backend.js
+++ b/ext/bg/js/backend.js
@@ -98,6 +98,10 @@ class Backend {
         }
 
         this.anki = options.anki.enable ? new AnkiConnect(options.anki.server) : new AnkiNull();
+
+        if (options.parsing.enableMecabParser) {
+            this.mecab.startListener();
+        }
     }
 
     async getFullOptions() {
--- a/ext/bg/js/backend.js
+++ b/ext/bg/js/backend.js
@@ -21,6 +21,7 @@ class Backend {
     constructor() {
         this.translator = new Translator();
         this.anki = new AnkiNull();
+        this.mecab = new Mecab();
         this.options = null;
         this.optionsContext = {
             depth: 0,
@@ -181,6 +182,7 @@ Backend.messageHandlers = {
     kanjiFind: ({text, optionsContext}) => apiKanjiFind(text, optionsContext),
     termsFind: ({text, details, optionsContext}) => apiTermsFind(text, details, optionsContext),
     textParse: ({text, optionsContext}) => apiTextParse(text, optionsContext),
+    textParseMecab: ({text, optionsContext}) => apiTextParseMecab(text, optionsContext),
     definitionAdd: ({definition, mode, context, optionsContext}) => apiDefinitionAdd(definition, mode, context, optionsContext),
     definitionsAddable: ({definitions, modes, optionsContext}) => apiDefinitionsAddable(definitions, modes, optionsContext),
     noteView: ({noteId}) => apiNoteView(noteId),
--- a/ext/bg/js/mecab.js
+++ b/ext/bg/js/mecab.js
@@ -60,4 +60,4 @@ class Mecab {
     }
 }
 
-Mecab.timeout = 1000;
+Mecab.timeout = 5000;
--- a/ext/bg/js/mecab.js
+++ b/ext/bg/js/mecab.js
@@ -34,15 +34,7 @@ class Mecab {
     startListener() {
         if (this.port !== null) { return; }
         this.port = chrome.runtime.connectNative('yomichan_mecab');
-        this.port.onMessage.addListener((message) => {
-            const {sequence, data} = message;
-            const {callback, timer} = this.listeners[sequence] || {};
-            if (timer) {
-                clearTimeout(timer);
-                delete this.listeners[sequence];
-                callback(data);
-            }
-        });
+        this.port.onMessage.addListener(this.onNativeMessage.bind(this));
     }
 
     stopListener() {
@@ -51,6 +43,15 @@ class Mecab {
         this.port = null;
         this.listeners = {};
         this.sequence = 0;
+    }
+
+    onNativeMessage({sequence, data}) {
+        if (this.listeners.hasOwnProperty(sequence)) {
+            const {callback, timer} = this.listeners[sequence];
+            clearTimeout(timer);
+            callback(data);
+            delete this.listeners[sequence];
+        }
     }
 
     invoke(action, params) {
--- a/ext/bg/js/mecab.js
+++ b/ext/bg/js/mecab.js
@@ -25,7 +25,7 @@ class Mecab {
     }
 
     onError(error) {
-        logError(error, true);
+        logError(error, false);
     }
 
     async checkVersion() {
--- a/ext/bg/js/mecab.js
+++ b/ext/bg/js/mecab.js
@@ -70,7 +70,7 @@ class Mecab {
 
     invoke(action, params) {
         if (this.port === null) {
-            return {};
+            return Promise.resolve({});
         }
         return new Promise((resolve, reject) => {
             const sequence = this.sequence++;
--- a/ext/bg/js/mecab.js
+++ b/ext/bg/js/mecab.js
@@ -58,12 +58,10 @@ class Mecab {
             const sequence = this.sequence++;
 
             this.listeners[sequence] = {
-                callback: (data) => {
-                    resolve(data);
-                },
+                callback: resolve,
                 timer: setTimeout(() => {
                     delete this.listeners[sequence];
-                    reject(`Mecab invoke timed out in ${Mecab.timeout} ms`);
+                    reject(new Error(`Mecab invoke timed out in ${Mecab.timeout} ms`));
                 }, Mecab.timeout)
             }
 
--- a/ext/bg/js/mecab.js
+++ b/ext/bg/js/mecab.js
@@ -52,7 +52,7 @@ class Mecab {
                 timer: setTimeout(() => {
                     delete this.listeners[sequence];
                     reject(`Mecab invoke timed out in ${Mecab.timeout} ms`);
-                }, 1000)
+                }, Mecab.timeout)
             }
 
             this.port.postMessage({action, params, sequence});
--- a/ext/bg/js/mecab.js
+++ b/ext/bg/js/mecab.js
@@ -29,7 +29,7 @@ class Mecab {
     }
 
     startListener() {
-        this.port = chrome.runtime.connectNative('mecab');
+        this.port = chrome.runtime.connectNative('yomichan_mecab');
         this.port.onMessage.addListener((message) => {
             const {sequence, data} = message;
             const {callback, timer} = this.listeners[sequence] || {};
--- a/ext/bg/js/mecab.js
+++ b/ext/bg/js/mecab.js
@@ -24,10 +24,23 @@ class Mecab {
         this.sequence = 0;
     }
 
-    async parseText(text) {
-        if (this.port === null) {
-            return {};
+    onError(error) {
+        logError(error, true);
+    }
+
+    async checkVersion() {
+        try {
+            const {version} = await this.invoke('get_version', {});
+            if (version !== Mecab.version) {
+                this.stopListener();
+                throw new Error(`Unsupported MeCab native messenger version ${version}. Yomichan supports version ${Mecab.version}.`);
+            }
+        } catch (error) {
+            this.onError(error);
         }
+    }
+
+    async parseText(text) {
         return await this.invoke('parse_text', {text});
     }
 
@@ -35,6 +48,7 @@ class Mecab {
         if (this.port !== null) { return; }
         this.port = chrome.runtime.connectNative('yomichan_mecab');
         this.port.onMessage.addListener(this.onNativeMessage.bind(this));
+        this.checkVersion();
     }
 
     stopListener() {
@@ -55,6 +69,9 @@ class Mecab {
     }
 
     invoke(action, params) {
+        if (this.port === null) {
+            return {};
+        }
         return new Promise((resolve, reject) => {
             const sequence = this.sequence++;
 
@@ -72,3 +89,4 @@ class Mecab {
 }
 
 Mecab.timeout = 5000;
+Mecab.version = 1;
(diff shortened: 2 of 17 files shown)


toasted-nutbread:
> Are we transitioning away from Handlebars and should the text parsing view be implemented using ```<template>``` instead?

I have been using ```<template>``` for a few reasons:
* It's supported by the minimum version of all browsers the extension supports.
* It's a native HTML5 functionality to create a new instance. Potentially faster than HTML string parsing using the string faster constructed by handlebars.
* Most of the templates I've converted have been fairly simple, without many conditionals or loops. This is why I've largely left terms.html and kanji.html untouched.
* Faster to develop and iterate on. Doesn't require a third-party software to be installed to update them, nor does it require running a script to update the JS. All HTML can also be found and edited from the same page.

So my personal thoughts are if there is something simple, a ```<template>``` may be faster to develop with. As an example, 085881d3426c5a872739d56e286514317b2ec8f3: the motivation for converting this into a ```<template>``` was so the [anki-field-marker-template](https://github.com/FooSoft/yomichan/commit/085881d3426c5a872739d56e286514317b2ec8f3#diff-ae1e5c86819334b79790c1b8ea76fc95R723) template could be used for another purpose, so I converted model.html into two HTML ```<template>```s. If you look at the function used to instantiate the template, it's [fairly simple](https://github.com/FooSoft/yomichan/commit/085881d3426c5a872739d56e286514317b2ec8f3#diff-50f53029998c7730aee7588b7888bf9fR492-R505).

ext/bg/js/settings.js
```diff
@@ -433,6 +440,20 @@ function onMessage({action, params}, sender, callback) {
 }
 
 
+/*
+ * Text parsing
+ */
+
+function onParseMecabChanged(e) {
```
toasted-nutbread: I think this functionality should be in ```Backend.onOptionsUpdated```.

ext/bg/js/translator.js
```diff
@@ -218,9 +218,15 @@ class Translator {
...
         for (const deinflection of deinflections) {
             for (const definition of deinflection.definitions) {
-                const definitionTags = await this.expandTags(definition.definitionTags, definition.dictionary);
-                definitionTags.push(dictTagBuildSource(definition.dictionary));
-                const termTags = await this.expandTags(definition.termTags, definition.dictionary);
+                let definitionTags, termTags;
+                if (expandTags) {
+                    definitionTags = await this.expandTags(definition.definitionTags, definition.dictionary);
+                    definitionTags.push(dictTagBuildSource(definition.dictionary));
+                    definitionTags = dictTagsSort(definitionTags);
```
toasted-nutbread: ```dictTagsSort``` operates in place; assignment is redundant for ```definitionTags``` and ```termTags```.

ext/bg/js/translator.js
```diff
@@ -231,8 +237,8 @@ class Translator {
                     expression: definition.expression,
                     reading: definition.reading,
                     glossary: definition.glossary,
-                    definitionTags: dictTagsSort(definitionTags),
-                    termTags: dictTagsSort(termTags),
+                    definitionTags: definitionTags,
```
toasted-nutbread: Can be simplified:
```js
definitionTags,
termTags,
```

ext/manifest.json
```diff
@@ -42,7 +42,8 @@
         "<all_urls>",
         "storage",
         "clipboardWrite",
-        "unlimitedStorage"
+        "unlimitedStorage",
+        "nativeMessaging"
```
toasted-nutbread: May be another case for using optional_permissions for nativeMessaging, since it's a non-essential functionality.

ext/mixed/css/display.css
```diff
@@ -88,6 +88,19 @@ ol, ul {
...
+    margin-top: 10px;
+    font-size: 24px;
+}
+
+.query-parser-term {
+    margin-right: 5px;
+}
+
+html:root[data-yomichan-page=search] body {
+    min-height: 101vh; /* always show scroll bar to avoid scanning problems */
```
toasted-nutbread: I think ```overflow-y: scroll;``` accomplishes what you want.

ext/mixed/js/japanese.js
```diff
@@ -86,3 +85,29 @@ function jpDistributeFurigana(expression, reading) {
...
+    const shortest = Math.min(source.length, expression.length);
+    const sourceHiragana = jpKatakanaToHiragana(source);
+    const expressionHiragana = jpKatakanaToHiragana(expression);
+    while (stemLength < shortest && sourceHiragana[stemLength] === expressionHiragana[stemLength]) {
+        ++stemLength;
+    }
+    const offset = source.length - stemLength;
+
+    for (const segment of jpDistributeFurigana(
+        source.slice(0, offset === 0 ? source.length : source.length - offset),
```
toasted-nutbread: For improved semantics, I would recommend doing something like:

```js
const expression2 = source.slice(0, offset === 0 ? source.length : source.length - offset);
const reading2 = reading.slice(0, offset === 0 ? reading.length : reading.length - expression.length + stemLength);
for (const segment of jpDistributeFurigana(expression2, reading2)) {
```

I think expression2 can also be simplified to
```js
const expression2 = source.slice(0, source.length - offset);
```

ext/bg/js/backend.js
```diff
@@ -97,6 +98,10 @@ class Backend {
         }
 
         this.anki = options.anki.enable ? new AnkiConnect(options.anki.server) : new AnkiNull();
+
+        if (options.parsing.enableMecabParser) {
```
toasted-nutbread: Move into ```onOptionsUpdated```; there is also a related comment about how this is toggled on/off from the settings page.

ext/bg/js/mecab.js
```diff
@@ -0,0 +1,75 @@
...
+        if (this.port === null) {
+            return {};
+        }
+        return await this.invoke('parse_text', {text});
+    }
+
+    startListener() {
+        if (this.port !== null) { return; }
+        this.port = chrome.runtime.connectNative('yomichan_mecab');
+        this.port.onMessage.addListener((message) => {
```
toasted-nutbread: Consider making this handler a function defined on the Mecab class, for similarity with how other classes handle onMessage listeners. E.g. below:

https://github.com/FooSoft/yomichan/blob/6f530304693534bfac41dfbe77cbbd9c79387617/ext/fg/js/frontend.js#L63

ext/bg/js/mecab.js
```diff
@@ -0,0 +1,75 @@
...
+        return new Promise((resolve, reject) => {
+            const sequence = this.sequence++;
+
+            this.listeners[sequence] = {
+                callback: (data) => {
+                    resolve(data);
+                },
+                timer: setTimeout(() => {
+                    delete this.listeners[sequence];
+                    reject(`Mecab invoke timed out in ${Mecab.timeout} ms`);
```
toasted-nutbread: Reject with an error. A while back I changed all exceptions/rejections to use errors because it provides stack trace information.

```js
reject(new Error(`Mecab invoke timed out in ${Mecab.timeout} ms`));
```

ext/bg/js/search-query-parser.js
```diff
@@ -0,0 +1,172 @@
...
+        })();
+    }
+
+    async setText(text) {
+        this.search.setSpinnerVisible(true);
+        await this.setPreview(text);
+
+        const results = {};
+        if (this.search.options.parsing.enableScanningParser) {
+            results['scan'] = await apiTextParse(text, this.search.getOptionsContext());
```
toasted-nutbread: ```results.scan```

toasted-nutbread:
>    Should `QueryParser` be a subclass of `Frontend`? It works quite dfiferently so I just copied the required functionality from other places of code.

I think from what I saw it's probably fine as it is. It looks like it's using a few shared functions from ```Frontend```, which is already good.

siikamiika:
> > Are we transitioning away from Handlebars and should the text parsing view be implemented using `<template>` instead?
> 
> I have been using `<template>` for a few reasons:
> 
>     * It's supported by the minimum version of all browsers the extension supports.
> 
>     * It's a native HTML5 functionality to create a new instance. Potentially faster than HTML string parsing using the string faster constructed by handlebars.
> 
>     * Most of the templates I've converted have been fairly simple, without many conditionals or loops. This is why I've largely left terms.html and kanji.html untouched.
> 
>     * Faster to develop and iterate on. Doesn't require a third-party software to be installed to update them, nor does it require running a script to update the JS. All HTML can also be found and edited from the same page.
> 
> 
> So my personal thoughts are if there is something simple, a `<template>` may be faster to develop with. As an example, [085881d](https://github.com/FooSoft/yomichan/commit/085881d3426c5a872739d56e286514317b2ec8f3): the motivation for converting this into a `<template>` was so the [anki-field-marker-template](https://github.com/FooSoft/yomichan/commit/085881d3426c5a872739d56e286514317b2ec8f3#diff-ae1e5c86819334b79790c1b8ea76fc95R723) template could be used for another purpose, so I converted model.html into two HTML `<template>`s. If you look at the function used to instantiate the template, it's [fairly simple](https://github.com/FooSoft/yomichan/commit/085881d3426c5a872739d56e286514317b2ec8f3#diff-50f53029998c7730aee7588b7888bf9fR492-R505).

I've also found Handlebars quite hard to develop with because of the compilation step and difficult debugging, but I haven't tried using `<template>`s before. I'll study your examples to see if it would be good for my use case.

When I worked on a project before where I had to generate lots of elements dynamically, I got fed up with typing `document.createElement` and made a [function](https://github.com/siikamiika/language-learning/blob/25e212977e963ecfbee05056563e348cb56b7b70/chinese/cantonese/canto_parser/static/js/helpers.js#L7-L86) that turns JS objects into elements that you can append to DOM. After that I saw a library that does pretty much the same but is a function like `React.createElement` where the first argument is the tag name, but can't recall the name of the library.

ext/bg/js/search-query-parser.js
```diff
@@ -0,0 +1,172 @@
...
+        }
+        if (this.search.options.parsing.enableMecabParser) {
+            let mecabResults = await apiTextParseMecab(text, this.search.getOptionsContext());
+            for (const mecabDictName in mecabResults) {
+                results[`mecab-${mecabDictName}`] = mecabResults[mecabDictName];
+            }
+        }
+
+        const contents = await Promise.all(Object.values(results).map(async result => {
```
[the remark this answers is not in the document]
siikamiika: This part is work in progress, I was planning to add a menu where you can select which parser to show instead of showing all with `.join('<hr>')`. You're right it's a bit too complex for a `setText`.

ext/bg/js/settings.js
siikamiika: Hmm, that's true. I guess settings profiles wouldn't work with this code.

ext/manifest.json
siikamiika: Unfortunately this permission doesn't work in optional_permissions, at least on Firefox.

ext/mixed/css/display.css
siikamiika: Yes, it seems to do what I was trying to do here. However, in a situation where you have already scrolled a bit down on a long result and scan text that returns a short result, only `padding-bottom: 100vh;` or similar helps.

siikamiika:
> > Should `QueryParser` be a subclass of `Frontend`? It works quite dfiferently so I just copied the required functionality from other places of code.
> 
> I think from what I saw it's probably fine as it is. It looks like it's using a few shared functions from `Frontend`, which is already good.

Another thing that crossed my mind was making scanning a separate class that could be used in many places, but it's a too complex change for this pull request.

siikamiika:
Thank you for the code review, your suggestions were spot on and I'll start working on them next. I see you also refactored some of the code I already pushed to master for this branch.

toasted-nutbread:
The only conflict should be due to me changing the name of a variable in a catch from ```catch (e)``` to ```catch (error)```, which I thought I avoided but got a conflict anyway.

siikamiika:
That probably happened because I took the data fetching part of `onTermLookup` and moved it to a new `termLookup` method.

ext/bg/js/backend.js
siikamiika: Are you sure you meant that this `if` block should be moved to `onOptionsUpdated`? It's in `applyOptions` that's called from `onOptionsUpdated` and other things in this block do similar things.

toasted-nutbread: You're right, sorry about that; I overlooked where it was called from after seeing the ```mecab.start/stopListener``` calls in the settings.js

siikamiika:
@toasted-nutbread 
I have addressed the issues you found in the code, and rebased to master.

I thought a bit about the versioning of the MeCab protocol, and decided that I will not add versions now. The component is not supposed to do anything else but receive text and output arrays of objects where only three keys are used so far. If I add part of speech tagging in the future, I can see if the dictionary supports the feature by looking for keys with specific names.

If a new version is needed, it can be versioned using the name of the component.

FooSoft:
@siikamiika some thoughts:

* Versioning is pretty important IMO, even if it is just a single field that says "version". Even though the interface may appear to be straightforward today, I can guarantee that at some point there will need to be changes made that will either require "version guessing" or will just result in breaks. Even if there is no code for versioning, we can at least detect that there is a version mismatch and tell the user about it so that they can fix the problem.
* Regarding the bins in the distribution (for Windows), perhaps it is better to just give the users direct links to the files instead? I know that I for one, would not just download executables from an arbitrary github repo unless I can see the source for it, or hashes, or something. I've had the same thing pointed out to me about Yomichan-Import, so speaking from experience :)

siikamiika:
@FooSoft 
Thank you for the comments, my experience is quite limited when it comes to releasing software that is supposed to work with other software when they both evolve over time. I can see the problem when Yomichan can't update the native component by itself and breaking changes are introduced.

Because the version of the native component doesn't change during a connection, I might add a method such as this
```javascript
async getVersion() {
    if (this.port === null) {
        return {};
    }
    return await this.invoke('get_version', {});
}
```
and check if it matches `Mecab.version` right after connecting.

I agree with all your points about distributing the binary for my own part (I would just install the software separately and replace the binaries manually). Getting the registry keys to work correctly was very frustrating because I first tried to create a key called `yomichan_mecab`, but only after installing [another extension with native messaging functionality](https://addons.mozilla.org/en-US/firefox/addon/open-with/) I found out that I was actually supposed to create a key with empty string as the name, and that would show up as `(Default)` in `regedit`. My motivation to research how to get the installed mecab binary path correctly was quite low after that.

Now I've given it more thought, and the most official looking mecab download link seems to be [here](http://taku910.github.io/mecab/#download), but the executable is an installer that places the binary in program files. I might be able to use environment variables like `%programfiles(x86)%\mecab\bin\` to copy `mecab.exe` and `libmecab.dll` to the place where the script is, or make a check in `mecab.py` that changes the binary location when the platform is Windows.

toasted-nutbread:
There are merge conflicts caused by #274 and some other commits I made. Would you like me to help fixing those at all?

siikamiika:
@toasted-nutbread I'm actually almost done with that, but thank you for the offer. The reason I left this pull request waiting is that I participated in a CTF and had lots of studies.

siikamiika:
I dropped the expandTags "optimization" because the tags seem to be cached already which I didn't realize at the time.

ext/bg/js/mecab.js
```diff
@@ -0,0 +1,92 @@
...
+            const {callback, timer} = this.listeners[sequence];
+            clearTimeout(timer);
+            callback(data);
+            delete this.listeners[sequence];
+        }
+    }
+
+    invoke(action, params) {
+        if (this.port === null) {
+            return {};
```
toasted-nutbread: For consistent return type:
```js
return Promise.resolve({});
```

Issue #134, opened 2018-09-02
Non-selectable text in context sentences


HelenFoster:
This has come up on Satori Reader recently ( e.g. https://satorireader.com/articles/sanzu-no-kawa-episode-1-edition-m ), where they added "訳" and "再" to the inline "notes" and "play" buttons, and Yomichan is sending them to Anki along with the sentence.

If I select a block of text in the browser and copy-and-paste it, the button text doesn't get copied. Checking the CSS on those buttons for the reason, I found this:

```
.noselect {
  -webkit-touch-callout: none;
  -webkit-user-select: none;
  -moz-user-select: none;
  -ms-user-select: none;
  user-select: none;
}
```

Would it make sense to avoid grabbing non-selectable text? I'm not sure what this CSS feature is normally used for, and whether this use is typical.

toasted-nutbread:
```user-select``` is an interesting case. In typical cases, ```user-select: none``` _should_ indicate that the text should be ignored. However, it may still be useful to scan such text, so disabling it entirely may not be optimal. (There is a similar difficulty with the CSS [```content```](https://developer.mozilla.org/en-US/docs/Web/CSS/content) attribute, which currently is never scanned.)

However, in the case of the [satorireader.com](https://satorireader.com/articles/sanzu-no-kawa-episode-1-edition-m) URL you linked, the ```user-select: none``` style is combined with ```color: transparent```, which I think we can be 99% certain that the element should be ignored when scanning. The scanning algorithm already ignores other non-visible elements, so I think this combination of styles should also qualify as non-visible.

This particular case will be resolved by #286.

Issue #276, opened 2019-11-08
[Bug Report] Pulling wrong word on netflix


edoreld:
I'm using the subadub extension to make Netflix subtitles highlightable. 

When I then use yomichan's "Add to Anki" functionality (the big green + button) on Netflix, the card is added to Anki but for some reason the last word of the current episode's title is added to the beginning of yomichan's {sentence} field.

How to reproduce:
1. Install SubaDub, Yomichan, Anki, AnkiConnect
2. Set up Yomichan integration with Anki
3. Open episode in Netflix ([ex](https://www.netflix.com/watch/80240866?trackId=13752289&tctx=0%2C1%2Cbd44b0bb3be55a2f2421a5c980f7d469d8fe5ca5%3Ac2f366debda0a1b6b78ca6bd2436659dc90a1cf5%2C%2C))
4. Highlight word with Yomichan
5. Click on big, green + button.

When word is added to Anki, the episode's last word will be added to the beginning of the expression field. 

Related issue on subadub's repo: https://github.com/rsimmons/subadub/issues/9

toasted-nutbread:
The issue is likely on Yomichan's side. The issue stems from the difficulties in determine what the sentence bounds are of the scanned word. The current method heavily relies on the assumption that there will be some sort of delimiting characters around the sentence, which is not always the case. Therefore, the [docSentenceExtract](https://github.com/FooSoft/yomichan/blob/b05960967098262c619e28be5238539c26b7b697/ext/fg/js/document.js#L142) function would likely need some tweaking. It's difficult to have it capture every use case due to the complexities of HTML and CSS, but I definitely think there are some improvements which are possible.

Related: #273, #221

oudajosefu:
As a temporary fix, I suggest adding a regex formatting tool as mentioned in #273. This will allow @edoreld to delete parts of the {sentence} that shouldn't be there, given it's something matchable by regex.

FooSoft:
It might be worthwhile to have different profiles for "sentence" extraction, as it may simply not be possible to come up with a solution that everyone is happy with. We've already seen requests from other users who want to save out "sentences" which are not really sentences at all.

edoreld:
Thanks, I haven't tried it the regex replace thing yet, but if it works as advertised, it looks to be a possible solution.

toasted-nutbread:
[...]

Issue #273, opened 2019-11-04
[Feature Request] Regex formatting when adding cards to Anki


oudajosefu:
Whenever I try to add a card from youtube with its subtitles as the sentence, the timestamps are sometimes added too. With some sort of regex matching similar to the conditions option to match urls with regex, I can add something like `[0-9]:[0-5][0-9].*$` to match the `1:50 / 1:06:07` part in this automatically generated subtitle `金井雅彦「無限と連続」ー高校生のための東京大学オープンキャンパス2018 模擬講義この位相という言葉なんだけれどもこれは物理で使ういそうと数学で使ういそうとは1:50 / 1:06:07` and replace it with a blank string.

It would be nice to have this ability so as to save time deleting parts of added sentences that occur for every part of a video.

toasted-nutbread:
b59a5b8cdb9bf8cdc28701e16c3d581fb96f4ef8 You can now use regex in the handlebars templates. There are two options: one which replaces, and one which just returns the match(es).

```
{{#regexReplace "REGULAR_EXPRESSION" "REPLACEMENT_STRING" "REGEX_FLAGS"}}Whatever content or {{tag}} text{{/regexReplace}}

{{#regexMatch "REGULAR_EXPRESSION" "REGEX_FLAGS"}}Whatever content or {{tag}} text{{/regexReplace}}
```

REGULAR_EXPRESSION is the pattern passed to the [RegExp](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp) constructor.
REPLACEMENT_STRING is the [replacement text](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#Specifying_a_string_as_a_parameter).
REGEX_FLAGS are the [flags passed to the RegExp constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp#Parameters).

toasted-nutbread:
FYI: You are now able to test templates on the settings page. Syntax errors will also show exception information that handlebars returns. Unfortunately, the sentence context is empty, so you can't directly test that value, but other placeholders can be used.

184cc4cf28594f3bef9e141a8cbf9d7eb7a39e88

---

![image](https://user-images.githubusercontent.com/11037431/68535352-2813cb00-030f-11ea-9942-8920dda86fb4.png)

oudajosefu:
Thanks.
```
[...]

PR #286, opened 2019-11-23
Sentence scanning updates

toasted-nutbread:
Due to some recent comments related to sentence scanning, I have tried to make sentence scanning more consistent. The changes are mostly made to ```TextSourceRange```, and the sentence extraction feature is mostly the same. This change also includes some refactoring to reduce redundant code.

A few things to note:
* Newlines in [```Text```](https://developer.mozilla.org/en-US/docs/Web/API/Text) nodes are now only treated as newlines if the CSS [```white-space```](https://developer.mozilla.org/en-US/docs/Web/CSS/white-space) mode lets newline characters break text.
* Element boundaries which cause line breaks add newline characters to the text. This causes sentence detection to more correctly stop when it encounters a newline, since the newlines in the returned text are more accurate.
* This does change the behaviour of how sentence scanning works. IMO it is likely to feel more consistent, but there are potentially some situations where the old method would return more text than the current method.
* This also "fixes" a potential issue which used to exist where text split across lines were scanned as a single term since there was no separation text. For example:
  ```html
  <div><div>小ぢん</div>まり</div>
  <div>小ぢん<div>まり</div></div>
  ```
  Scanning starting at ```小``` would always detect the full word ```小ぢんまり``` despite there being a line break. This example is contrived, but I've seen this happen to two text nodes which only incidentally form a larger phrase.
* This change also includes an optimization where sentence scanning used to scan ```docSentenceExtract.extent``` 3x instead of 2x (only once forward and once backward should be necessary).

There is potentially another improvement that can be done to sentence extraction. Currently the algorithm will terminate a sentence at newline characters, which leads to partial phrases being detected at ```<br>``` elements or line breaks in elements with ```white-space=pre || pre-wrap || pre-line || break-spaces```. This can potentially be improved by only terminating when there are two sequential newline characters, combined with forcing element boundaries to be treated as ```'\n\n'``` rather than just ```'\n'```. I have not yet made this change as I first wanted to get the base changes out of the way.

This addresses #221, #273, #276.

### DIFF
--- a/ext/fg/js/source.js
+++ b/ext/fg/js/source.js
@@ -47,14 +47,14 @@ class TextSourceRange {
     }
 
     setEndOffset(length) {
-        const state = TextSourceRange.seekForward(this.range.startContainer, this.range.startOffset, length);
+        const state = TextSourceRange.seek(this.range.startContainer, this.range.startOffset, length);
         this.range.setEnd(state.node, state.offset);
         this.content = state.content;
         return length - state.remainder;
     }
 
     setStartOffset(length) {
-        const state = TextSourceRange.seekBackward(this.range.startContainer, this.range.startOffset, length);
+        const state = TextSourceRange.seek(this.range.startContainer, this.range.startOffset, -length);
         this.range.setStart(state.node, state.offset);
         this.rangeStartOffset = this.range.startOffset;
         this.content = state.content;
@@ -98,14 +98,25 @@ class TextSourceRange {
         }
     }
 
-    static seekForward(node, offset, length) {
-        const state = {node, offset, remainder: length, content: ''};
-        if (length <= 0) {
+    static seek(node, offset, length) {
+        const forward = (length >= 0);
+        const state = {
+            node,
+            offset,
+            content: '',
+            remainder: (forward ? length : -length)
+        };
+        if (length === 0) {
             return state;
         }
 
         const TEXT_NODE = Node.TEXT_NODE;
         const ELEMENT_NODE = Node.ELEMENT_NODE;
+
+        const seekTextNode = forward ? TextSourceRange._seekForwardTextNode : TextSourceRange._seekBackwardTextNode;
+        const getNextNode = forward ? TextSourceRange._getNextNode : TextSourceRange._getPreviousNode;
+        const shouldEnter = TextSourceRange._shouldEnter;
+
         let resetOffset = false;
 
         const ruby = TextSourceRange._getRubyElement(node);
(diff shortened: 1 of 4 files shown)


toasted-nutbread:
38ff0d820bf5bd571053add467697cb72e9145c9 resolves #134.

siikamiika:
I tested this a bit, and it looks pretty cool. One improvement is that you can now scan the readings on the kanji info page. I can review the code some day next week.

ext/fg/js/source.js
```diff
@@ -117,17 +118,19 @@ class TextSourceRange {
 
             if (nodeType === TEXT_NODE) {
                 state.node = node;
-                if (resetOffset) {
-                    state.offset = forward ? 0 : node.nodeValue.length;
-                }
-                if (lineBreak) {
-                    if (resetOffset && addLineBreak(state, forward)) {
+                if (!resetOffset || shouldSeekTextNode(node)) {
```
siikamiika: For some reason, `shouldSeekTextNode` gets an off-screen ARIA element on Google Docs, returns `false` and after that `Range.setEnd` errors like this:

```core.js:81 Yomichan v1.9.5 has encountered an error.
Originating URL: https://docs.google.com/document/d/1GvxFCXYpAFte9_qEwg2fCfFP0HvX68Dee3RmHAR2MAU/edit
IndexSizeError: Failed to execute 'setEnd' on 'Range': The offset 50 is larger than the node's length (28).
Error: Failed to execute 'setEnd' on 'Range': The offset 50 is larger than the node's length (28).
    at TextSourceRange.setEndOffset (chrome-extension://jpjldhmgaijbpomlojkhkghbbebbiblb/fg/js/source.js:52:20)
    at docSentenceExtract (chrome-extension://jpjldhmgaijbpomlojkhkghbbebbiblb/fg/js/document.js:148:17)
    at Frontend.showContent (chrome-extension://jpjldhmgaijbpomlojkhkghbbebbiblb/fg/js/frontend.js:395:26)
    at Frontend.searchSource (chrome-extension://jpjldhmgaijbpomlojkhkghbbebbiblb/fg/js/frontend.js:368:26)
    at async Frontend.searchAt (chrome-extension://jpjldhmgaijbpomlojkhkghbbebbiblb/fg/js/frontend.js:345:17)
    at async search (chrome-extension://jpjldhmgaijbpomlojkhkghbbebbiblb/fg/js/frontend.js:104:13)

Issues can be reported at https://github.com/FooSoft/yomichan/issues```

**Comment by siikamiika** [comment]
Apart from the Google Docs bug, I didn't notice any issues. ed19a8b185f6ba27662b08123aafb6f8cc7b9d9c looked like it could have an off-by-one bug, but after closer inspection it seems to work the same.

File: ext/fg/js/source.js (hunk 1 above)
Comment (toasted-nutbread) [reply ext/fg/js/source.js]: This issue has been fixed in 047cfba227f7f01a34763ffe6867842662cc02b5, but it has brought another issue to my attention. Will document it in a separate comment.

**Comment by toasted-nutbread** [comment]
Due to how Google docs works, lines are effectively formatted as follows:
```html
<div>... This is a single sentence</div><div>which word wraps.</div>
```
In English, this works (mostly) fine since there are whitespace/newline word boundaries. However, this causes issues in Japanese since characters can wrap mid-word. Take for example:

```html
<div>...吹</div><div>き</div>
```

The previous algorithm handled this "correctly", in that the text would be detected as a single word. The updated algorithm inserts a `\n` character between `吹` and `き`, which makes the word scanner not pick it up.

The issue is related to both the HTML markup _and_ how we decide to treat visual line breaks due to the HTML presentation. For example, despite being presented identically to the example above, the following HTML would _not_ detect `吹き` as a single word, neither with the old algorithm nor the updated one:

```html
<div>...吹</div>
<div>き</div>
```

This looks to be more complicated than I had originally thought.

siikamiika:
It could make sense to add an option for this as well so that you could toggle line break autodetection on and off, and users could disable the detection on Google Docs or other problematic sites. Most of the time you will be scanning regular paragraphs. The option default could also be the other way around, because most of the issues caused by too eager scanning are related to sentence extraction which is probably less used than just scanning.

toasted-nutbread:
Replaced by #536.

PR #288, opened 2019-11-24
Remove wanakana reference from float.html

toasted-nutbread:
Creating this as a PR in order to ask: does anyone know if this reference is used for something? As far as I can tell it's unused, but I want to make sure I'm not overlooking anything.

### DIFF
--- a/ext/fg/float.html
+++ b/ext/fg/float.html
@@ -31,8 +31,6 @@
             </div>
         </div>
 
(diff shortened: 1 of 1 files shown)


FooSoft:
No idea why it's there, probably copy-pasta from before? Removal makes sense to me.

toasted-nutbread:
Thanks

PR #291, opened 2019-11-26
Cloze fixes

### LINKED ISSUE #52: [feature request] Add cards as suspended
Would it be possible to add an option to add the card as suspended?  I usually have a few extra modifications to make to my cards before I actually want to see them.

### LINKED ISSUE #58: [bug] Katakana lookup changed sentence
Looked up オオカミ.
Found 狼 (おおかみ) which has audio, and 狼 (オオカミ) which doesn't.
Added the first one to Anki (in kanji).
The cloze-body then had the word written in hiragana (instead of katakana).
(Yomichan 1.1.18)

toasted-nutbread:
Does this fix #58, specifically in afddec66eb749aa0a39c7e7142f929e68697995d? I'm not sure if there's something I'm overlooking, but my understanding is that ```text.substring(offset, offset + source.length)``` should equal ```source``` most of the time, with the only difference being the use of katakana vs hiragana.

### DIFF
--- a/ext/mixed/js/display.js
+++ b/ext/mixed/js/display.js
@@ -640,7 +640,7 @@ class Display {
 
         if (source) {
             result.prefix = sentence.text.substring(0, sentence.offset).trim();
-            result.body = source.trim();
+            result.body = sentence.text.substring(sentence.offset, sentence.offset + source.length);
             result.suffix = sentence.text.substring(sentence.offset + source.length).trim();
         }
 
--- a/ext/mixed/js/display.js
+++ b/ext/mixed/js/display.js
@@ -633,18 +633,13 @@ class Display {
         return index >= 0 && index < entries.length ? entries[index] : null;
     }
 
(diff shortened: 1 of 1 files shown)


FooSoft:
That looks fine to me. That being said, it has been a pretty long time since I've looked at the particulars of the cloze stuff; you probably have a better grip on that at this point in time :smile: 

I say we ship it and then if there are any issues in testing, we can deal with them then. I'll push out a new test build tonight with all of the changes since the last one.

toasted-nutbread:
Thanks, I appreciate the sanity check because [occasionally I'll overlook something obvious](https://github.com/FooSoft/yomichan/pull/203) :)

FooSoft:
It's fine to miss stuff, I've honestly kind of given up on the idea of catching every potential problem. Ideally we would have unit tests that cover every single feature, but that would be an insane amount of effort for something that works almost as well (testing branch), hehe.

Issue #289, opened 2019-11-24
.eslintrc file


toasted-nutbread:
The [README.md](https://github.com/FooSoft/yomichan/blob/91c54e1853025c82b9f9e60c490c989ff7149d7d/README.md) file has a guideline under the development heading:

> Follow the [conventions and style](https://raw.githubusercontent.com/FooSoft/dotfiles/master/.eslintrc.json) of the existing code.

This URL is now 404'd due to the repository being hidden or deleted. @FooSoft: do you have a copy of this file available that could be added to this project? If not, I can make a new one.

I already have a linter enabled, but the settings I have may differ from others.

FooSoft:
Whoops -- that file went away when I switched my dotfiles repository to private. I forgot that I had other things referencing it. I've re-added `.eslintrc.json` to the project root.

FooSoft:
What kinds of things do you have in your lint config btw, @toasted-nutbread? We should add anything that can help keep the code qualify and consistency up.

toasted-nutbread:
I have been using jshint rather than eslint, but it looks like eslint provides more options, which is good. Listed below is a list of some things I noticed after enabling eslint and scanning through the code. I can go through and update all of the code to be conformant after this.

* ```"curly": ["error", "all"]``` should be enabled. The current style already appears to be using this, but it's not technically enforced by the ruleset.
* ```"no-constant-condition": "off"``` should be added. Effectively this allows for ```while (true)``` statements, which are used (sparingly) at 5 locations in the code.
* ```"arrow-parens": ["error", "as-needed"]``` might not be the best option. It looks like the project currently has 142 occurrences of using a single-argument arrow functions, 88 of which include parentheses around the argument, and 54 of which don't (62% vs 38%). Should ```"as-needed"``` be changed to ```"always"```? IMO this would make the syntax of all arrow functions more consistent, and eslint's [arrow-parens docs](https://eslint.org/docs/rules/arrow-parens) even links to [Airbnb's JS Style Guide](https://github.com/airbnb/javascript#arrows--one-arg-parens), which has a similar rationale.
* Having [```require-atomic-updates```](https://eslint.org/docs/rules/require-atomic-updates) enabled affects the code in some weird ways. While some of the warnings are completely understandable, there are a few places where it warnings unexpectedly. Might be best to leave enabled but flag code with ```// eslint-disable-line require-atomic-updates```.
  For example, eslint complains about the assignment ```persisted = true;``` below:
[...]

PR #295, opened 2019-11-26
Lint updates

toasted-nutbread:
Addresses lint issues in #289.

The question about whether ```"arrow-parens"``` should be ```"as-needed"``` or ```"always"``` is still an open question. Once answered, I will add a commit to address whatever the preference is.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -8,7 +8,7 @@
         "/ext/bg/js/templates.js"
     ],
     "rules": {
-        "arrow-parens": ["error", "as-needed"],
+        "arrow-parens": ["error", "always"],
         "comma-dangle": ["error", "never"],
         "curly": ["error", "all"],
         "dot-notation": "error",
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -3,19 +3,26 @@
     "parserOptions": {
         "ecmaVersion": 8
     },
+    "ignorePatterns": [
+        "/ext/mixed/lib/",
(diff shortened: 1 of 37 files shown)


FooSoft:
Very nice, thanks for the cleanup :+1:

toasted-nutbread:
@FooSoft do you have a preference for parentheses and arrow functions? I.e. should they be mandatory (```"always"``` / ```(e) => something(e)```) or omitted when single-arg (```"as-needed"``` / ```e => something(e)```)? The former is currently the more common in the codebase.

FooSoft:
I tend to avoid using unnecessary characters where possible (no extra parens in `if` statements for example), but also at the same time I prefer consistency (so always do curly braces, even for single line control statements), so this could go either way.

I've noticed that you do mandatory parens in your code, and you are probably right to favour consistency over brevity for this. As you mentioned, the fact that this style is adopted by Airbnb; that may very well indicate that is the more "standard" way of doing things (so we might as well make it be the standard for this codebase as well).

toasted-nutbread:
> I've noticed that you do mandatory parens in your code, and you are probably right to favour consistency over brevity for this.

I don't think I followed a consistent pattern in all of the code I've wrote. Sometimes I followed the existing ```e => something(e)```, but other times I didn't. I do default to parens, but if there is a reference in close proximity which was using no-parens, I dropped them. Since there wasn't a lint file previously, I didn't have anything to go by.

Issue #267, opened 2019-10-26
Keyboard, mouse, and other inputs


toasted-nutbread:
This issue is intended to be an overview of issues features related to inputs. There are some changes I want to make to text-scanning inputs, and I'm going to list all other input-related issues I can find.

**Add support for pen devices using [pointer events](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events)**
* Refactor frontend.js to be more simple, easier to understand, and more compatible with the changes that will be necessary for pointer devices.
* Implement support for pen devices using pointer events on devices which support it.

**Customizable scanning modifier keys**
* Custom modifiers for searching either terms, kanji, or both. #79

**Customizable webpage hotkeys**
* Trigger lookup of selected text. #161
* Editing existing hotkey mapping & creating new hotkeys. #127 #133

**UI changes**
* Potentially some tweaks to the organization and labeling of some of the options on the settings page. #120
* UI for editing and assigning custom keyboard shortcuts.

siikamiika:
Just adding a mention of gestures here as well in addition to https://github.com/FooSoft/yomichan/issues/292#issuecomment-560111090

siikamiika:
Another issue: the search page query parser and glossary clicking don't currently support touch events. I could copy and paste code from `frontend.js`, but that would mean a lot of duplication. Instead, it could be useful to refactor scanning to a new class and use it from `frontend.js` and other places as needed. Also mentioned in https://github.com/FooSoft/yomichan/pull/279#issue-339144238, but `Frontend` might be too tied to popups for subclassing.

siikamiika:
@toasted-nutbread Are you working on something that would conflict if I extract the scanning functionality from `frontend.js` or general objections?

toasted-nutbread:
The only thing I've somewhat started on is pointer events, so I don't have anything that should conflict.

siikamiika:
Swipe gestures could work only on some parts of the popup or search page, similar to what's done in https://github.com/FooSoft/yomichan/commit/e99c8c40a1b87e307d5365fe619fa1a71996b500 and how Chrome for Android switches between tabs.

siikamiika:
[...]

PR #299, opened 2019-12-05
Extract text scanner from Frontend 

siikamiika:
This refactoring was done because other parts of Yomichan are also using text scanning. In popups you can click kanji and glossary, and on search page you can scan text using query parser. Query parser needs to do almost everything that normal scanning does but it doesn't have a popup. Instead of copying and modifying all touch event code from Frontend I created a new class that takes in the scanned element and rules about what elements and places to ignore, and offers a pub/sub interface that emits text sources and other events related to scanning.

After this is reviewed, Im going to replace the duplicated scanning code in query parser and possibly also Display.

https://github.com/FooSoft/yomichan/issues/267
https://github.com/FooSoft/yomichan/pull/279

### DIFF
--- a/ext/fg/js/frontend.js
+++ b/ext/fg/js/frontend.js
@@ -185,10 +185,6 @@ class Frontend {
             type,
             {definitions, context: {sentence, url, focus, disableHistory: true}}
         );
-
-        if (this.options.scanning.selectText) {
-            textSource.select();
-        }
     }
 
     async findTerms(textSource) {
--- a/ext/fg/js/frontend.js
+++ b/ext/fg/js/frontend.js
@@ -17,26 +17,22 @@
  */
 
 
-class Frontend {
+class Frontend extends TextScanner {
     constructor(popup, ignoreNodes) {
-        this.popup = popup;
-        this.textScanner = new TextScanner(
+        super(
             window,
             ignoreNodes,
-            [this.popup.container],
+            [popup.container],
             [(x, y) => this.popup.containsPoint(x, y)]
         );
-        this.textScanner.subscribe('textSearch', ({textSource, cause}) => this.searchSource(textSource, cause));
-        this.textScanner.subscribe('searchClear', () => this.searchClear(true));
+
+        this.popup = popup;
         this.options = null;
 
         this.optionsContext = {
             depth: popup.depth,
             url: popup.url
         };
-
-        this.enabled = false;
-        this.eventListeners = [];
 
         this.isPreparedPromiseResolve = null;
         this.isPreparedPromise = new Promise((resolve) => { this.isPreparedPromiseResolve = resolve; });
@@ -70,7 +66,7 @@ class Frontend {
     }
 
     async onResize() {
-        const textSource = this.textScanner.getCurrentTextSource();
+        const textSource = this.textSourceCurrent;
         if (textSource !== null && await this.popup.isVisibleAsync()) {
             this.lastShowPromise = this.popup.showContent(
                 textSource.getRect(),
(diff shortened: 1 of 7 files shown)


ext/fg/js/frontend.js
```diff
@@ -71,162 +62,9 @@ class Frontend {
...
-            return;
-        }
-
-        this.primaryTouchIdentifier = null;
-        this.preventScroll = false;
-        this.preventNextClick = false;
-        // Don't revert context menu and mouse down prevention,
-        // since these events can occur after the touch has ended.
-        // this.preventNextContextMenu = false;
-        // this.preventNextMouseDown = false;
```
siikamiika: Was this commented-out code here on purpose? The explanation above looks like it was, but I'm not sure. I just copied it over to the new class.

toasted-nutbread: I believe the code was commented on purpose, with the comment indicating that it's necessary to _not_ run that code.

ext/mixed/js/text-scanner.js
```diff
@@ -0,0 +1,386 @@
...
+            }
+        }
+        this.pendingLookup = false;
+    }
+
+    onError(error) {
+        logError(error, false);
+    }
+
+    subscribe(eventName, subscriber) {
```
toasted-nutbread: Looks like an event listener pattern is implemented here with ```subscribe``` and ```publish```. Is there a situation where there will be multiple listeners for events? If not, you might be able to simplify the code by just having overridable stub functions, similar to either of the below:

https://github.com/FooSoft/yomichan/blob/14c5857dfe58eb14599438f8cfb8d436d38e2006/ext/mixed/js/display.js#L45-L47
https://github.com/FooSoft/yomichan/blob/14c5857dfe58eb14599438f8cfb8d436d38e2006/ext/bg/js/conditions-ui.js#L55-L57

The reason I bring this up is because typically the event listener pattern is synchronous, whereas this publish version is async.

It looks like there are two places where ```publish``` is called:
* ```await this.publish('searchClear', {});``` doesn't need to be async because the searchClear handler isn't async.
* ```await this.publish('textSearch', {textSource, cause});``` does need to be async, but it looks like it only expects one result.

I noticed this because I recently added a similar event listener pattern on one of my development branches (which isn't async), and also used the "overridable stub" pattern a bit.
https://github.com/toasted-nutbread/yomichan/commit/648f807e53adbfc3d7c33a8d8b26a022ea047842

siikamiika: > Looks like an event listener pattern is implemented here with `subscribe` and `publish`.

Oops, you're right. I tried to choose a name that can't be confused with the already existing `eventListeners` array and code related to DOM event handling, but this implementation doesn't have an event bus.

> Is there a situation where there will be multiple listeners for events? If not, you might be able to simplify the code by just having overridable stub functions

I can't think of one for this use case or for query parser, but I usually like to use lists by default when writing generalized code (although this doesn't *have* to be generalized). I assume you meant that I would do `class Frontend extends TextScanner` and `class QueryParser extends TextScanner` in the future; one problem I see is that TextScanner needs access to the Popup element and its containsPoint method in `Frontend`, but because Popups don't exist in query parser, I don't want to make the code depend on it explicitly. Having a different namespace makes that distinction clearer. It could also be useful to be able to add a TextScanner to any number of elements.

Anyway I'm all for simplification where possible, and I'll see how it goes.

siikamiika: It seemed to work quite nicely, and I managed to shave off some extra lines of code too. Thanks for the idea :+1: Are there other things you think could be better?

toasted-nutbread: > I assume you meant that I would do ...

There's a few ways of going about it:
* Inheritance
  Subclass and provide override definitions in the subclass. This is what you're currently doing after the update.
* Composition (pure)
  This is basically what you originally did: `Frontend` has a private reference to an instance of `TextScanner`, and it's implementation is not modified.
* Composition (hybrid)
  I'm not sure what the technical name of this would be, but this would be where `Frontend` has a private reference to `TextScanner`, and it overrides methods by assigning properties to the instance which override the prototype chain. This is kind of a mix of both.

Here are some examples of the "hybrid" composition pattern that have been used:
https://github.com/FooSoft/yomichan/blob/a5a6351cceacde6f440d28ff7f9afa7c6c1ff251/ext/bg/js/settings/audio.js#L25-L30

https://github.com/FooSoft/yomichan/blob/a5a6351cceacde6f440d28ff7f9afa7c6c1ff251/ext/bg/js/settings/profiles.js#L90-L100

I've generally found the third pattern to be the most brief of the three. Pick whichever one you feel is the best for this case. The only real deciding factor I can think of for this change would be whether you want `TextScanner`'s public APIs to be exposed on the `Frontend` class.

siikamiika: I like your hybrid approach in that it makes it really obvious what's being overridden. Inheritance has always felt like magic to me, but at least it's well understood magic. It's also used to extend `Display` to get `DisplayFloat` and `DisplaySearch`, which is why I assumed you meant only that after seeing your example about `Display.onError`.

In this case the original code was already in a class, so extracting a superclass was probably the best first step. That way I can reuse the `hookEvents` architecture and don't have to duplicate a lot of code or do other refactoring.

I don't necessarily want `TextScanner`'s public APIs to be exposed, but right now `setTextSourceScanLength` is needed in `Frontend` and I think it's good that `ignoreNodes` and such are in `TextScanner` instead. Maybe `TextSourceElement` could could take `ignoreNodes` so that the method wouldn't be needed in textScanner.

siikamiika:
I'll merge this as it is now, and move on to `QueryParser`.

Issue #303, opened 2019-12-08
Firefox dead objects


toasted-nutbread:
Anything using `utilBackend` needs to also use `utilBackgroundIsolate` on the arguments in order to prevent dead objects on Firefox.

```
TypeError: "can't access dead object"
```

I recently saw this happen while the database updates, causing the backend to lag. This issue probably always existed, it was just never noticed due to nothing stalling the action.

toasted-nutbread:
Looking more into this, it's not completely straightforward to add `isolate` calls to everything without having redundancies. It's easy to fix `bg/util.js`, but `bg/api.js` is a bit more involved. The reason is that there are two ways that API functions from `bg/api.js` can be called:
* The first is when the functions are mapped and called directly from the background window. This case would not need any isolation since all objects are already owned by the correct scope.
  https://github.com/FooSoft/yomichan/blob/2dad7f888b77c7c0293512076b719026f6b95de9/ext/bg/js/backend.js#L185-L204
* The second is where functions are called from extension pages with direct access to the background window, i.e. the settings page, search page, context popup page, etc. These pages need isolation.
  https://github.com/FooSoft/yomichan/blob/2dad7f888b77c7c0293512076b719026f6b95de9/ext/bg/context.html#L184
  and
  https://github.com/FooSoft/yomichan/blob/2dad7f888b77c7c0293512076b719026f6b95de9/ext/bg/js/context.js#L68

The issue is that there is currently no distinction between these, so adding isolation calls inside of `bg/api.js` would incur unnecessary overhead. I think the best thing to do might be to add an abstraction layer between the background page and other extension pages. I will continue to look at the `bg/api.js` and `bg/util.js` to see the optimal method of doing this.

PR #306, opened 2019-12-12
Message handler refactoring

toasted-nutbread:
This PR changes several things related to message handlers:
* Static message handler definitions now use `Map` rather than a standard object. The API for `Map` is a bit cleaner than having to do `hasOwnProperty` all the time.
* The `'getUrl'` message handler is now defined in core, rather than being manually defined in multiple files.
* Moved all `api*` function definitions from `bg/js/api.js` into `bg/js/backend.js`. Most of these functions were already using backend state via `utilBackend()`, and the message mapping was already in `backend.js`, so it made sense to move the definitions here.
* Removed `bg/js/api.js`, moved `fg/js/api.js` to `mixed/js/api.js`. This ensures that all API calls are consistent, even if they are called from extension pages. This also fixes some of the issues mentioned in #303 which were caused by `api*` calls. This also makes it so that the authoritative options reference isn't unintentionally passed around, since the values returned by `apiOptionsGet` and `apiOptionsGetFull` now return copies.
* In order to make the settings page work, two functions have been added which provide direct reference to the backend options object: `getOptionsMutable` and `getOptionsFullMutable`. These functions make it more explicit that the object may be modified, which should make it easier to track down usage issues.


siikamiika:
The change looks good and seems to mostly work, but currently global keyboard shortcuts are broken:
```
Uncaught (in promise) Error: Unexpected response of type undefined ({"action":"commandExec","params":{"command":"search"}})
    at api.js:114
```

toasted-nutbread:
> 
> 
> The change looks good and seems to mostly work, but currently global keyboard shortcuts are broken:
> 
> ```
> Uncaught (in promise) Error: Unexpected response of type undefined ({"action":"commandExec","params":{"command":"search"}})
>     at api.js:114
> ```

Fixed that and another bug with a naming issue.

PR #307, opened 2019-12-13
Clipboard monitor refactor

toasted-nutbread:
This addresses a few things:
* Improves performance while the page is focused. Increased delay from 100ms to 250ms. When the page is in the background, delays are typically throttled to minimum 1000ms anyways.
* Ensures there are no overlapping interval callbacks. The `async` delay isn't taken into account when starting the next delay, so technically it's possible for two callbacks to be running 'simultaneously'.
* Added some info about clipboard copying, removing a TODO.
* Improved Firefox browser detection.
* Improved clipboard paste process in the backend to only paste plaintext instead of rich text.

### DIFF
--- a/ext/bg/background.html
+++ b/ext/bg/background.html
@@ -12,7 +12,7 @@
         <link rel="icon" type="image/png" href="/mixed/img/icon128.png" sizes="128x128">
     </head>
     <body>
-        <div id="clipboard-paste-target" contenteditable="true"></div>
+        <textarea id="clipboard-paste-target" contenteditable="true"></textarea>
 
         <script src="/mixed/lib/handlebars.min.js"></script>
         <script src="/mixed/lib/jszip.min.js"></script>
--- a/ext/bg/background.html
+++ b/ext/bg/background.html
@@ -12,7 +12,7 @@
         <link rel="icon" type="image/png" href="/mixed/img/icon128.png" sizes="128x128">
     </head>
     <body>
-        <textarea id="clipboard-paste-target" contenteditable="true"></textarea>
+        <textarea id="clipboard-paste-target"></textarea>
 
         <script src="/mixed/lib/handlebars.min.js"></script>
         <script src="/mixed/lib/jszip.min.js"></script>
--- a/ext/bg/js/search.js
+++ b/ext/bg/js/search.js
@@ -16,13 +16,6 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-
-let IS_FIREFOX = null;
-(async () => {
-    const {browser} = await apiGetEnvironmentInfo();
-    IS_FIREFOX = ['firefox', 'firefox-mobile'].includes(browser);
-})();
-
 class DisplaySearch extends Display {
     constructor() {
         super(document.querySelector('#spinner'), document.querySelector('#content'));
@@ -43,8 +36,12 @@ class DisplaySearch extends Display {
         this.introVisible = true;
         this.introAnimationTimer = null;
 
-        this.clipboardMonitorIntervalId = null;
-        this.clipboardPrevText = null;
+        this.isFirefox = false;
+
+        this.clipboardMonitorTimerId = null;
+        this.clipboardMonitorTimerToken = null;
+        this.clipboardInterval = 250;
+        this.clipboardPreviousText = null;
     }
 
     static create() {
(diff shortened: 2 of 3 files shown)


siikamiika:
Thanks for this! I tested a bit and it seems to work as before, but looks cleaner under the hood.

Unrelated to the scope of this PR, but I'm wondering why https://addons.mozilla.org/en-US/firefox/addon/clipboard-inserter/ that inspired me to implement this feature works without user interaction using `document.execCommand('paste')` on Firefox.

ext/bg/background.html
```diff
@@ -12,7 +12,7 @@
         <link rel="icon" type="image/png" href="/mixed/img/icon128.png" sizes="128x128">
     </head>
     <body>
-        <div id="clipboard-paste-target" contenteditable="true"></div>
+        <textarea id="clipboard-paste-target" contenteditable="true"></textarea>
```
siikamiika: The `contenteditable="true"` is probably redundant for a `<textarea>`

ext/bg/js/search.js
```diff
@@ -241,39 +239,63 @@ class DisplaySearch extends Display {
...
-                curText = (await navigator.clipboard.readText()).trim();
-            } else if (IS_FIREFOX === false) {
-                curText = (await apiClipboardGet()).trim();
-            }
-            if (curText && (curText !== this.clipboardPrevText) && jpIsJapaneseText(curText)) {
-                if (this.isWanakanaEnabled()) {
-                    this.setQuery(window.wanakana.toKana(curText));
-                } else {
-                    this.setQuery(curText);
+        const token = {};
```
siikamiika: Is this used to avoid overlapping callbacks? First time I see an object used like this. :smile: Maybe there could be a comment explaining why the object isn't used to store data.

toasted-nutbread: Yeah, it's effectively a unique identifier used to indicate that the current callback is still active. An incrementing counter could could be similarly used, but technically that has issues with overflow (which in practice would likely not be encountered).

This "token" is comparing to see that the reference is the same, which is only true for the same object. If it's different, a new timer has been started during the `await`.

ext/bg/background.html
toasted-nutbread: Unclear why this is the case, but the cause seems to be the misformatted `<div>` tag.

https://github.com/kmltml/clipboard-inserter/blob/master/bg/index.html#L7
```html
<div id="paste-target" contenteditable="true" />
```

Changing this to a proper `<div>` tag makes it not work.
```html
<div id="paste-target" contenteditable="true"></div>
```

I've created an repository demonstrating the issue. If you want, you can test and confirm that you see the same issue (you have to open up the dev tools for the background page).
https://github.com/toasted-nutbread/firefox-clipboard-paste-bug

I've also created a Firefox bug report because it feels like a bug.
https://bugzilla.mozilla.org/show_bug.cgi?id=1603985

toasted-nutbread:
> Unrelated to the scope of this PR, but I'm wondering why https://addons.mozilla.org/en-US/firefox/addon/clipboard-inserter/ that inspired me to implement this feature works without user interaction using `document.execCommand('paste')` on Firefox.

Unclear why this is the case, but the cause seems to be the misformatted `<div>` tag.

https://github.com/kmltml/clipboard-inserter/blob/master/bg/index.html#L7
```html
<div id="paste-target" contenteditable="true" />
```

Changing this to a proper `<div>` tag makes it not work.
```html
<div id="paste-target" contenteditable="true"></div>
```

I've created an repository demonstrating the issue. If you want, you can test and confirm that you see the same issue (you have to open up the dev tools for the background page).
https://github.com/toasted-nutbread/firefox-clipboard-paste-bug

I've also created a Firefox bug report because it feels like a bug.
https://bugzilla.mozilla.org/show_bug.cgi?id=1603985

siikamiika:
Wow, that's fascinating. I remember editing the self-closing `div` to valid HTML when porting that to Yomichan, but apparently it was intentionally written that way. I added a comment to the bug to reduce the probability of the bug being "fixed" by preventing the clipboard from being read with the invalid `div` too.

I tested the extensions in your repo, and indeed ext1 didn't get the text while ext2 did (BTW I think the names in the manifests are the wrong way around).

toasted-nutbread:
> BTW I think the names in the manifests are the wrong way around

Yep, I've fixed that. It's so confusing that the _invalid_ HTML causes it to work, whereas the _valid_ HTML doesn't work.

Thanks for testing to confirm. I'll update this PR's comments with better information about the issue, along with addressing other feedback.

siikamiika:
Thanks for the fixes and sorry for the delay. I think the PR looks good to merge now, but I want to add this Bugzilla thread because it's related to `contenteditable="true"`:
https://bugzilla.mozilla.org/show_bug.cgi?id=1375794#c2

In our case it doesn't matter that we remove it because it works on Chrome either way and Firefox doesn't use it.

Issue #296, opened 2019-12-01
Settings improvements


toasted-nutbread:
As the number of features Yomichan has has grown, so too has the number of settings. Some of the settings are more complex and require more logic than just a simple HTML input field to adjust. As such, new scripts have been added to control the more advanced behaviour. Such features currently include:
* Anki template preview & validation
* Storage settings and information
* Dictionary list
* Multiple audio playback sources
* Popup preview
* Profiles and conditions

In order to simplify development of future settings and the corresponding UI, there are a few tasks that have come up which could improve the workflow:
* [x] Make all the existing settings controls more modular. This will initially include splitting the settings.js file up into multiple files.
* [x] Following the above, the dependency coupling between files should be reduced to be unidirectional rather than bidirectional. This should make the code easier to understand without having to jump back and forth between two (or more) files.
* [x] Make it more explicit when a mutable reference to the options object is being used. #306
* [x] Add support for a validation schema for settings. See: https://json-schema.org/. Settings are currently not validated, so invalid values can be typed into the input fields and they will be saved, which can cause Yomichan to misbehave. #310
* [x] Enforce schema validation on all settings changes. This can be done using Javascript's proxy system.
* [x] Have a common system for assigning settings which can be used from both the frontend and backend. Make this system the _only_ method for assigning settings. I.e. the settings page will not have any direct references to the backend's options object. Instead, only copies should be passed around, and assignments/updates should use the API calls. This will primarily involve a refactoring of ```apiOptionsSet```.
* [x] Simplify how settings inputs are declared in the settings HTML/JS pages. Adding a new input for a simple setting (boolean, string, etc.) currently requires adjustments in at least three places: settings.html, settings.js/formRead, and settings.js/formWrite. This can be streamlined to just having to update the HTML. #335
* [x] Add support for importing/exporting settings. #112, #311
* [x] Add quick navigation links to settings page. #319
* [x] Add improved descriptions to the settings page.

toasted-nutbread:
For reference, I have currently started on most if not all of these tasks. The first one is taken care of by 403b86675d288a8082125475f87f9b1740e3480a...767a480a847b641768b94b835a6ac10c1c3d69bf, which is just moving code around.

The remaining tasks are all somewhat intertwined, so fully implementing them will take a bit longer. I will put up a pull request for it when it is completed.

siikamiika:
[...]

PR #310, opened 2019-12-15
Options schema

toasted-nutbread:
This change adds support for a JSON validation schema for the options. Currently based on https://json-schema.org/.

The scheme is not fully used yet; validation is currently only performed on extension startup. In the future, the proxy system will be applied to the entire settings object such that all modifications are validated. This will be done in a later commit, otherwise the settings page would be throwing exceptions it's not currently equipped to handle.

This is tracked in #296.

### DIFF
--- a/ext/bg/background.html
+++ b/ext/bg/background.html
@@ -31,6 +31,7 @@
         <script src="/bg/js/deinflector.js"></script>
         <script src="/bg/js/dictionary.js"></script>
         <script src="/bg/js/handlebars.js"></script>
+        <script src="/bg/js/json-schema.js"></script>
         <script src="/bg/js/options.js"></script>
         <script src="/bg/js/profile-conditions.js"></script>
         <script src="/bg/js/request.js"></script>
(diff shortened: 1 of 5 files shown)


siikamiika:
Great for improving reliability working with the options object. I have one concern regarding this and https://github.com/FooSoft/yomichan/pull/311: what happens when you bisect without backing up your options first? AFAIK previously the newer options were not modified. Now you could lose some data when the schema doesn't recognize new options.

Not saying the new settings should be preserved on older versions, but it could be confusing when trying to find the cause for other bugs and I just wanted to bring it up.

toasted-nutbread:
> what happens when you bisect without backing up your options first?

You're right that the schema would remove unknown fields or incorrectly formatted data. Forwards compatibility may be partially supported currently, but I think that's only incidental. It's still very easy to corrupt options when rolling versions back depending on what changes were made. For example, the shift to having profiles was not forwards compatible.

This change makes it so that your options shouldn't be able to become corrupted due to incorrect data types, missing fields, invalid values, etc. You may have settings data loss when rolling back, but the extension should be guaranteed to still have valid settings, and thus the extension will still work.

PR #311, opened 2019-12-15
Settings import export

### LINKED ISSUE #112: [Feature Request] Import & Export Settings
It would be cool if it were possible to export my current settings, so that I can use them on different devices without having to reconfigure everything (importing the dictionaries + setting up anki).
Perhaps just pointing out in the "README" where these settings are saved would be more than enough, I'll happily copy them over manually.

toasted-nutbread:
Adds support for exporting settings, importing settings, and resetting all settings to default values.

**This commit is dependent on the features added in #309 and #310.** I am putting this up now because there are no code conflicts, but the features added in those PRs are required for this change to work.

* The export button exports to a .json file. The file includes the options and some additional metadata for debugging/informational purposes.
* The import button imports this file and overwrites the settings. Since certain options could be considered insecure if they are coming from an untrusted source, a warning will appear if these settings are considered suspicious. Currently this includes:
  * Non-default `anki.fieldTemplates`, since this value can effectively `eval` code.
  * `anki.server` values which are not localhost. Pointing to a remote server could potentially be dangerous.
  * `audio.customSourceUrl` values are not localhost, for the same reasons above.
* The reset button resets all settings to the default values. This removes all profiles, but does not clear dictionary data.

Resolves #112.
This is tracked in #296.

### DIFF
--- a/ext/bg/js/backend.js
+++ b/ext/bg/js/backend.js
@@ -112,6 +112,18 @@ class Backend {
         return this.options;
     }
 
+    async setFullOptions(options) {
+        if (this.isPreparedPromise !== null) {
+            await this.isPreparedPromise;
+        }
+        try {
+            this.options = JsonSchema.getValidValueOrDefault(this.optionsSchema, utilIsolate(options));
+        } catch (e) {
+            // This shouldn't happen, but catch errors just in case of bugs
+            logError(e);
+        }
+    }
+
     async getOptions(optionsContext) {
         if (this.isPreparedPromise !== null) {
             await this.isPreparedPromise;
--- a/ext/bg/js/options.js
+++ b/ext/bg/js/options.js
@@ -481,3 +481,7 @@ function optionsSave(options) {
         });
     });
 }
+
+function optionsGetDefault() {
+    return optionsUpdateVersion({}, {});
+}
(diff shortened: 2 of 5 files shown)


siikamiika:
> Since certain options could be considered insecure if they are coming from an untrusted source, a warning will appear if these settings are considered suspicious

This reminds me of a recent wave of disabled addons on Firefox:
https://www.jeremiahlee.com/posts/page-translator-is-dead/

Although the feature doesn't automatically load remote code, someone reviewing could interpret the rules differently. Takedowns are nasty. I can to find more information about this later.

toasted-nutbread:
> 
> 
> > Since certain options could be considered insecure if they are coming from an untrusted source, a warning will appear if these settings are considered suspicious
> 
> This reminds me of a recent wave of disabled addons on Firefox:
> https://www.jeremiahlee.com/posts/page-translator-is-dead/
> 
> Although the feature doesn't automatically load remote code, someone reviewing could interpret the rules differently. Takedowns are nasty. I can to find more information about this later.

Yeah, I could see how it could be interpreted that way, which was my initial motivation behind trying to do #203. If that ever becomes an issue, I am willing to write a simplified template engine to avoid any `eval`s / remote code.

siikamiika:
I tested this on master with #309 and #310 merged briefly and it seemed to work fine. I think this would be a useful feature for testing and backing up your settings in general. #310 will need maintenance when adding new features like #316. Do you think this is ready to merge?

Issue #313, opened 2019-12-22
[Feature Request] Show readings for numerals written with Arabic numerals


pavlukivan:
Because these readings aren't quite intuitive, especially for exceptions like 1人 (Bad example because it's common, but you get the idea).

siikamiika:
Do you mean that it's not possible to scan 1人 and 2人 etc because of the number?

In general, Yomichan should be able to do that when you enable "Search alphanumeric text" under Scanning Options, but JMdict seems to have those words with full-width digits (i.e. １人、２人), causing your example to not work.

Fixing this would require adding normalization such as [String.prototype.normalize](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize) to all input and also to the installed dictionaries.

On the topic of normalization, maybe also UPPER/lowercase, ひらがな/カタカナ and even variant kanji should be normalized. Full/half-width normalization would be the only one that doesn't require a new schema, though.

toasted-nutbread:
Resolved with #314's option **Convert numeric characters to full width**

PR #314, opened 2019-12-22
Deinflection variants

toasted-nutbread:
This PR adds support for #313 and a few other things.

* Changed how deinflection variants are handled. This was previously only used for finding deinflections using both the raw text and the text source converted to hiragana, but has been extended with some additional options:
  * **Convert katakana to hiragana** (ヨミチャン → よみちゃん) _[default: true]_
  * **Convert half width characters to full width** (ﾖﾐﾁｬﾝ → ヨミチャン) _[default: false]_
  * **Convert numeric characters to full width** (1234 → １２３４) _[default: false]_
  * **Convert alphabetic characters to hiragana** (yomichan → よみちゃん) _[default: false]_
* Each of the above options has three possible values: `'false'`, `'true'`, `'variant'`.
  * `'false'` indicates that the conversion will not be performed.
  * `'true'` indicates that the conversion will always be performed.
  * `'variant'` indicates that both the converted and non-converted text will be searched.
* The above options mean that rōmaji can be scanned for the most part. Exceptions include:
  * Multi-word phrases, since words are separated by spaces when romanized.
  * Roman characters which convert to multiple Japanese characters (cha => ちゃ) can have misleading results. E.g. _yomichan_ => よみちゃん, which can come up with <ruby>夜道<rt>よみち</rt></ruby> as a result, but will highlight _yomicha_.
* Simplified the `jpIsKana` function and added `jpIsCharFullWidth` and `jpIsKanaHalfWidth`, which are also now used to classify whether text is Japanese or not. This makes the "Search alphanumeric text" option work in more cases, e.g. "１人" (full width 1).

One issue that had to be solved for this change was that the conversions had to be able to be mapped back to the source text in order to select the correct amount of characters. This is necessary when the number of pre-conversion characters does not match the number of post-conversion characters, which happens during romanization and with half width kana followed by diacritics. Therefore, I did not go the route of using [String.prototype.normalize](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize), since it does not provide this mapping.

### DIFF
--- a/ext/bg/js/translator.js
+++ b/ext/bg/js/translator.js
@@ -157,7 +157,7 @@ class Translator {
     async findTermsGrouped(text, details, options) {
         const dictionaries = dictEnabledSet(options);
         const titles = Object.keys(dictionaries);
-        const [definitions, length] = await this.findTermsInternal(text, dictionaries, options.scanning.alphanumeric, details, options);
+        const [definitions, length] = await this.findTermsInternal(text, dictionaries, details, options);
 
         const definitionsGrouped = dictTermsGroup(definitions, dictionaries);
         await this.buildTermFrequencies(definitionsGrouped, titles);
@@ -175,7 +175,7 @@ class Translator {
         const dictionaries = dictEnabledSet(options);
         const secondarySearchTitles = Object.keys(options.dictionaries).filter((dict) => options.dictionaries[dict].allowSecondarySearches);
         const titles = Object.keys(dictionaries);
-        const [definitions, length] = await this.findTermsInternal(text, dictionaries, options.scanning.alphanumeric, details, options);
+        const [definitions, length] = await this.findTermsInternal(text, dictionaries, details, options);
         const {sequencedDefinitions, defaultDefinitions} = await this.getSequencedDefinitions(definitions, options.general.mainDictionary);
         const definitionsMerged = [];
         const mergedByTermIndices = new Set();
@@ -212,15 +212,15 @@ class Translator {
     async findTermsSplit(text, details, options) {
         const dictionaries = dictEnabledSet(options);
         const titles = Object.keys(dictionaries);
-        const [definitions, length] = await this.findTermsInternal(text, dictionaries, options.scanning.alphanumeric, details, options);
+        const [definitions, length] = await this.findTermsInternal(text, dictionaries, details, options);
 
         await this.buildTermFrequencies(definitions, titles);
 
         return [definitions, length];
     }
 
-    async findTermsInternal(text, dictionaries, alphanumeric, details, options) {
-        if (!alphanumeric && text.length > 0) {
+    async findTermsInternal(text, dictionaries, details, options) {
+        if (!options.scanning.alphanumeric && text.length > 0) {
             const c = text[0];
             if (!jpIsKana(c) && !jpIsKanji(c)) {
                 return [[], 0];
--- a/ext/bg/js/translator.js
+++ b/ext/bg/js/translator.js
@@ -325,27 +325,31 @@ class Translator {
     getAllDeinflections(text, options) {
         const translationOptions = options.translation;
         const textOptionVariantArray = [
-            Translator.getTextOptionEntryVariants(translationOptions.convertKatakanaToHiragana),
             Translator.getTextOptionEntryVariants(translationOptions.convertHalfWidthCharacters),
             Translator.getTextOptionEntryVariants(translationOptions.convertNumericCharacters),
-            Translator.getTextOptionEntryVariants(translationOptions.convertAlphabeticCharacters)
+            Translator.getTextOptionEntryVariants(translationOptions.convertAlphabeticCharacters),
+            Translator.getTextOptionEntryVariants(translationOptions.convertKatakanaToHiragana)
         ];
 
         const deinflections = [];
         const used = new Set();
-        for (const [hiragana, halfWidth, numeric, alphabetic] of Translator.getArrayVariants(textOptionVariantArray)) {
+        for (const [halfWidth, numeric, alphabetic, hiragana] of Translator.getArrayVariants(textOptionVariantArray)) {
             let text2 = text;
             let sourceMapping = null;
             if (halfWidth) {
                 if (sourceMapping === null) { sourceMapping = Translator.createTextSourceMapping(text2); }
                 text2 = jpConvertHalfWidthKanaToFullWidth(text2, sourceMapping);
             }
-            if (numeric) { text2 = jpConvertNumericTofullWidth(text2); }
+            if (numeric) {
+                text2 = jpConvertNumericTofullWidth(text2);
+            }
             if (alphabetic) {
                 if (sourceMapping === null) { sourceMapping = Translator.createTextSourceMapping(text2); }
                 text2 = jpConvertAlphabeticToKana(text2, sourceMapping);
             }
-            if (hiragana) { text2 = jpKatakanaToHiragana(text2); }
+            if (hiragana) {
+                text2 = jpKatakanaToHiragana(text2);
+            }
 
             for (let i = text2.length; i > 0; --i) {
                 const text2Substring = text2.substring(0, i);
--- a/ext/bg/js/translator.js
+++ b/ext/bg/js/translator.js
@@ -243,6 +243,7 @@ class Translator {
 
                 definitions.push({
                     source: deinflection.source,
+                    rawSource: deinflection.rawSource,
                     reasons: deinflection.reasons,
                     score: definition.score,
                     id: definition.id,
@@ -262,7 +263,7 @@ class Translator {
 
         let length = 0;
         for (const definition of definitions) {
-            length = Math.max(length, definition.source.length);
+            length = Math.max(length, definition.rawSource.length);
         }
 
         return [definitions, length];
@@ -276,6 +277,7 @@ class Translator {
 
         return [{
             source: text,
+            rawSource: text,
             term: text,
             rules: 0,
             definitions,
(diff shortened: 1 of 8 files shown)


toasted-nutbread:
One issue I have found in Firefox is that the text scanning algorithm doesn't scan words starting with a half width character followed by a diacritic, e.g. `ｶﾞｷ`; it seems to start the string at the diacritic rather than the `ｶ`. This works correctly on Chrome, so I assume this is due to a browser difference in calculating glyph size.

For now, I will consider this a separate issue, since it has minimal impact and the changes would have to be made in a completely different place in the codebase (frontend instead of backend).

siikamiika:
Do you think it would make sense to also normalize imported dictionaries so that the `'variant'` option wouldn't be needed? I'm not up to date with how the database works right now, but my vision was that there could be a normalized index column that's used for lookup at all times, and the results could be filtered depending on user preference if they want exact matches for katakana or full/half width characters.

The romaji scanning looks pretty awesome :smile:

toasted-nutbread:
> Do you think it would make sense to also normalize imported dictionaries so that the 'variant' option wouldn't be needed?

I think a potential issue might be having additional unexpected/false positive results appear in the search, though I'm not sure how big of a deal this would be. Consider the following contrived example:
* `アイゴ属`, normalized as `あいご属`. Currently, only `アイゴ属` brings up JMdict's entry for a genus of fish. If normalization was used for everything, `あいご` and `属` found in succession would also bring up this result.

Some other potential issues might include:
* Increased database size, since an additional reading normalization would be stored. As I noticed in #287, the increase in size is not necessarily marginal.
* Users would have to reimport dictionaries to use the feature. It could certainly be possible to support both by checking database features, but that would add increased code complexity for something which is likely rarely used.

The overhead added by these additional variants is currently pretty minimal. If the option isn't enabled, the variant is obviously not searched at all. If the option is enabled, then the variant is only used if necessary. E.g. half width conversion only adds search term variants if half width characters are detected (rare).

siikamiika:
> I think a potential issue might be having additional unexpected/false positive results appear in the search

You're right, the information contained in writing a word in katakana can often rule out false positives. However, in general I think using normalization is more versatile, because it also works when both the query and the result use mixed hiragana/katakana and the orthography is not well established, or it's intentionally changed for an effect (robotic speak in katakana etc). You can always filter out stuff that isn't an exact match, though it costs some processing time. I checked how Jisho does this, and it also seems to convert [あいご属](https://jisho.org/search/%E3%81%82%E3%81%84%E3%81%94%E5%B1%9E) to katakana, but fails to match mixed text such as [あいゴ属](https://jisho.org/search/%E3%81%82%E3%81%84%E3%82%B4%E5%B1%9E).

> Increased database size, since an additional reading normalization would be stored. As I noticed in #287, the increase in size is not necessarily marginal.

This is a fair point. If I undestand the DB correctly, normalization would require adding a normalized column for everything including the reverse columns. There's no way around exploding the size requirements unless it's possible to disable indexing for everything except the primary key, dictionary and sequence in `dict` and creating new tables for normal and reverse lookups that reference the `dict` primary key.

ext/mixed/js/japanese.js
```diff
@@ -17,18 +17,102 @@
  */
 
 
+const jpHalfWidthCharacterMapping = new Map([
```
siikamiika: Do you think a codepoint-based mapping could work here too?

siikamiika: It actually seems quite difficult because halfwidth katakana doesn't have separate codepoints for kana with diacritics and the conversion would get off sync:
https://en.wikipedia.org/wiki/Katakana_(Unicode_block)#Block
https://en.wikipedia.org/wiki/Halfwidth_and_Fullwidth_Forms_(Unicode_block)#Block
The ordering is also different.

siikamiika:
I'm using this branch merged with master now and I'll report any issues I notice. I can do some more focused testing later.

ext/bg/js/translator.js
```diff
@@ -518,4 +556,39 @@ class Translator {
...
+                const c = text[i];
+                if (!jpIsCharacterJapanese(c)) {
+                    text = text.substring(0, i);
+                    break;
+                }
+            }
+        }
+
+        return text;
+    }
```
siikamiika: This is probably more correct than the previous approach of checking the first character only, but it uncovered a bug: in 散々 the kanji iteration mark isn't detected as Japanese text because `jpIsCharacterJapanese` only checks kanji and kana.

WanaKana has its [own version](https://wanakana.com/docs/global.html#isJapanese) that handles quite many special cases: https://github.com/WaniKani/WanaKana/blob/cf3f9e948d96ea40c311135b5082d1ffa77a96bf/src/constants.js#L102

toasted-nutbread: You're right; I also just noticed I introduced a bug with `jpIsKana` since it now only checks a single character:
https://github.com/FooSoft/yomichan/blob/a163a37811f0b230023553d0f1b371d29069902d/ext/mixed/js/japanese.js#L162
^ used to check the entire string.

I moved away from using WanaKana's implementations directly for these functions for two reasons:
* Yomichan's `jp*` character test functions are (for the most part) only testing a single character. With this assumption, the code can be optimized a bit.
* WanaKana's `isJapanese` implemention looks somewhat unoptimized. WanaKana is written for a general use case/environment, but we can make more assumptions based on Yomichan's more specific use cases. (No good way to reference the built code, but you can search for the function definition in the [non-minimized build file](https://unpkg.com/wanakana@4.0.2/umd/wanakana.js).)

Do you see any issue with copying some of the range definitions WanaKana uses and repurposing them inside `japanese.js`? If not, I'll probably do that and do a bit more cleanup and fixing of some of the other `jp*` character testing methods.

Another thing I realized: I don't think `japanese.js` should be in the `mixed` folder, since it's only used by the backend. Using it on the frontend would have an undefined `wanakana` reference.

siikamiika: > Do you see any issue with copying some of the range definitions WanaKana uses and repurposing them inside japanese.js? If not, I'll probably do that and do a bit more cleanup and fixing of some of the other jp* character testing methods.

The only issue I see is having to maintain our own version of a Japanese range detector, which might get changes to its magic numbers or logic in the future when Unicode adds more characters or Japan decides to mix things up, or there's something that hasn't been thought of yet and has to be added. WanaKana (WaniKani) probably has more knowledge and preparation about that kind of stuff.

Anyway, I understand the point that our use case is quite different from what most people need (a browser based kana IME or simple range checker), and it might have a performance impact when you do checks as fast as your browser can emit `mousemove` events. So the priorities of this issue for me are:

1. Have the `jp*` functions abstracted and give them well-defined behavior such as checking a single character at once only
2. Keep scanning as fast as possible
3. If performance doesn't take a hit, prefer using an external lib to minimize the maintenance cost

> Another thing I realized: I don't think japanese.js should be in the mixed folder, since it's only used by the backend. Using it on the frontend would have an undefined wanakana reference.

Probably safe to move it to `bg`. I also grepped the function names in `japanese.js` and looks good.

toasted-nutbread: Another interesting note: WanaKana's `JAPANESE_RANGES` includes `KANA_PUNCTUATION` twice. The order of `JAPANESE_RANGES` is also sub-optimal for the general case since the kanji ranges are at the end.

> Probably safe to move it to bg. I also grepped the function names in japanese.js and looks good.

Cool, I'll probably do that after this commit because it's more of a housekeeping task.

toasted-nutbread: One other thing related to `japanese.js`: DisplayGenerator has its own set of ranges for kanji which was copied out because DisplayGenerator didn't have access to `japanese.js` because it can be used on pages which don't include it.

https://github.com/FooSoft/yomichan/blob/04727a8cd7dc29a02595690389979ec1070bbded/ext/mixed/js/display-generator.js#L326-L332

ext/mixed/js/japanese.js
```diff
@@ -17,24 +17,153 @@
...
+    JP_HIRAGANA_RANGE,
+    JP_KATAKANA_RANGE,
+
+    JP_CJK_COMMON_RANGE,
+    JP_CJK_RARE_RANGE,
+
+    [0xff66, 0xff9f], // Halfwidth katakana
+
+    [0x30fb, 0x30fc], // Katakana punctuation
+    [0xff61, 0xff65], // Kana punctuation
```
siikamiika: Are these left out of `JP_KANA_RANGES` on purpose? It seems to be used in `jpIsStringEntirelyKana` which exists in `jpConvertReading` only. WanaKana's `toRomaji` doesn't seem to be able to handle these anyway, but this would work
```javascript
wanakana.toRomaji('か'+String.fromCharCode('0x3099'))
"ka゙"
wanakana.toRomaji(('か'+String.fromCharCode('0x3099')).normalize())
"ga"
```
These character codes differ between `jpIsCharCodeKana` and `wanakana.isKana`. Both ignore halfwidth katakana:
```
0x3040	぀
0x3097	゗
0x3098	゘
0x3099	゙
0x309a	゚
0x309b	゛
0x309c	゜
0x309d	ゝ
0x309e	ゞ
0x309f	ゟ
0x30a0	゠
0x30fd	ヽ
0x30fe	ヾ
0x30ff	ヿ
```

ext/bg/js/translator.js
siikamiika: > DisplayGenerator has its own set of ranges for kanji which was copied

In that case maybe it would still make sense to have `japanese.js` in mixed and add it to pages where `display-generator.js` is used, but that isn't many lines of duplication.

siikamiika:
Something for later, I noticed that everything in the codebase uses `String.charCodeAt` and `String.fromCharCode`, but some rare Japanese kanji can also have code points higher than `0xffff`. The ranges in this PR don't go above that value, but the dictionaries might as well contain them. The solution is to switch to `String.codePointAt` and `String.fromCodePoint`. `for...of` and `Array.from` should handle surrogate pairs too.

siikamiika:
Should there also be a conversion for hiragana to katakana? That would allow scanning "baito" and getting out バイト when alphabetic to hiragana and hiragana to katakana are enabled. Now only 陪都 is found. Some YouTubers also write ちゃんねる in their channel name, for example.

If normalization seems viable at some point, this would become obsolete, though. The current conversions seem to provide a path to hiragana and full width characters everywhere.

ext/mixed/js/japanese.js
toasted-nutbread: I can just revert `jpIsStringEntirelyKana` to be `wanakana.isKana`, which is the previous behaviour. I'm not concerned about performance in that area. That functionality was outside of the scope of what I was trying to change in this commit, it just got caught up in the changes of what `jpIsKana` was being used for (both single char and multi char). Do you think that sounds like a good idea? If we need more changes such as the `normalize` thing, that could be done separately.

toasted-nutbread:
> Something for later, I noticed that everything in the codebase uses `String.charCodeAt` and `String.fromCharCode`, but some rare Japanese kanji can also have code points higher than `0xffff`. The ranges in this PR don't go above that value, but the dictionaries might as well contain them. The solution is to switch to `String.codePointAt` and `String.fromCodePoint`. `for...of` and `Array.from` should handle surrogate pairs too.

Good point. There may be some other things that would have to be changed to properly support this too, such as text scanning. One example I found with a [quick Google search](https://japanese.stackexchange.com/questions/44221/what-unicode-code-points-are-needed-to-write-japanese-j%C5%8Dy%C5%8D-kanji): &#x53F1; (`&#x53F1;`) vs &#x20B9F; (`&#x20B9F;`)


> Should there also be a conversion for hiragana to katakana? That would allow scanning "baito" and getting out バイト when alphabetic to hiragana and hiragana to katakana are enabled. Now only 陪都 is found. Some YouTubers also write ちゃんねる in their channel name, for example.
> 
> If normalization seems viable at some point, this would become obsolete, though. The current conversions seem to provide a path to hiragana and full width characters everywhere.

I forgot that katakana readings aren't converted to hiragana readings in the database. I can add this as an option for now. If we do go the route of normalization at some point, it wouldn't come automatically for dictionaries which have already been added: they would have to be reimported. So I think it's good to have these options regardless of the future.

ext/mixed/js/japanese.js
siikamiika: The change shouldn't break anything unless some Yomichan or MeCab dictionaries use the combining marks instead of direct character codes for が、ぱ etc. The romaji ruby text is probably also quite an obscure feature. I'm fine if it's left like this for now.

I think it's good to be explicit about what the function does, because I was quite surprised when `jpIsKana` and `wanakana.isKana` returned false for a string that was only partially kana. Now the name tells the functionality even if `wanakana.isKana` is used.

toasted-nutbread:
Added `convertHiraganaToKatakana` option in 4d87681a2301b0bfda091c7e21ce5884e8c206d3. Also added some explanatory text about how these options are used.

siikamiika:
I actually just noticed that wildcard search doesn't work with romaji, but not sure if that's worth fixing now. Nobody has been able to do that before so it's not a regression

ext/mixed/js/japanese.js
```diff
@@ -175,3 +305,145 @@ function jpDistributeFuriganaInflected(expression, reading, source) {
...
+        let c = text.charCodeAt(i);
+        if (c >= 0x41 && c <= 0x5a) { // ['A', 'Z']
+            c -= 0x41;
+        } else if (c >= 0x61 && c <= 0x7a) { // ['a', 'z']
+            c -= 0x61;
+        } else if (c >= 0xff21 && c <= 0xff3a) { // ['A', 'Z'] full width
+            c -= 0xff21;
+        } else if (c >= 0xff41 && c <= 0xff5a) { // ['a', 'z'] full width
+            c -= 0xff41;
+        } else {
```
siikamiika:
```diff
diff --git a/ext/mixed/js/japanese.js b/ext/mixed/js/japanese.js
index 93366db..d595acc 100644
--- a/ext/mixed/js/japanese.js
+++ b/ext/mixed/js/japanese.js
@@ -384,6 +384,9 @@ function jpConvertAlphabeticToKana(text, sourceMapping) {
             c -= 0xff21;
         } else if (c >= 0xff41 && c <= 0xff5a) { // ['a', 'z'] full width
             c -= 0xff41;
+        } else if (c === 0x2d) { // '-'
+            part += String.fromCharCode(c);
+            continue;
         } else {
             if (part.length > 0) {
                 result += jpToHiragana(part, sourceMapping, result.length);
```

I think something similar to this is worth adding before merge to support '-' as an "alphabetic" substitute for ー. Macron '◌̄' is often used too, but there can be combining and direct variants of those so it would require significant changes, and it's also used to mean あ, い or う in text that's normally written in hiragana. I still approve the PR with or without this change.

Wildcards seem to be separated from deinflections anyway so it's probably too complex to add support for them.

toasted-nutbread: I updated the format a little bit in that area and added support for both the standard dash and the fullwidth dash.

Tested with:
```
シーツ
shiitsu
SHIITSU
ｓｈｉｉｔｓｕ
ＳＨＩＩＴＳＵ
shi-tsu
SHI-TSU
ｓｈｉーｔｓｕ
ＳＨＩーＴＳＵ
ｓｈｉ－ｔｓｕ
ＳＨＩ－ＴＳＵ
```

siikamiika: Confirmed working as expected. Thanks for the test case when repeated vowels can mean the long vowel mark. Japanese romanization is full of little things like this that make it a pain to work with. People also use a mix of Hepburn, Nihon-shiki/Kunrei and IME shortcuts or conventions like xa/la for ぁ or thi for ティ.

PR #316, opened 2019-12-23
Popup scaling

### LINKED ISSUE #315: [Feature Request] Automatically adjust popup size in proportion to webpage zoom amount
I actually really dislike when I'm zoomed in and want to scan a word, but the popup stays the same size, and I then have to scroll around to see the definition. I think the popup should automatically adjust itself to the amount the webpage is zoomed in by. Currently I have a solution to this where I made a "Zoomed in" profile, but every time I want to use it, I have to add the website to that profile's conditions (and also the main profile's conditions to make sure it's excluding the website using RegExp). It would be ideal for Yomichan to detect the zoomed in percentage and adjust its size accordingly.

toasted-nutbread:
Adds support for custom scaling of popups, including scaling relative to page zoom level. Also includes a little bit of cleanup near where some related changes were made.

Resolves #315.

### DIFF
--- a/ext/bg/data/options-schema.json
+++ b/ext/bg/data/options-schema.json
@@ -93,6 +93,7 @@
                                     "popupVerticalTextPosition",
                                     "popupScalingFactor",
                                     "popupScaleRelativeToPageZoom",
+                                    "popupScaleRelativeToVisualViewport",
                                     "showGuide",
                                     "compactTags",
                                     "compactGlossaries",
@@ -175,6 +176,10 @@
                                     "popupScaleRelativeToPageZoom": {
                                         "type": "boolean",
                                         "default": false
+                                    },
+                                    "popupScaleRelativeToVisualViewport": {
+                                        "type": "boolean",
+                                        "default": true
                                     },
                                     "showGuide": {
                                         "type": "boolean",
--- a/ext/bg/data/options-schema.json
+++ b/ext/bg/data/options-schema.json
@@ -91,6 +91,8 @@
                                     "popupVerticalOffset2",
                                     "popupHorizontalTextPosition",
                                     "popupVerticalTextPosition",
+                                    "popupScalingFactor",
+                                    "popupScaleRelativeToPageZoom",
                                     "showGuide",
                                     "compactTags",
                                     "compactGlossaries",
@@ -165,6 +167,14 @@
                                         "type": "string",
                                         "enum": ["default", "before", "after", "left", "right"],
                                         "default": "before"
+                                    },
+                                    "popupScalingFactor": {
+                                        "type": "number",
+                                        "default": 1
+                                    },
+                                    "popupScaleRelativeToPageZoom": {
+                                        "type": "boolean",
+                                        "default": true
                                     },
                                     "showGuide": {
                                         "type": "boolean",
--- a/ext/bg/data/options-schema.json
+++ b/ext/bg/data/options-schema.json
@@ -174,7 +174,7 @@
                                     },
                                     "popupScaleRelativeToPageZoom": {
                                         "type": "boolean",
-                                        "default": true
+                                        "default": false
                                     },
                                     "showGuide": {
                                         "type": "boolean",
--- a/ext/bg/js/backend.js
+++ b/ext/bg/js/backend.js
@@ -542,6 +542,15 @@ class Backend {
 
         return new Promise((resolve, reject) => {
             const tabId = sender.tab.id;
+            if (!(
+                chrome.tabs !== null &&
+                typeof chrome.tabs === 'object' &&
+                typeof chrome.tabs.getZoom === 'function'
+            )) {
+                // Not supported
+                resolve({zoomFactor: 1.0});
+                return;
+            }
             chrome.tabs.getZoom(tabId, (zoomFactor) => {
                 const e = chrome.runtime.lastError;
                 if (e) {
--- a/ext/bg/js/backend.js
+++ b/ext/bg/js/backend.js
@@ -51,8 +51,11 @@ class Backend {
 
         this.onOptionsUpdated('background');
 
-        if (chrome.commands !== null && typeof chrome.commands === 'object') {
+        if (isObject(chrome.commands) && isObject(chrome.commands.onCommand)) {
             chrome.commands.onCommand.addListener((command) => this._runCommand(command));
+        }
+        if (isObject(chrome.tabs) && isObject(chrome.tabs.onZoomChange)) {
+            chrome.tabs.onZoomChange.addListener((info) => this._onZoomChange(info));
         }
         chrome.runtime.onMessage.addListener(this.onMessage.bind(this));
 
@@ -92,6 +95,11 @@ class Backend {
             callback({error: errorToJson(error)});
             return false;
         }
+    }
+
+    _onZoomChange({tabId, oldZoomFactor, newZoomFactor}) {
+        const callback = () => this.checkLastError(chrome.runtime.lastError);
+        chrome.tabs.sendMessage(tabId, {action: 'zoomChanged', params: {oldZoomFactor, newZoomFactor}}, callback);
     }
 
     applyOptions() {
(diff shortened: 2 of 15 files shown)


siikamiika:
Nice feature :smile: 

I was testing this zooming a page in and out, and noticed that if you zoom out, the popup size won't be reset to settings default but it stays huge until you scroll so much in that the size in settings doesn't fit anymore. That being said, maybe the poup size shouldn't change at all when zooming in and out. Would there be a way to do that on pages that don't have `font-size: 14px;`?

toasted-nutbread:
> That being said, maybe the poup size shouldn't change at all when zooming in and out.

The reason it does this currently is because the popup size can be expanded by the user and remain expanded. This does act a bit funny with the scaling, so maybe I should just reset the size to default when a scale change occurs.

siikamiika:
I had actually forgotten that it can be resized by the user because the resize handle doesn't appear on Firefox. Your proposed fix sounds good.

toasted-nutbread:
> That being said, maybe the poup size shouldn't change at all when zooming in and out.

Fixed in a032d964c9b165e0f026c380dc8ca7b755dadf7f

siikamiika:
One thing about this and #317 is that now some of the new relative size units don't have a comment and it's also quite hard to edit the style sheets without an editor that can substitute math expressions with their results (and a comment). The font size 14px is also expected everywhere. It could be the time to pick a tool for the job, scss or something.

siikamiika:
The fix for zoom issues is working great :+1:

toasted-nutbread:
> 
> 
> One thing about this and #317 is that now some of the new relative size units don't have a comment and it's also quite hard to edit the style sheets without an editor that can substitute math expressions with their results (and a comment).

I considered using `calc`, but apparently it's not supported (correctly) in the min Firefox version Yomichan supports. I also largely moved away from using `#px => #px` conversions and only used approximations. In general, there is not really a requirement that the units should be corresponding to an exact number of pixels, which is why I used shorter values like `0.36em` in #317. The only reason I used exact `#px => #px` conversions in this PR is because I wasn't trying to change the styles here.

> The font size 14px is also expected everywhere

This is also why I used the "approximate" values in #317 and omitted conversion comments, since it isn't a hard requirement that the font size is `14px`. I chose values that were close in order to maintain the similar visual style, but the intent is that all units are relative to the font size; the font size itself shouldn't really matter.

Personally, whenever I've written CSS, I use `em` units for most things since it makes it easier to change the font size without having to change many other different units to match.

oudajosefu:
Not working for me in Chrome so far. All I did though was set the Popup Size Mode to "Relative to zoom level" so I don't know it that's enough to make the feature work.

toasted-nutbread:
> 
> 
> Not working for me in Chrome so far. All I did though was set the Popup Size Mode to "Relative to zoom level" so I don't know it that's enough to make the feature work.
> 
> Edit: My bad, zooming with Cmd+"+" and Cmd+"-" does work but it wasn't the zooming ability that I was thinking of when I complained initially. I usually zoom by pinching in or out with the track pad, so that's what I had in mind at first. Anyway to make it work with this kind of zoom? I'm on a Macbook Pro if that helps.
> 
> Maybe [this](https://stackoverflow.com/a/28685082) will help. On second thought, [this](https://medium.com/@auchenberg/detecting-multi-touch-trackpad-gestures-in-javascript-a2505babb10e) might be less outdated.

I've looked into this before, and from what I can tell, there doesn't seem to be a way to detect pinch-zoom level. The type of zoom I think you're referring to can be simulated by opening Chrome's dev tools, toggle the device toolbar on, and then hold shift and click and drag with the mouse. The same type of zoom is on mobile Firefox.

siikamiika:
I found [this](https://bugs.chromium.org/p/chromium/issues/detail?id=571297), which lead to a new experimental API called [`visualViewport`](https://developer.mozilla.org/en-US/docs/Web/API/Visual_Viewport_API). It isn't supported in our minimum target versions, but that can always be bumped. Desktop Firefox might need some additional handling because the API only works on mobile by default.

siikamiika:
@toasted-nutbread 
About the CSS issues, thanks for explaining your reasoning behind the comments and using 14px as a baseline somewhere. I dont' want to push a dependency just for this case, but the long floating point values caught my eye in review. `calc` would be nice if it can be made to work on all targets, but that could depend on Firefox fixing a bug or finding a workaround.

toasted-nutbread:
@siikamiika Cool, thanks for finding that; I couldn't find anything when I was searching. I'll take a look into that API to see if that works. Support for it could probably be considered non-essential, so it can be an optional feature which is only used if it's present, without changing the minimum version

> using 14px as a baseline somewhere

Need at least one `px` value to base everything off of, so I just went with what was already there. `16px` is usually easier to have short decimal `em` values which correspond to an exact pixel count, but ¯\\\_(ツ)\_/¯

toasted-nutbread:
I've added a few commits which add initial support for the visual viewport. I'll work on adding some more commits to address #318 and some other positioning issues with the visual viewport soon.

siikamiika:
I tested the new visual viewport option on Chrome in touch emulation mode and the scaling worked great when zooming in and out.

On Firefox Mobile it worked quite weirdly, the scaling depends on the pinch zoom level set when reloading or opening the page. So if you zoom in and reload, the popup size will be normal if you don't zoom at all after that. To get a normal sized popup on the full width page, you must zoom out and reload again. Maybe the onresize events aren't being registered.

toasted-nutbread:
Okay, I've fixed a few zoom issues with the previous few commits. The popup position should now be limited to the viewport, and full width mode should work as well as it can. There are issues with the precision of the viewport values (integer vs float), so the position can appear to jitter. Additionally, on Firefox for Android, the onscroll/onresize events are throttled, so jumping is also noticeable in that situation.

siikamiika:
Looking good! I think this should now handle everything as long as the browser follows the spec.

The issue on Firefox for Android now manifests in a different way, and I'm starting to think it's a bug in the browser itself. The steps to reproduce it are the same (open page and change zoom level), but now I see two other effects instead of the popup zoom not updating:
- If you try to interact with the popup, the browser will sometimes register the touch event elsewhere on the page. You can end up in a situation where you're trying to scroll the popup but instead you see text being scanned in a place that's not under your finger and the popup contents changing.
- Sometimes scanning is blocked, and I think it happens in places where the hidden popup is. Refreshing the page and not resizing after that fixes the problem temporarily.

It seems as if there's a "ghost" version of the popup left behind after pinch zooming that interacts with touch events on the visible version of the popup. Happens on both stable 68.4.1 and beta 68.5 from Play Store.

I don't think this issue should block merge if you agree that the bug is in the browser. Maybe `popupScaleRelativeToVisualViewport` could be off by default, but not sure if that's worth the trouble.

toasted-nutbread:
> If you try to interact with the popup, the browser will sometimes register the touch event elsewhere on the page.

I noticed that as well for a bit but didn't think too much of it, I just restarted the browser and it went away. But it seems like it may be a bigger issue; I'll take another look to see if this is reproducible and if the issue exists independent of this change or not.

> Sometimes scanning is blocked, and I think it happens in places where the hidden popup is.

I'll try to take a look at this too.

toasted-nutbread:
Still looking into the touch events registering at the wrong location. I can get it to happen sometimes, but it's inconsistent and it tends to go away when I try to add more debugging code.

The [MDN page for VisualViewport](https://developer.mozilla.org/en-US/docs/Web/API/VisualViewport) links to some [good documentation](https://github.com/bokand/bokand.github.io/blob/master/web_viewports_explainer.md) on the different viewports and some of the coordinate spaces that various functions/types use. I've been referencing some of the information in there to see if anything Yomichan uses isn't using layout viewport coordinates, but I haven't found anything suspect yet.

I did fix popup positioning when `popupScaleRelativeToVisualViewport` is disabled in 8a577ce58443245d82fae42feffc36ca45eefcdb. Turns out the visual viewport still needs to be used (if it's available), just differently.

toasted-nutbread:
So after looking into the touch events issue, it seems to be another Firefox bug. I've created a [repo](https://github.com/toasted-nutbread/firefox-touch-position-bug) to reproduce the issue and filed a [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1610145), which includes a video demonstration. You can try reproducing using the test I made, and let me know if you test this and notice any other issues with this change.

siikamiika:
Thanks for the thorough investigation and bug report! I tested the repro a bit and it seems that the bug is triggered whenever a touch starts so that the red square *would* be outside of the green frame. If it starts inside of the frame, it will be either scrolled without issues or the visual viewport of the page underneath the frame will be moved, but the red square won't appear.

siikamiika:
I tested the latest commits on all supported platforms and didn't notice any other issues.

Issue #292, opened 2019-11-26
Styling and display of results


toasted-nutbread:
There are a few things that have occasionally come up related to the styling of the results displayed. This issue is intended to track a lot of them, as well potential future research.

* [ ] Additional cosmetic control #119
  While this is partially completed, the usefulness of custom CSS could further extended by adding more metadata attributes to the display HTML and tweaking how certain content is displayed. For example, it could be possible to use the CSS [```content```](https://developer.mozilla.org/en-US/docs/Web/CSS/content) property to change what is displayed for tags in case someone wants to abbreviate (or omit).
* [ ] Update how the audio button is displayed when using "Result grouping = Group results by main dictionary entry". Currently, the audio button only appears when mouse hovering over the text. This does not work nicely on devices using touch inputs.
* [ ] Update existing icon buttons to have better hitboxes for touch inputs. There is a lot of empty space between buttons currently.
* [ ] Add additional icon buttons for other features, e.g. #237. Having some sort of menu popup with customizable links might be helpful.
* [ ] Additional options for the existing buttons. #284, #52. Maybe add support for right-click/shift-click/long-press opening a popup menu with additional options.
* [ ] More sizing options for popup #121. I haven't looked into this, but due to the nature of iframes and automatic content sizing, this may not be possible, but might be worth looking into at some point.
* [ ] Maybe have an option for context sentence extents modification since sentence extraction can't be 100% perfect all the time? Not sure how feasible this would be yet though.
* [ ] Ability to add annotations or update counters for notes #285.
* [ ] Hiding tooltips during scanning #236. This would involve mutation of webpage DOM which Yomichan doesn't "own", which should really be done sparingly to avoid breaking a site's default behaviour.
* [ ] Look into using a shadow DOM for encapsulating popups as an alternative to iframes. See if there is any worthwhile benefit to doing this.

siikamiika:
> Update how the audio button is displayed when using "Result grouping = Group results by main dictionary entry". Currently, the audio button only appears when mouse hovering over the text. This does not work nicely on devices using touch inputs

I did this to save space (mostly because of the tags), but back then touch devices weren't a concern. I'm obviously biased when I say I like the current behavior on devices that have a pointer. Do you have ideas how the tags and the audio button could be displayed on touch devices? Maybe there could be something with a static width that reacts to a tap and expands the extra information for each expression. Currently it's possible to tap somewhere near the fullwidth comma to do the same thing, but it's not obvious to users.

[...]

PR #317, opened 2019-12-28
Display html templates

toasted-nutbread:
This PR removes the the precompiled handlebars templates for `terms.html` and `kanji.html` in favor of using HTML `<template>`s and a generator class. The intent is to generate more standardized HTML which can be more quickly and easily styled and changed, as well as support future additions (#292).

**Change details**
* `terms.html` and `kanji.html` removed in favor of HTML templates and a Javascript DOM creation class. This removes the step of having to compile the handlebars templates.
* The same HTML will now be generated regardless of what **Result grouping** option is used. Presentation is now fully controlled by the CSS. This should make it easier to control custom styling.
* The layout should provide more customization options for experimentation, such as sticky headers.
* The action buttons are now true buttons and have improved hitboxes, which should help on mobile devices. There are no more separation spaces between buttons (which couldn't be clicked).
* Fixed a few styling issues related to the kanji info tables which were introduced when I removed the bootstrap.css dependency from `float.html`.
* Updated/simplified how some content is shown/hidden by using the [`hidden`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/hidden) property.
* Switched CSS units to all use relative units rather than `px`. This change was also made in #316, therefore this change should be merged before that one.
* Styling on the next/previous term buttons has been improved. There is now a fixed header which is only shown when there is navigation history.

**Behaviour differences**
* There is a very slight difference in speed. The entry for each result is created and appended to the document one at a time. This means the top results will be displayed faster, but the total time to generate all results may be a few milliseconds longer. The perceived load time should be a frame or two faster. The scrollbar also now visibly grows, but this is very brief.
  * Performance could be further improved in the future by caching the created DOM nodes and reusing them, since the instantiation process is primarily what takes up the most time.
* Less calls to the backend. In order to display content, the popup (or search page) doesn't have to run API calls to the backend in order to generate the template HTML. Instead, the popup (or search page) directly generates its content. This change may be more noticeable on mobile devices where backend communication is a bit slower.


siikamiika:
I'm getting the orphaned error (Yomichan Updated!) with an existing install, but it works fine after reinstall. Not sure yet what causes it.

> The same HTML will now be generated regardless of what Result grouping option is used. Presentation is now fully controlled by the CSS. This should make it easier to control custom styling.

Great! I should learn to use `dataset` more often too. I can do this to search page parser later together with the refactoring started in https://github.com/FooSoft/yomichan/pull/299. I like the new sticky header that only appears and occupies height when you have history. I noticed it doesn't work on the search page though, and stays on top.

> There is a very slight difference in speed. The entry for each result is created and appended to the document one at a time. This means the top results will be displayed faster, but the total time to generate all results may be a few milliseconds longer. The perceived load time should be a frame or two faster. The scrollbar also now visibly grows, but this is very brief.

There's an issue on search page query parser with the new appending behavior, at least on Firefox. You can see the results blinking while scanning because it generates the same results all over again. Do you think the speed improvement is worth it for the popup only if it's disabled on search page, or should it be made atomic? I'm also wondering if this could cause a situation where results are appended from multiple sources, but I didn't see that happening. Anyway:

> Performance could be further improved in the future by caching the created DOM nodes and reusing them, since the instantiation process is primarily what takes up the most time.

This should fix the blinking.

toasted-nutbread:
>I'm getting the orphaned error (Yomichan Updated!) with an existing install, but it works fine after reinstall. Not sure yet what causes it.

I think that will show up whenever any general error occurs, even if it's not technically orphaned. I noticed that while testing when I had some unhandled errors. Might be able to add some logging.

>I like the new sticky header that only appears and occupies height when you have history. I noticed it doesn't work on the search page though, and stays on top.

It's technically not a sticky header, it's just a fixed header. I've noticed that sticky elements have issues where they jitter when you scroll, and I was noticing this issue. I also made it so it's only fixed on the popup and standard position on the search page, but I suppose it could be changed to sticky on the search page. Another alternative would be to just make the browser forward/back buttons work, but I don't know how much that would require to be changed.

> You can see the results blinking while scanning because it generates the same results all over again.

This can probably be fixed by making it so it doesn't re-search the current term. I think that's already what scanning does on regular pages, and it would also prevent excess redundant database searches for the same term.

> should it be made atomic?

It should minimally not be made atomic on the popup page, since then it is noticeably slower. The above should probably fix it as a short-term solution, and the below can fix it in the long term.

> >  Performance could be further improved in the future by caching the created DOM nodes and reusing them, since the instantiation process is primarily what takes up the most time.
>
>This should fix the blinking.

This would also fix it, and likely allow the update to be made atomic again, but this will require significantly more changes. I consider this a first step towards that goal. This change would likely also have other benefits, such as improved memory usage (less objects created & thrown away).

siikamiika:
> Another alternative would be to just make the browser forward/back buttons work, but I don't know how much that would require to be changed.

This could make the history behavior on search page less confusing (two histories at the same time), but the problem with popups is that the user might not want to lose whatever history they had on the page when first navigating back on the page, then scanning a word and then clicking a kanji.

> It should minimally not be made atomic on the popup page, since then it is noticeably slower. The above should probably fix it as a short-term solution, and the below can fix it in the long term.

Sounds reasonable to me.

toasted-nutbread:
> This can probably be fixed by making it so it doesn't re-search the current term. I think that's already what scanning does on regular pages, and it would also prevent excess redundant database searches for the same term.

For now I am going to consider this task to be outside the scope of this change for two reasons: 1) the issue exists independently of this change, it's just slightly more noticeable due to it; and 2) there's no quick fix which doesn't add technical debt. Some of the display/frontend/text scanner/query parser classes could use some cleanup to facilitate this; I've ran into similar problems with the most recently scanned text a few times before also.

toasted-nutbread:
> I noticed it doesn't work on the search page though, and stays on top.

Updated with f83e6d04696dd12982e76abb4761d06b421bdac7.

> I'm also wondering if this could cause a situation where results are appended from multiple sources, but I didn't see that happening.

Forgot to address this, but this shouldn't be able to happen. There is another unique token being used to ensure that another call hasn't been started.

ext/bg/js/translator.js
```diff
@@ -194,7 +188,7 @@ class Translator {
 
         const strayDefinitions = defaultDefinitions.filter((definition, index) => !mergedByTermIndices.has(index));
         for (const groupedDefinition of dictTermsGroup(strayDefinitions, dictionaries)) {
-            groupedDefinition.expressions = [{expression: groupedDefinition.expression, reading: groupedDefinition.reading}];
+            groupedDefinition.expressions = [Translator.createExpression(expression, reading)];
```
siikamiika: `groupedDefinition.expressions = [Translator.createExpression(groupedDefinition.expression, groupedDefinition.reading)];`
or destructure above

siikamiika:
The header seems to overlap with entries when hotkeys (alt+up/down or mouse wheel) are used. I used to solve this by adding static padding to the top of each entry, but there could be other good ways to approach the problem.

ext/bg/js/translator.js
toasted-nutbread: Fixed with 019db21d05349003e1fc98c6dc52c927db1e5a22

toasted-nutbread:
> 
> 
> The header seems to overlap with entries when hotkeys (alt+up/down or mouse wheel) are used. I used to solve this by adding static padding to the top of each entry, but there could be other good ways to approach the problem.

Fixed with d0b66d96ac54076f4b222f6ac881f49b1d19830e

siikamiika:
This is a pretty large change so maybe it's good to give it a bit more time and testing, but at the moment it looks good to me

ext/mixed/display-templates.html
```diff
@@ -0,0 +1,81 @@
+<!DOCTYPE html><html><head></head><body>
+
+<template id="term-entry-template"><div class="entry" data-type="term">
+    <div class="entry-header1">
```
siikamiika: Add the class `expression-scan-toggle` or change the `ignoreNodes` class name to `entry-header1` in `search-frontend.js` and `popup-nested.js`

ext/mixed/display-templates.html
```diff
@@ -0,0 +1,81 @@
...
+    <div class="term-definition-tag-list tag-list"></div>
+    <div class="term-definition-only-list"></div>
+    <ul class="term-glossary-list"></ul>
+</li></template>
+<template id="term-definition-only-template"><span class="term-definition-only"></span></template>
+<template id="term-glossary-item-template"><li class="term-glossary-item"><span class="term-glossary"></span></li></template>
+<template id="term-reason-template"><span class="term-reason"></span></template>
+
+<template id="kanji-entry-template"><div class="entry" data-type="kanji">
+    <div class="entry-header1">
```
siikamiika: Same thing about `ignoreNodes`

ext/mixed/display-templates.html
toasted-nutbread: `term-expression-text` more directly corresponds to the old `expression-scan-toggle`; do you think the entire header should be omitted also? I think in practice, the effect will be the same unless there is Japanese text in the tags or something.

siikamiika: I think it's fine the way you did it, because this feature is mainly useful for preventing a popup from showing up when you try to click a kanji. It also makes sense that `.entry-header1` encapsulates both the expressions and frequencies.

siikamiika: Actually it seems that another one is needed for `.kanji-glyph` below

toasted-nutbread: Updated to `source-text` style and added it to both the relevant templates. This should make it easier to extend in the future if necessary.

toasted-nutbread:
Fixed the selector.

> Another issue is that the last (or only) expression requires moving the mouse cursor above the term before the audio button and tags appear. This isn't such a huge issue on desktop, but I'd imagine that it's almost impossible to play such audio on touch devices without accidentally opening a kanji definition.

Can you explain this further? Not sure what exactly you mean by "above the term", since I can just mouseover the term to show the buttons. To my understanding, it functions the same as how it used to.

I can (and probably should) make it so that if there is only one term in the list, the buttons are always shown by default.

siikamiika:
> Can you explain this further? Not sure what exactly you mean by "above the term", since I can just mouseover the term to show the buttons. To my understanding, it functions the same as how it used to.

Before:
before

After:
after

> I can (and probably should) make it so that if there is only one term in the list, the buttons are always shown by default.

Sure, unless it looks weird that only some of the terms have the buttons visible by default. I didn't test this before.

toasted-nutbread:
Added the missing space and also enabled persistent action buttons when only one term is present.

siikamiika:
Looking good :+1:

siikamiika:
I just noticed that scanning `<textarea>`s and `<input>`s still causes the results to refresh when scanning the same text again. Probably something to do with the imposter element. Not a huge issue, so not opening a new ticket for this

siikamiika:
The refreshing probably happens because `TextSourceRange.equals` returns `false` when comparing the ranges on two different imposter elements after the first one has been removed. Maybe it could be fixed by extending its lifetime and passing it around, but that would increase the risk of having the site filled with orphaned imposter elements.

Issue #325, opened 2020-01-16
Too much padding above results


Geniusssmit:
Why so much space here? on 1.8 everything was ok
![chrome_vH6odbXo24](https://user-images.githubusercontent.com/24941285/72537330-01cd5700-388d-11ea-8347-715c5460281e.jpg)

siikamiika:
This has been fixed in https://github.com/FooSoft/yomichan/pull/317, but it's not merged yet. I'll see if there's something to fix before it can be merged.

PR #326, opened 2020-01-16
use TextScanner in QueryParser

siikamiika:
Continuation of https://github.com/FooSoft/yomichan/pull/299.

Fixes blinking results in https://github.com/FooSoft/yomichan/pull/317 so that it can be merged and used to fix https://github.com/FooSoft/yomichan/issues/325.

Also enables scanning query parser text on touchscreen devices.

`TextScanner` could potentially be used in `Display` as well, but it's more work and has less benefits because it only uses click events.

### DIFF
--- a/ext/bg/js/search-query-parser.js
+++ b/ext/bg/js/search-query-parser.js
@@ -17,73 +17,47 @@
  */
 
 
-class QueryParser {
+class QueryParser extends TextScanner {
     constructor(search) {
+        super(document.querySelector('#query-parser'), [], [], []);
         this.search = search;
-        this.pendingLookup = false;
-        this.clickScanPrevent = false;
 
         this.parseResults = [];
         this.selectedParser = null;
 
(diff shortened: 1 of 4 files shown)


ext/bg/js/search-query-parser.js
```diff
@@ -17,73 +17,47 @@
...
     constructor(search) {
+        super(document.querySelector('#query-parser'), [], [], []);
         this.search = search;
-        this.pendingLookup = false;
-        this.clickScanPrevent = false;
 
         this.parseResults = [];
         this.selectedParser = null;
 
         this.queryParser = document.querySelector('#query-parser');
```
siikamiika: ESLint doesn't like setting `this.queryParser` before `super()` (`no-this-before-super`), so `document.querySelector` is used twice. It could be avoided by creating a new local constant, but that would make it more complex.

PR #330, opened 2020-01-25
Query parser helper disabling

### LINKED ISSUE #324: [Feature Request] Text spacing and furigana toggles
Add a toggle for text spacing in the search page and under text parsing options. 

Allow reading mode to be set to none
image

siikamiika:
Resolves #324

### DIFF
--- a/ext/mixed/js/japanese.js
+++ b/ext/mixed/js/japanese.js
@@ -80,6 +80,8 @@ function jpConvertReading(expressionFragment, readingFragment, readingMode) {
                 }
             }
             return readingFragment;
+        case 'none':
+            return null;
         default:
             return readingFragment;
     }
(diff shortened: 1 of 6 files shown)


ext/mixed/js/japanese.js
```diff
@@ -80,6 +80,8 @@ function jpConvertReading(expressionFragment, readingFragment, readingMode) {
                 }
             }
             return readingFragment;
+        case 'none':
+            return null;
```
toasted-nutbread: Just to confirm: `null` is already handled correctly by anything that uses this functionality?

siikamiika: It ends up in Handlebars.js after going through `QueryParser.processParseResultForDisplay`. I didn't see any errors when testing this.

PR #332, opened 2020-01-25
Fix imposter scanning

toasted-nutbread:
Fixes https://github.com/FooSoft/yomichan/pull/317#issuecomment-578397929

### DIFF
--- a/ext/fg/js/source.js
+++ b/ext/fg/js/source.js
@@ -25,14 +25,16 @@ const IGNORE_TEXT_PATTERN = /\u200c/;
  */
 
 class TextSourceRange {
-    constructor(range, content, imposterContainer) {
+    constructor(range, content, imposterContainer, imposterSourceElement) {
         this.range = range;
+        this.rangeStartOffset = range.startOffset;
         this.content = content;
         this.imposterContainer = imposterContainer;
+        this.imposterSourceElement = imposterSourceElement;
     }
 
     clone() {
-        return new TextSourceRange(this.range.cloneRange(), this.content, this.imposterContainer);
+        return new TextSourceRange(this.range.cloneRange(), this.content, this.imposterContainer, this.imposterSourceElement);
     }
 
     cleanup() {
@@ -55,6 +57,7 @@ class TextSourceRange {
     setStartOffset(length) {
         const state = TextSourceRange.seekBackward(this.range.startContainer, this.range.startOffset, length);
         this.range.setStart(state.node, state.offset);
+        this.rangeStartOffset = this.range.startOffset;
         this.content = state.content;
         return length - state.remainder;
     }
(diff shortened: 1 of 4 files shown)


ext/fg/js/source.js
```diff
@@ -79,7 +82,17 @@ class TextSourceRange {
...
+        if (other === null) {
+            return false;
+        }
+        if (this.imposterSourceElement !== null) {
+            return (
+                this.imposterSourceElement === other.imposterSourceElement &&
+                this.rangeStartOffset === other.rangeStartOffset
+            );
+        } else {
+            return this.range.compareBoundaryPoints(Range.START_TO_START, other.range) === 0;
```
siikamiika: Is the comparison swapped here on purpose? I think `other` is the newer `TextSourceRange` here and `this.range` might have already lost its element. I don't know how `compareBoundaryPoints` behaves when called on a range with a removed element vs when it's given one as an argument.

ext/mixed/js/text-scanner.js
```diff
@@ -299,11 +299,11 @@ class TextScanner {
             }
 
             const textSource = docRangeFromPoint(x, y, this.options.scanning.deepDomScan);
-            if (this.textSourceCurrent !== null && this.textSourceCurrent.equals(textSource)) {
-                return;
-            }
-
             try {
+                if (this.textSourceCurrent !== null && this.textSourceCurrent.equals(textSource)) {
```
siikamiika: If functionality is affected, it could also be changed here if that makes more sense code wise

siikamiika:
I tested on regular `<input type="text">` and Google Translate `<textarea>`, seemed to work fine

ext/fg/js/source.js
toasted-nutbread: I would hope that there's no difference between the two, besides one being the negative of the other. I swapped it because it more closely matches the order of the this/other ops of the `equals` function.

https://dom.spec.whatwg.org/#dom-range-compareboundarypoints

ext/mixed/js/text-scanner.js
toasted-nutbread: There's only two places in the codebase that `equals` is used, and there wasn't anything apparent that this change would break. The function is still performing the equality check, the only difference is that if an imposter is used, the imposter is allowed to be a different element since the original is removed.

The only case where this might be a bit odd is if you scan an input, that input changes, and then the user scans the same _position_ of that input. The text at the location would be different, but `equals` would return `true`. Technically this is also the case for text elements, but those are arguably less likely to change. In both cases, the side effects of this rare case would just be that the user has to move the scanning cursor somewhere else (±1 character) and then return to the original position.

ext/fg/js/source.js
siikamiika: I agree this looks more logical, but I was thinking the decision to do it the other way around could have been due to a bug in the past. Anyway I tested it and it worked fine so I'll approve it like this. Just wanted to hear your point of view.

ext/mixed/js/text-scanner.js
siikamiika: > The only case where this might be a bit odd is if you scan an input, that input changes, and then the user scans the same position of that input. The text at the location would be different, but equals would return true. Technically this is also the case for text elements, but those are arguably less likely to change. In both cases, the side effects of this rare case would just be that the user has to move the scanning cursor somewhere else (±1 character) and then return to the original position.

True, didn't think of that. Shouldn't be a big issue as you said.

Issue #262, opened 2019-10-20
APIs between Yomichan and the operating system


siikamiika:
So, Yomichan started as a Rikai replacement with Anki functionality, but over the years, it's accumulated a lot of useful features. I saw this issue https://github.com/FooSoft/yomichan/issues/237 and started thinking that there actually isn't much that's needed outside of Yomichan itself when reading text.

It has come up a few times that Yomichan would be useful as a general text reading tool for text that doesn't originate from websites. While it would be almost impossible to make Yomichan's scan work the same way in all applications on all platforms as suggested in https://github.com/FooSoft/yomichan/issues/146, it's not all that difficult to hook into the text of a specific application such as visual novels or video players and their subtitles.

Currently it's possible to open the search page with a link that also works outside of the browser thanks to https://github.com/FooSoft/yomichan/issues/247, and you can use a hack with [Clipboard Inserter](https://addons.mozilla.org/en-US/firefox/addon/clipboard-inserter/) (I did the same with WebSockets [here](https://github.com/siikamiika/scripts/blob/master/userscripts/ws-inserter.html)) to have all the text coming from an application outside of the browser available for scanning with Yomichan. Discussion about the use case in this issue: https://github.com/FooSoft/yomichan/issues/158

With the background out of the way, what I'm getting at is that there is a lot that could be done with an API in Yomichan that takes in raw Japanese text and inputs it to the search page. I'm aware that many of the features in Yomichan are already far from the initial goal of the extension and this goes even further. If nothing else, I'm writing down the idea for a web application that uses Yomichan as a dictionary.

Here are some ideas for the possible API implementation:

- Probably the easiest step would be to enable searching clipboard contents directly from the search page similar to Clipboard Inserter.
    - Because one might be copying an entire sentence, one possibility would be to actually parse the sentence ([JGlossator](http://jglossator.sourceforge.net/), [my Chinese parser](https://github.com/siikamiika/language-learning/blob/master/chinese/cantonese/canto_parser/canto_parser.py#L97), mention in https://github.com/FooSoft/yomichan/issues/153), or simply show it below the input box where it could be scanned and clicked.
- WebSockets could be used as a generic way to feed text into the search page so that the search page itself connects to a server.
- I found [this](https://github.com/kzahel/web-server-chrome) when wondering if the browser itself could run a web server, but that feels more unsafe
- [Native messaging](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Native_messaging).

[...]

PR #333, opened 2020-01-25
Native popup windows

siikamiika:
Uses clipboard as a text source for now, but adding this to the API discussed in #262 is going to be done in a separate PR later.

Refactoring a new class `ClipboardMonitor` out of `search.js` and `backend.js` is planned, but I'm opening a PR in case there are bigger problems to fix first.

Right now it works by opening the search page in a native popup window. It could make sense to hide text input and disable the regular clipboard monitor when the search page is opened in a popup, but query parser is useful even there.

### DIFF
--- a/ext/bg/js/backend.js
+++ b/ext/bg/js/backend.js
@@ -629,6 +629,10 @@ class Backend {
                 chrome.tabs.create({url});
                 return;
             case 'popup':
+                if (!isObject(chrome.windows)) {
+                    // chrome.windows not supported (e.g. on Firefox mobile)
+                    return;
+                }
                 if (this.popupWindow !== null) {
                     const callback = () => this.checkLastError(chrome.runtime.lastError);
                     chrome.windows.remove(this.popupWindow.id, callback);
--- a/ext/bg/js/backend.js
+++ b/ext/bg/js/backend.js
@@ -585,8 +585,7 @@ class Backend {
     async _onCommandSearch(params) {
         const {mode, query} = params || {};
 
-        const optionsContext = {depth: 0};
-        const options = await this.getOptions(optionsContext);
+        const options = await this.getOptions(this.optionsContext);
         const {popupWidth, popupHeight} = options.general;
 
         const baseUrl = chrome.runtime.getURL('/bg/search.html');
--- a/ext/bg/js/backend.js
+++ b/ext/bg/js/backend.js
@@ -600,7 +600,7 @@ class Backend {
     // Command handlers
 
     async _onCommandSearch(params) {
-        const {mode, query} = params || {};
+        const {mode, query} = params || {mode: 'sameTab'};
 
         const options = await this.getOptions(this.optionsContext);
         const {popupWidth, popupHeight} = options.general;
--- a/ext/bg/js/backend.js
+++ b/ext/bg/js/backend.js
@@ -34,6 +34,9 @@ class Backend {
 
         this.clipboardPasteTarget = document.querySelector('#clipboard-paste-target');
         this.popupWindow = null;
+        this.clipboardPopupTimerId = null;
+        this.clipboardInterval = 250;
+        this.clipboardPreviousText = null;
 
         this.apiForwarder = new BackendApiForwarder();
     }
@@ -121,6 +124,20 @@ class Backend {
             this.mecab.startListener();
         } else {
             this.mecab.stopListener();
+        }
+
+        window.clearInterval(this.clipboardPopupTimerId);
+        if (options.general.enableClipboardPopups) {
+            this.clipboardPopupTimerId = setInterval(() => {
+                this._onApiClipboardGet()
+                .then((result) => {
+                    if (this.clipboardPreviousText === result) {
+                        return;
+                    }
+                    this._onCommandSearch({mode: 'popup', query: result});
+                    this.clipboardPreviousText = result;
+                });
+            }, this.clipboardInterval);
         }
     }
 
--- a/ext/bg/js/backend.js
+++ b/ext/bg/js/backend.js
@@ -619,7 +619,9 @@ class Backend {
             if (tab !== null) {
                 await Backend._focusTab(tab);
                 if (queryParams.query) {
-                    await new Promise((resolve) => chrome.tabs.update(tab.id, {url}, resolve));
+                    await new Promise((resolve) => chrome.tabs.sendMessage(
+                        tab.id, {action: 'searchQueryUpdate', params: {query: queryParams.query}}, resolve
+                    ));
                 }
                 return true;
             }
--- a/ext/bg/js/backend.js
+++ b/ext/bg/js/backend.js
@@ -630,10 +630,10 @@ class Backend {
                 if (this.popupWindow !== null) {
                     chrome.windows.remove(this.popupWindow.id);
                 }
-                chrome.windows.create(
+                this.popupWindow = await new Promise((resolve) => chrome.windows.create(
                     {url, width: popupWidth, height: popupHeight, type: 'popup'},
-                    (popupWindow) => { this.popupWindow = popupWindow; }
-                );
+                    resolve
+                ));
                 return;
         }
     }
--- a/ext/bg/js/backend.js
+++ b/ext/bg/js/backend.js
@@ -603,8 +603,10 @@ class Backend {
         const {popupWidth, popupHeight} = options.general;
 
         const baseUrl = chrome.runtime.getURL('/bg/search.html');
-        const queryString = (query && query.length > 0) ? `?query=${encodeURIComponent(query)}` : '';
-        const url = baseUrl + queryString;
+        const queryParams = {mode};
+        if (query && query.length > 0) { queryParams.query = query; }
+        const queryString = new URLSearchParams(queryParams).toString();
+        const url = `${baseUrl}?${queryString}`;
 
         switch (mode) {
             case 'sameTab':
--- a/ext/bg/js/backend.js
+++ b/ext/bg/js/backend.js
@@ -628,7 +628,8 @@ class Backend {
                 return;
             case 'popup':
                 if (this.popupWindow !== null) {
-                    chrome.windows.remove(this.popupWindow.id);
+                    const callback = () => this.checkLastError(chrome.runtime.lastError);
+                    chrome.windows.remove(this.popupWindow.id, callback);
                 }
                 this.popupWindow = await new Promise((resolve) => chrome.windows.create(
                     {url, width: popupWidth, height: popupHeight, type: 'popup'},
--- a/ext/bg/js/backend.js
+++ b/ext/bg/js/backend.js
@@ -522,13 +522,30 @@ class Backend {
     }
 
     async _onApiClipboardGet() {
-        const clipboardPasteTarget = this.clipboardPasteTarget;
-        clipboardPasteTarget.value = '';
-        clipboardPasteTarget.focus();
-        document.execCommand('paste');
-        const result = clipboardPasteTarget.value;
-        clipboardPasteTarget.value = '';
-        return result;
+        /*
+        Notes:
+            document.execCommand('paste') doesn't work on Firefox.
+            This may be a bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1603985
+            Therefore, navigator.clipboard.readText() is used on Firefox.
+
+            navigator.clipboard.readText() can't be used in Chrome for two reasons:
+            * Requires page to be focused, else it rejects with an exception.
+            * When the page is focused, Chrome will request clipboard permission, despite already
+              being an extension with clipboard permissions. It effectively asks for the
+              non-extension permission for clipboard access.
+        */
+        const browser = await Backend._getBrowser();
+        if (browser === 'firefox' || browser === 'firefox-mobile') {
+            return await navigator.clipboard.readText();
+        } else {
+            const clipboardPasteTarget = this.clipboardPasteTarget;
+            clipboardPasteTarget.value = '';
+            clipboardPasteTarget.focus();
+            document.execCommand('paste');
+            const result = clipboardPasteTarget.value;
+            clipboardPasteTarget.value = '';
+            return result;
+        }
     }
 
     async _onApiGetDisplayTemplatesHtml() {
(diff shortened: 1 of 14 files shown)


ext/bg/js/backend.js
```diff
@@ -101,6 +102,10 @@ class Backend {
         }
     }
 
+    _onClipboardText(text) {
+        this._onCommandSearch({mode: 'popup', query: text});
+    }
+
```
siikamiika: Not sure if this is a good way to call the search command

toasted-nutbread: Could make a utility function(s) for the various search functionalities if you want to avoid extra (de)structuring, but I don't think it's a big concern; looks fine as-is to me

ext/bg/js/search.js
```diff
@@ -202,6 +198,17 @@ class DisplaySearch extends Display {
         }
     }
 
+    onCopy() {
+        // ignore copy from search page
+        this.clipboardMonitor.setPreviousText(document.getSelection().toString().trim());
+    }
```
siikamiika: I might add a similar check to the global clipboard monitor that can be used to toggle the feature on/off within the browser without affecting other software. It could be less reliable because it takes some time to serialize API commands and meanwhile the `setTimeout` can finish.

ext/bg/js/backend.js
```diff
@@ -565,23 +597,45 @@ class Backend {
...
+            case 'newTab':
+                chrome.tabs.create({url});
+                return;
+            case 'popup':
+                if (this.popupWindow !== null) {
+                    chrome.windows.remove(this.popupWindow.id);
+                }
+                chrome.windows.create(
+                    {url, width: popupWidth, height: popupHeight, type: 'popup'},
+                    (popupWindow) => { this.popupWindow = popupWindow; }
```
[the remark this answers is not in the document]
toasted-nutbread: I think promise wrappers is the predominant means of handling chrome* callback functions, and it's not been a lot of excess code to write it either, so +1 to that idea. It's usually as simple as:

```js
await new Promise((resolve) => chrome.whatever(args, resolve));
```

toasted-nutbread:
I get this warning when doing the following:
`Unchecked runtime.lastError: No window with id: #.`
* Enable the popups
* Select JP text to show a popup
* Close popup
* Select new JP text which creates a new popup

Basically, the property just has to be read in a non-trivial manner. See:
https://github.com/FooSoft/yomichan/blob/05bb4aef71c5c062b451dd04dcd55111f7ba7377/ext/bg/js/backend.js#L75
https://github.com/FooSoft/yomichan/blob/05bb4aef71c5c062b451dd04dcd55111f7ba7377/ext/bg/js/backend.js#L216-L218

I've tested this before, and I think if it's too simple, it will be optimized as a no-op and the read will not occur. E.g.
```js
chrome.runtime.lastError;
const x = chrome.runtime.lastError; // unused x
const callback = () => {}; /* ... */ callback(chrome.runtime.lastError);
```

ext/bg/js/search.js
siikamiika: I won't do this for now because of its complexity and small benefit (has to run in every frame)

siikamiika:
> Basically, the property just has to be read in a non-trivial manner.

I think this is fixed in https://github.com/FooSoft/yomichan/pull/333/commits/f91a18489d06d4a8214b2a1a1195f0e9226c3ff4.

I consider the feature ready, let me know if there are any issues to fix.

ext/bg/js/backend.js
```diff
@@ -565,23 +597,48 @@ class Backend {
...
+        const {popupWidth, popupHeight} = options.general;
+
+        const baseUrl = chrome.runtime.getURL('/bg/search.html');
+        const queryParams = {mode};
+        if (query && query.length > 0) { queryParams.query = query; }
+        const queryString = new URLSearchParams(queryParams).toString();
+        const url = `${baseUrl}?${queryString}`;
+
+        switch (mode) {
+            case 'sameTab':
```
toasted-nutbread: `'existingOrNewTab'` or `'existingTab'` may be more semantically clear than `'sameTab'`. ('same' as what?)

ext/bg/js/backend.js
```diff
@@ -565,23 +597,48 @@ class Backend {
...
+                return;
+            case 'newTab':
+                chrome.tabs.create({url});
+                return;
+            case 'popup':
+                if (this.popupWindow !== null) {
+                    const callback = () => this.checkLastError(chrome.runtime.lastError);
+                    chrome.windows.remove(this.popupWindow.id, callback);
+                }
+                this.popupWindow = await new Promise((resolve) => chrome.windows.create(
```
toasted-nutbread: Not necessary to add anything until we encounter something, but keep an eye out in case you ever see a `runtime.lastError` here. I've seen cases where `runtime.lastError` will be set despite the docs not listing anything, particularly on Firefox for Android. (I know this feature shouldn't work on Android, but could happen for strange edge cases also.)

ext/bg/js/backend.js
```diff
@@ -565,23 +597,48 @@ class Backend {
...
-            } catch (e) {
-                // NOP
-            }
+                chrome.tabs.create({url});
+                return;
+            case 'newTab':
+                chrome.tabs.create({url});
+                return;
+            case 'popup':
+                if (this.popupWindow !== null) {
```
toasted-nutbread: Firefox for Android doesn't support `chrome.windows`. While this feature shouldn't be enabled on Android, validate `chrome.windows` is valid before using.

https://github.com/FooSoft/yomichan/blob/d5708de4eed15567e14e0c1fd4998561eee1680e/ext/bg/js/backend.js#L705-L708

ext/bg/js/clipboard-monitor.js
```diff
@@ -0,0 +1,78 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+
+class ClipboardMonitor {
+    constructor() {
+        this.timerId = null;
+        this.timerToken = null;
+        this.interval = 250;
```
toasted-nutbread: No action, just out of curiosity I tested something that's maybe helpful to know. `setTimeout` and `setInterval` are both throttled on webpages when they are in the background, usually to about 1000ms minimum. However, this is not the case for the background page; it will always be able to have these lower intervals.

ext/bg/js/search.js
```diff
@@ -166,13 +162,12 @@ class DisplaySearch extends Display {
...
-        const query = DisplaySearch.getSearchQueryFromLocation(window.location.href) || '';
-        if (this.query !== null) {
-            if (this.isWanakanaEnabled()) {
-                this.setQuery(window.wanakana.toKana(query));
-            } else {
-                this.setQuery(query);
-            }
+        const {query='', mode=''} = DisplaySearch.parseQueryStringFromLocation(window.location.href);
+        document.documentElement.dataset.searchMode = mode;
+        if (this.isWanakanaEnabled()) {
```
toasted-nutbread: Could be simplified
```js
this.setQuery(this.isWanakanaEnabled() ? window.wanakana.toKana(query) : query);
```

toasted-nutbread: This is also done on line 206, and could also be simplified on line 83. Maybe update `setQuery` to do this automatically? It doesn't look like there's a current need for this, but a `setQueryRaw` or something similar could be added if necessary to have one which doesn't modify input; or a second `raw` (or something) parameter to `setQuery`.

ext/bg/js/search.js
```diff
@@ -69,7 +66,7 @@ class DisplaySearch extends Display {
                         this.wanakanaEnable.checked = false;
                     }
                     this.wanakanaEnable.addEventListener('change', (e) => {
-                        const query = DisplaySearch.getSearchQueryFromLocation(window.location.href) || '';
```
toasted-nutbread: Out of context for this change, but the listener callbacks for `this.wanakanaEnable.addEventListener('change', ...)` here and `this.clipboardMonitorEnable.addEventListener('change', ...)` on line 97 should be moved into their own member function definitions on `DisplaySearch`; I think it would make it easier to read the `prepare` function.

ext/bg/settings.html
```diff
@@ -134,6 +134,10 @@ <h3>General Options</h3>
                     <label><input type="checkbox" id="enable"> Enable content scanning</label>
                 </div>
 
+                <div class="checkbox">
+                    <label><input type="checkbox" id="enable-clipboard-popups"> Enable native popups when copying Japanese text</label>
```
toasted-nutbread: This option should be hidden for Android, or at least disabled.

If you want a multi-purpose solution, you could do something similar to the existing CSS rules `data-show-for-browser="firefox-mobile"`, except like `data-hide-for-browser`.

ext/bg/js/backend.js
```diff
@@ -121,6 +130,12 @@ class Backend {
         } else {
             this.mecab.stopListener();
         }
+
+        if (options.general.enableClipboardPopups) {
```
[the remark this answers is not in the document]
siikamiika: Yes, as far as I know. `stop()` doesn't do anything dangerous and `start()` has a token that is used to abort the callback in case it has changed.

ext/bg/js/search.js
siikamiika: Yeah, this method exploded out of hand when I added query parser. What's the reason for checking nulls on those elements btw? I just replicated it here.

ext/bg/settings.html
siikamiika: Added `data-hide-for-browser`, but noticed that it's visible for a while before the settings page script has run. Shouldn't be a big issue

ext/bg/js/search.js
siikamiika: I updated it to line 206 and also to the checkbox event listener because the value is checked directly from the checkbox

ext/bg/js/backend.js
```diff
@@ -613,11 +613,14 @@ class Backend {
...
-                        url2.startsWith(url) &&
-                        (url2.length === url.length || url2[url.length] === '?' || url2[url.length] === '#')
+                        url2.startsWith(baseUrl) &&
+                        (url2.length === baseUrl.length || url2[baseUrl.length] === '?' || url2[baseUrl.length] === '#')
                     ));
                     if (tab !== null) {
                         await Backend._focusTab(tab);
+                        if (queryParams.query) {
+                            await new Promise((resolve) => chrome.tabs.update(tab.id, {url}, resolve));
+                        }
```
siikamiika: Changed this in d4e74a05723b1b4347c51606bcee61dc72e34e51 because the new `mode` query parameter broke focusing an existing tab. Now it's also possible to set a new query on an existing search tab.

ext/bg/js/search.js
toasted-nutbread: I suppose there's not really a good purpose as it currently stands. While technically it's possible for elements to be removed or modified from other scripts or dev tools, it's probably not necessary to handle these cases. (I may have a habit of doing this due to using other environments in which null checks are more necessary.)

ext/bg/js/backend.js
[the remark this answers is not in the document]
toasted-nutbread: It looks like something like
```js
for (let i = 0; i< 1000000; ++i) { this.clipboardMonitor.start(); }
```
would actually start 1000000 timers and then get the clipboard 1000000 times. Only the last call would progress beyond that, but there's some potential waste. I would recommend just clearing any pre-existing timers in the `start` call also.

ext/bg/js/search.js
siikamiika: Ok. We can look at that again when this is refactored. The nested checks add some extra indentation level.

ext/bg/js/backend.js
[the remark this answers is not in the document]
siikamiika: Fixed by calling `stop()` before moving on in `start()`

toasted-nutbread:
This is another issue that I run into on Windows; not sure how it manifests on other systems.
* The window is repeatedly re-opened rather than reusing an existing window. There is a noticeable animation of the window closing and then re-opening, as well as the taskbar icon flashes each time, and the window is repositioned.
* Chrome also seems to have a bug where the window content isn't rendered properly until the window is moved or resized. The window will appear white until this happens, and this only seems to happen in the case where an existing window is closed and then a new window is created; I haven't seen it only happen the first time.
* On Firefox, the position of the new window is only _sometimes_ reset. Sometimes it will keep the position of the previous popup window. The blank window issue that I see in Chrome is not present on Firefox either.

test

siikamiika:
> The window is repeatedly re-opened rather than reusing an existing window. There is a noticeable animation of the window closing and then re-opening, as well as the taskbar icon flashes each time, and the window is repositioned.

This could probably be handled with something similar to what's done in d4e74a05723b1b4347c51606bcee61dc72e34e51. I'll look into it. There's a problem with using query strings without a hash though: the page is reloaded when the URL is updated. This causes a smaller flash to happen. It's probably ok or even desirable to push a new history entry when updating the URL of an existing popup.

> Chrome also seems to have a bug where the window content isn't rendered properly until the window is moved or resized. The window will appear white until this happens, and this only seems to happen in the case where an existing window is closed and then a new window is created; I haven't seen it only happen the first time.

Never happened to me before, maybe something about graphics drivers?

> On Firefox, the position of the new window is only _sometimes_ reset. Sometimes it will keep the position of the previous popup window. The blank window issue that I see in Chrome is not present on Firefox either.

I was actually planning to add settings for specifying the absolute position of the popup and use those as fallback when none are specified when opening a new popup, such as when clipboard change triggers it. Maybe I'll add it before merging.

toasted-nutbread:
> This could probably be handled with something similar to what's done in [d4e74a0](https://github.com/FooSoft/yomichan/commit/d4e74a05723b1b4347c51606bcee61dc72e34e51). I'll look into it. There's a problem with using query strings without a hash though: the page is reloaded when the URL is updated. This causes a smaller flash to happen. It's probably ok or even desirable to push a new history entry when updating the URL of an existing popup.

You may have to do something like post a message to that tab which tells the tab to update its query, history state, etc, rather than having the backend try to force a URL change.

> Never happened to me before, maybe something about graphics drivers?

¯\\\_(ツ)\_/¯ I haven't looked into it much.

> I was actually planning to add settings for specifying the absolute position of the popup and use those as fallback when none are specified when opening a new popup, such as when clipboard change triggers it. Maybe I'll add it before merging.

My main issue is just that windows are recreated every time, which has significant overhead. If an existing window is reused, the positioning inconsistencies aren't a big deal to me at this point. So you can add that to this PR or a future one, whichever you like.

siikamiika:
The popup window should now be reused when possible, and I also added some checks to prevent confusing the popup with a regular search page tab.

siikamiika:
Also tested this on a Firefox on Mac, seemed to work

PR #334, opened 2020-01-26
add scannable tags for expression and reading

### LINKED ISSUE #53: [feature request] Make definition's furigana clickable (to use it as a search string for a new lookup)
As discussed on #23

siikamiika:
Fixes  #53 

Clicking the tags performs the same scan as clicking glossary and it can be hard to hit the first character on the tags due to their small font size. It would require some refactoring in `onTermLookup` to make clicks always search the contents of the entire tag. Maybe it could be facilitated with `TextScanner`.

### DIFF
--- a/ext/mixed/display-templates.html
+++ b/ext/mixed/display-templates.html
@@ -77,5 +77,6 @@
 
 <template id="tag-template"><span class="tag"><span class="tag-inner"></span></span></template>
 <template id="tag-frequency-template"><span class="tag" data-category="frequency"><span class="tag-inner"><span class="term-frequency-dictionary-name"></span><span class="term-frequency-separator"></span><span class="term-frequency-value"></span></span></template>
+<template id="tag-search-template"><span class="tag" data-category="search"></span></template>
 
 </body></html>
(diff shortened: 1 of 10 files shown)


siikamiika:
I'll wait for #339/#340 resolution first before adding this new feature to avoid conflicts

ext/mixed/display-templates.html
```diff
@@ -77,5 +77,6 @@
 
 <template id="tag-template"><span class="tag"><span class="tag-inner"></span></span></template>
 <template id="tag-frequency-template"><span class="tag" data-category="frequency"><span class="tag-inner"><span class="term-frequency-dictionary-name"></span><span class="term-frequency-separator"></span><span class="term-frequency-value"></span></span></template>
+<template id="tag-search-template"><span class="tag" data-category="search"></span></template>
```
siikamiika: Could be useful to abstract away `tag-inner` and add a separator as well for scanning so that it doesn't have to be handled in `DisplayGenerator` for each tag type

siikamiika: Separator because while double click is delimited by the block element, scanning isn't. Also https://github.com/FooSoft/yomichan/pull/286

Issue #337, opened 2020-01-28
Bug with termTags and Translator data format


toasted-nutbread:
There is currently a bug with `termTags` not being shown properly when result grouping mode is `merge`, causing the tags to never be shown. The issue stems from the `termTags` not being consistently defined, and `DisplayGenerator` is reading the property from the wrong object.

https://github.com/FooSoft/yomichan/blob/05bb4aef71c5c062b451dd04dcd55111f7ba7377/ext/mixed/js/display-generator.js#L74-L77

Above, `DisplayGenerator.createTermExpression` will read `termTags` from `details.expressions[n]` in `merge` mode, and from `details` otherwise. However, in `merge` mode, `termTags` is defined on `details`, while all of the other data used to generate the expression is still on `details.expressions[n]`.

In my opinion, there is a larger issue at hand here where the data returned from the translator backend returns an inconsistent format, a format which changes based on the _presentation_ mode. The `termTags` that is being read by `DisplayGenerator.createTermExpression` can be either `null` or an array, which is usually empty, which is another inconsistency.

It would be nice if these data structures could be standardized, but this also presents several problems. The biggest issue I see is the handlebars templates used to generate the Anki fields. These templates rely on the current data format, so if the internal data structure is updated to be more consistent, this format would still need a way to be converted to the old format to provide backwards compatibility with Anki template generation.

However, in the long run, this is probably worth it. It's somewhat difficult to figure out what some of the logic in translator.js does due to this issue, as well as some confusing naming conventions (deinflections, definitions, definitions.definitions, expressions.expressions, etc.) Now that HTML templates are used for display content generation, it should also be easier to update the internal data format, as all of the presentation styling is handled using CSS now.

I'll look into a way to fix the tag bug in the short term, but as a longer term goal, a standardized data format may make development and debugging easier.

siikamiika:
> There is currently a bug with termTags not being shown properly when result grouping mode is `merge`

Do you mean when it is *not* `merge`? I can reproduce with 少ない, but アジア works.

I agree that the data format is problematic, but if I remember correctly it was decided in #84/#95 to be the easier way to move forward. I should have probably considered it technical debt to be fixed as soon as possible after the feature had been deployed. Now there are special cases everywhere causing bugs like #336.

Do you think it would make sense to replace
```
expression
reading
furiganaSegments
termTags
expressions
```
with a `terms` array that contains everything currently in `expressions` in `merge` mode, and move
```
definitionTags
dictionary
```
[...]

PR #338, opened 2020-01-29
Translator refactoring

toasted-nutbread:
Fixes the bug noted in #337. First commit is some refactoring, the second is the fix.

### DIFF
--- a/ext/bg/js/dictionary.js
+++ b/ext/bg/js/dictionary.js
@@ -159,11 +159,14 @@ function dictTermsGroup(definitions, dictionaries) {
 }
 
 function dictTermsMergeBySequence(definitions, mainDictionary) {
-    const definitionsBySequence = {'-1': []};
+    const sequencedDefinitions = new Map();
+    const nonSequencedDefinitions = [];
     for (const definition of definitions) {
-        if (mainDictionary === definition.dictionary && definition.sequence >= 0) {
-            if (!definitionsBySequence[definition.sequence]) {
-                definitionsBySequence[definition.sequence] = {
+        const sequence = definition.sequence;
+        if (mainDictionary === definition.dictionary && sequence >= 0) {
+            sequencedDefinition = sequencedDefinitions.get(sequence);
+            if (typeof sequencedDefinition === 'undefined') {
+                sequencedDefinition = {
                     reasons: definition.reasons,
                     score: Number.MIN_SAFE_INTEGER,
                     expression: new Set(),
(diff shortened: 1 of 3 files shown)


ext/bg/js/dictionary.js
```diff
@@ -159,11 +159,14 @@ function dictTermsGroup(definitions, dictionaries) {
...
-    const definitionsBySequence = {'-1': []};
+    const sequencedDefinitions = new Map();
+    const nonSequencedDefinitions = [];
     for (const definition of definitions) {
-        if (mainDictionary === definition.dictionary && definition.sequence >= 0) {
-            if (!definitionsBySequence[definition.sequence]) {
-                definitionsBySequence[definition.sequence] = {
+        const sequence = definition.sequence;
+        if (mainDictionary === definition.dictionary && sequence >= 0) {
+            sequencedDefinition = sequencedDefinitions.get(sequence);
```
siikamiika: `let`

siikamiika:
I noticed another bug when testing: "Main dictionary for merged mode" is shown by "Active profile" instead of "Modifying profile". If you import a dictionary without `termTags` before one that has them and set `mainDictionary` to the one that has them without doing it for each profile while they're active, you will end up in a situation where the popup has wrong dictionary as the main dictionary. Example of the situation in attached config.

[yomichan-settings-2020-01-29-20-45-03.txt](https://github.com/FooSoft/yomichan/files/4130674/yomichan-settings-2020-01-29-20-45-03.txt)

ext/bg/js/dictionary.js
toasted-nutbread: I forgot that eslint won't catch that due to `"no-undef": "off"`

toasted-nutbread:
> 
> 
> I noticed another bug when testing: "Main dictionary for merged mode" is shown by "Active profile" instead of "Modifying profile". If you import a dictionary without `termTags` before one that has them and set `mainDictionary` to the one that has them without doing it for each profile while they're active, you will end up in a situation where the popup has wrong dictionary as the main dictionary. Example of the situation in attached config.
> 
> [yomichan-settings-2020-01-29-20-45-03.txt](https://github.com/FooSoft/yomichan/files/4130674/yomichan-settings-2020-01-29-20-45-03.txt)

Fixing this in #343 because it's outside the scope of this change.

siikamiika:
I'm still not sure why #337 originally happened, but the changes look good to me.

toasted-nutbread:
> I'm still not sure why #337 originally happened, but the changes look good to me.

I can't replicate the exact scenario either, but this does handle an issue where the format of sequenced definition data didn't match non-sequenced definitions, despite them all being added to the same results data.

PR #340, opened 2020-01-30
Fix tag double click

### LINKED ISSUE #339: Double clicking on compact glossary items or tags can select multiple words
Sometimes I don't know the definition of an English word in glossary, and double click it to look it up using GoldenDict. When it's not delimited by punctuation, the selection will extend past glossary item boundaries. Space also counts, but the glossary items don't have any space around them. I wonder if there's a CSS rule for this, because it would be quite dirty to actually add space around elements in JS. Editing a `<template>` could be a good solution too.

simplescreenrecorder-2020-01-30_00 43 44

toasted-nutbread:
Fixes #339 by changing display to `inline-block` and adding an inner `display: block` element. This also makes triple click work in case there are spaces. This change also improves the tag layout slightly. The vertical spacing is a bit better. [Old](https://user-images.githubusercontent.com/11037431/73411135-4962d280-42d2-11ea-82dd-1e9717fe3b08.png) vs [New](https://user-images.githubusercontent.com/11037431/73411137-4cf65980-42d2-11ea-834e-fc77d297fa48.png); Open in two new tabs and cycle between them.

### DIFF
--- a/ext/mixed/css/display.css
+++ b/ext/mixed/css/display.css
@@ -251,6 +251,7 @@ button.action-button {
 
 .term-reasons>.term-reason+.term-reason-separator+.term-reason:before {
     content: " \00AB  "; /* The two spaces is not a typo */
+    white-space: pre-wrap;
     display: inline;
 }
 
@@ -403,7 +404,16 @@ button.action-button {
 }
 
 :root[data-compact-glossaries=true] .term-glossary-list>li:not(:first-child):before {
+    white-space: pre-wrap;
     content: " | ";
+    display: inline;
+}
+
+.term-glossary-separator,
+.term-reason-separator {
+    display: inline;
+    font-size: 0;
+    opacity: 0;
 }
 
 /*
@@ -467,14 +477,4 @@ button.action-button {
 .kanji-glossary-list:not([data-count="0"]):not([data-count="1"]) {
     padding-left: 1.4em;
     list-style-type: decimal;
-}
-
-.term-glossary-separator,
-.term-reason-separator {
-    display: inline-block;
-    width: 0;
-    height: 0;
-    font-size: 0;
-    opacity: 0;
-    white-space: pre;
 }
--- a/ext/mixed/css/display.css
+++ b/ext/mixed/css/display.css
@@ -414,6 +414,7 @@ button.action-button {
     display: inline;
     font-size: 0;
     opacity: 0;
+    white-space: pre-wrap;
 }
 
 /*
--- a/ext/mixed/css/display.css
+++ b/ext/mixed/css/display.css
@@ -249,7 +249,7 @@ button.action-button {
     display: inline-block;
 }
 
-.term-reasons>.term-reason+.term-reason:before {
+.term-reasons>.term-reason+.term-reason-separator+.term-reason:before {
     content: " \00AB  "; /* The two spaces is not a typo */
     display: inline;
 }
@@ -467,4 +467,14 @@ button.action-button {
 .kanji-glossary-list:not([data-count="0"]):not([data-count="1"]) {
     padding-left: 1.4em;
     list-style-type: decimal;
+}
+
+.term-glossary-separator,
+.term-reason-separator {
+    display: inline-block;
+    width: 0;
+    height: 0;
+    font-size: 0;
+    opacity: 0;
+    white-space: pre;
 }
--- a/ext/mixed/css/display.css
+++ b/ext/mixed/css/display.css
@@ -208,15 +208,19 @@ button.action-button {
 }
 
 .tag {
-    display: inline;
+    display: inline-block;
     padding: 0.2em 0.6em 0.3em;
     font-size: 75%;
     font-weight: 700;
-    line-height: 1;
+    line-height: 1.25;
     text-align: center;
     white-space: nowrap;
     vertical-align: baseline;
     border-radius: 0.25em;
+}
+
+.tag-inner {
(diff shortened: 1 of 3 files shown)


siikamiika:
Cool! I can do the same for glossary items after knowing how, if you want.

toasted-nutbread:
The same technique wouldn't work for glossary and reasons since they need inline flow. A hidden zero-width does the trick though.

siikamiika:
Hmm, for some reason that zero-width doesn't work on Firefox. I tested and it worked on Chromium, though.

toasted-nutbread:
> Hmm, for some reason that zero-width doesn't work on Firefox. I tested and it worked on Chromium, though.

You're right, I'll see if there's a better solution.

siikamiika:
Moving the separator to the beginning, swapping zero-width space to regular space and allowing user-select fixed everything but the first gloss selecting also the tag left of it:

```diff
diff --git a/ext/mixed/css/display.css b/ext/mixed/css/display.css
index 98b932e..486ac37 100644
--- a/ext/mixed/css/display.css
+++ b/ext/mixed/css/display.css
@@ -474,5 +474,4 @@ button.action-button {
     display: inline;
     font-size: 0;
     opacity: 0;
-    user-select: none;
 }
diff --git a/ext/mixed/display-templates.html b/ext/mixed/display-templates.html
index b46c18b..242a9e1 100644
--- a/ext/mixed/display-templates.html
+++ b/ext/mixed/display-templates.html
@@ -31,8 +31,8 @@
     <ul class="term-glossary-list"></ul>
 </li></template>
 <template id="term-definition-only-template"><span class="term-definition-only"></span></template>
-<template id="term-glossary-item-template"><li class="term-glossary-item"><span class="term-glossary"></span><span class="term-glossary-separator">&#x200b;</span></li></template>
-<template id="term-reason-template"><span class="term-reason"></span><span class="term-reason-separator">&#x200b;</span></template>
+<template id="term-glossary-item-template"><li class="term-glossary-item"><span class="term-glossary-separator"> </span><span class="term-glossary"></span></li></template>
+<template id="term-reason-template"><span class="term-reason-separator"> </span><span class="term-reason"></span></template>

 <template id="kanji-entry-template"><div class="entry" data-type="kanji">
     <div class="entry-header1">
```

toasted-nutbread:
Try the new solution, which is to use a newline. Makes triple click work too!

siikamiika:
> Makes triple click work too!

Now triple click will select *less* than double click sometimes. Another Firefox bug?

simplescreenrecorder-2020-01-30_04 31 49

toasted-nutbread:
firefox please

toasted-nutbread:
Isolated bug and reported. I'll look to see if there's another better workaround where triple click works, otherwise I'll probably just use a space.

https://github.com/toasted-nutbread/firefox-text-selection-bug
https://toasted-nutbread.github.io/firefox-text-selection-bug/
https://bugzilla.mozilla.org/show_bug.cgi?id=1612236

toasted-nutbread:
Reverted to something that works a bit better on both browsers. Triple click doesn't work but oh well.

siikamiika:
I actually didn't even notice the bug you reported to Mozilla before I watched the gif again. I should have been more specific, sorry for the confusion. What I meant was that it's possible for a double click on the leftmost word to also select the tag text left of it as long as it doesn't contain punctuation such as `)` that JMdict dictionaries do, but doing a third click in quick succession does what you'd expect and selects the contents of the block, removing selection from the tag. You can add a quick custom CSS rule to test without renaming your JMdict and importing it again:
```css
.tag[data-category=dictionary] {
  display: none;
}
```

After your latest change everything else seems to work well, but in https://github.com/FooSoft/yomichan/pull/340/commits/d6e4230d62c9628faa45c681e81e54ef275a68be I noticed extra line breaks (empty space) after definitions that only had a single glossary item.

siikamiika:
It seems to work if you move the separator before the definition so that it's there before the tags and change its content to a `&nbsp;` instead (don't ask why newline or regular space doesn't work)

```diff
diff --git a/ext/mixed/display-templates.html b/ext/mixed/display-templates.html
index e762f66..9f4e04e 100644
--- a/ext/mixed/display-templates.html
+++ b/ext/mixed/display-templates.html
@@ -31,8 +31,8 @@
     <ul class="term-glossary-list"></ul>
 </li></template>
 <template id="term-definition-only-template"><span class="term-definition-only"></span></template>
-<template id="term-glossary-item-template"><li class="term-glossary-item"><span class="term-glossary"></span><span class="term-glossary-separator">&#10;</span></li></template>
-<template id="term-reason-template"><span class="term-reason"></span><span class="term-reason-separator">&#10;</span></template>
+<template id="term-glossary-item-template"><li class="term-glossary-item"><span class="term-glossary-separator">&nbsp;</span><span class="term-glossary"></span></li></template>
+<template id="term-reason-template"><span class="term-reason-separator">&nbsp;</span><span class="term-reason"></span></template>
 
 <template id="kanji-entry-template"><div class="entry" data-type="kanji">
     <div class="entry-header1">
```

toasted-nutbread:
Test out 88436ca3bc4bab06a2f37c706c02cad2d094b30f. Changed newline to regular space, moved term-glossary-separator before, and added `white-space: pre-wrap;`, which avoids the need to use `&nbsp;`.

siikamiika:
Works on both Firefox and Chrome, with or without compact glossaries. Thanks for your patience!

Issue #258, opened 2019-10-19
Code style for private variables/functions


toasted-nutbread:
Open questions about code style regarding the intended scope of use for variables and functions. I don't want to make significant stylistic changes unless others collaborators think there is merit to them.

1. One thing I've noticed while implementing some new features is that it's not immediately apparent whether class functions and variables are intended to be private/internal, or if they are also designed to be used externally. Same can apply to variables/functions defined on the global object: is the function intended to be publicly used or is it a local helper function. This can make it take longer to make changes, since you not only have to check references within the same file, but other files as well.

   A common convention is to prefix internals with a ```_``` as a hint that this is a private member. Does anyone else feel that this would be a beneficial change? Or any other motivations for or against?

2. There is a similar namespacing issue with functions being declared on the global object, which seems to be why some functions have prefixes like ```api*``` or ```audio*```. Would it make sense to create classes which encapsulate some of these functionalities?

   For example, in the case of the ```api*``` functions, would it make sense to do something like create a ```FrontendAPI``` and ```BackendAPI``` class? The public functions would be the same, but putting it into a class might make the encapsulation more clear. The end result might be something like:
   ```js
   const api = new FrontendAPI(); // or new BackendAPI();
   // ...
   await api.optionsGet(); // instead of old apiOptionsGet();
   ```
   This is somewhat similar to how the ```AnkiConnect/AnkiNull``` and ```Popup/PopupProxy``` pairs of classes work.

siikamiika:
I think many of the namespacing issues come from the nature of JavaScript in web browsers that is adding scripts into HTML in some order and hoping that other code running on the page doesn't have any name conflicts with it and that your own code doesn't immediately run something that depends on stuff that's in a script tag below.

JavaScript has been improving steadily on this and the most promising feature IMO that doesn't require a transpiler is [ES6 modules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import). I tried to use them for a feature in the past, but had to [revert](https://github.com/FooSoft/yomichan/pull/172/commits/0ee6d05cb15a8bc94f120c2896e7537fd969aaed) because content scripts don't support modules yet. So we're either stuck with the traditional way of using JavaScript or have to use a transpiler that turns everything into a single script file. Modules would allow the following:

```javascript
import * as name from "module-name";
```

[...]

PR #344, opened 2020-02-01
Eslint strengthening

toasted-nutbread:
Due to something I overlooked in https://github.com/FooSoft/yomichan/pull/338#discussion_r372594574 because it wasn't caught by eslint, I have gone through and updated the lint rules to be a bit more strict with regards to globals. Yomichan runs in the browser environment and uses `<script>` tags for loading scripts, so eslint doesn't check cross-file globals since `require` isn't used.

This change does require external globals to be defined at the top of the file to explicitly indicate that they are used. While this does add a bit of overhead, I think that it's a small trade-off for stronger code validity. Note that some of the global declarations I added are multi-line. This is primarily just to make it more readable. Functions on different lines are kind of grouped together with regard to similarity or source, but it's nothing formal.

After updating the lint rules, I checked the entire project and found one error (33e7f825e4497434fc3fd33d6ca3c35cc3aa2bf2) and one stylistic issue (ac5cf9b237f53e6e33934d9593125c2fca131b50) and resolved them.

---

One thing that comes up again relates to the discussion from #258, particularly point 2. Currently, all global functions must be explicitly declared. For example:
https://github.com/FooSoft/yomichan/blob/eaf46c4476160133810abf7f81e5fd512abf870b/ext/mixed/js/display.js#L19-L23
This can be somewhat repetitive when referring to functions which come from the same file. It could be potentially more simple if there was some object which contained all of these functions, such as `api.kanjiFind, api.termsFind`, and then only `api` has to be declared as global. This is a style similar to what most libraries do (jQuery, WanaKana).

I don't think it's mandatory to make such a change, but this style of declaring globals may be a point in favor of that.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,7 +1,14 @@
 {
+    "root": true,
     "extends": "eslint:recommended",
     "parserOptions": {
-        "ecmaVersion": 8
+        "ecmaVersion": 8,
+        "sourceType": "script"
+    },
+    "env": {
+        "browser": true,
+        "es2017": true,
+        "webextensions": true
     },
     "ignorePatterns": [
         "/ext/mixed/lib/",
@@ -18,6 +25,7 @@
         "no-constant-condition": "off",
         "no-undef": "off",
         "no-unused-vars": ["error", {"vars": "local", "args": "after-used", "argsIgnorePattern": "^_", "caughtErrors": "none"}],
+        "no-unused-expressions": "error",
         "no-var": "error",
         "prefer-const": ["error", {"destructuring": "all"}],
         "quote-props": ["error", "consistent"],
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -49,6 +49,7 @@
                 "stringReverse": "readonly",
                 "promiseTimeout": "readonly",
                 "stringReplaceAsync": "readonly",
+                "parseUrl": "readonly",
                 "EventDispatcher": "readonly",
                 "EXTENSION_IS_BROWSER_EDGE": "readonly"
             }
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -23,7 +23,8 @@
         "no-case-declarations": "error",
         "no-const-assign": "error",
         "no-constant-condition": "off",
-        "no-undef": "off",
+        "no-global-assign": "error",
+        "no-undef": "error",
         "no-unused-vars": ["error", {"vars": "local", "args": "after-used", "argsIgnorePattern": "^_", "caughtErrors": "none"}],
         "no-unused-expressions": "error",
         "no-var": "error",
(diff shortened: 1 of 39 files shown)


siikamiika:
Is there a way to get ESLint to notify about unused `global` defines in the comment on top of the file? I tried removing `no-var` and using `var globalFunction;` instead but ESLint didn't complain when it was unused.

toasted-nutbread:
`"no-unused-vars": ["error", {"vars": "all"}]` will show errors for that case, but that also affects other functions "exported" from the file.

toasted-nutbread:
Or at least not with Yomichan's current style.

[IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE) could be used and enforced with [no-implicit-globals](https://eslint.org/docs/rules/no-implicit-globals), and then `"no-unused-vars": ["error", {"vars": "all"}]` could be used. Personally, that's how I usually write scripts in the context of web browsers since I don't like implicit globals, but it would be a fairly large stylistic change.

siikamiika:
Ok, my concern is that while this helps to avoid accidentally assigning things to the global scope, it will be hard to keep track of what's actually still used after changing global function names or removing them. I still think this is an overall improvement.

Seems that there are still some places in `fg` where the `global` definitions are not added, are you planning to do it in this PR?

toasted-nutbread:
> Ok, my concern is that while this helps to avoid accidentally assigning things to the global scope, it will be hard to keep track of what's actually still used after changing global function names or removing them. I still think this is an overall improvement.

The best thing to do given the current structure is probably just intermittent checking to see that all globals are still used, which would be somewhat of a manual process. I don't think it would harm that much if a global is accidentally left in a file; if it's still used, it should cause an error. If functions are removed any, a find-all operation should usually be done to ensure that all uses of the function are removed also. This would also find the global definition. This basically already has to be done for any global var/function changes prior to this commit anyways.

> Seems that there are still some places in `fg` where the `global` definitions are not added, are you planning to do it in this PR?

Some files didn't use external globals.

toasted-nutbread:
I've cherry picked the errors into master since they are really independent of eslint changes.
33e7f825e4497434fc3fd33d6ca3c35cc3aa2bf2 (old) => d5708de4eed15567e14e0c1fd4998561eee1680e (new)
ac5cf9b237f53e6e33934d9593125c2fca131b50 (old) => f85f92c665736c759aa9d325d6d36f236fc21485 (new)

toasted-nutbread:
PR has now been updated after the latest merges, and also fixed the missing declarations.

siikamiika:
Sorry for leaving this PR hanging, I was just disappointed in what JS can do by default. The errors about `fg` missing globals are now gone for me and other things look good in general.

toasted-nutbread:
No problem, there were some things that I missed and it's better that I waited until after the other PRs were merged.

PR #345, opened 2020-02-02
JSON schema improvements

toasted-nutbread:
Improves the capabilities of the JsonSchema classes.
* Adds support for conditionals
* Adds support for `allOf`, `anyOf`, `oneOf`, and `not`
* Better array validation
* Simplified validation errors (now just using `throw` instead of `return`)
* Adds improved information about validation errors (path to the offending field is tracked)
* Fixes some bugs

### DIFF
--- a/ext/bg/js/json-schema.js
+++ b/ext/bg/js/json-schema.js
@@ -138,7 +138,13 @@ class JsonSchemaProxyHandler {
                 }
 
                 const additionalProperties = schema.additionalProperties;
-                return JsonSchemaProxyHandler.isObject(additionalProperties) ? additionalProperties : null;
+                if (additionalProperties === false) {
+                    return null;
+                } if (JsonSchemaProxyHandler.isObject(additionalProperties)) {
+                    return additionalProperties;
+                } else {
+                    return JsonSchemaProxyHandler._unconstrainedSchema;
+                }
             }
             case 'array':
             {
@@ -490,6 +496,8 @@ class JsonSchemaProxyHandler {
         return typeof value === 'object' && value !== null && !Array.isArray(value);
     }
 }
+
+JsonSchemaProxyHandler._unconstrainedSchema = {};
 
 class JsonSchema {
     static createProxy(target, schema) {
--- a/ext/bg/js/json-schema.js
+++ b/ext/bg/js/json-schema.js
@@ -131,19 +131,19 @@ class JsonSchemaProxyHandler {
             case 'object':
             {
                 const properties = schema.properties;
-                if (properties !== null && typeof properties === 'object' && !Array.isArray(properties)) {
+                if (JsonSchemaProxyHandler.isObject(properties)) {
                     if (Object.prototype.hasOwnProperty.call(properties, property)) {
                         return properties[property];
                     }
                 }
 
                 const additionalProperties = schema.additionalProperties;
-                return (additionalProperties !== null && typeof additionalProperties === 'object' && !Array.isArray(additionalProperties)) ? additionalProperties : null;
+                return JsonSchemaProxyHandler.isObject(additionalProperties) ? additionalProperties : null;
             }
             case 'array':
             {
                 const items = schema.items;
-                return (items !== null && typeof items === 'object' && !Array.isArray(items)) ? items : null;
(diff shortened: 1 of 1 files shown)


siikamiika:
Tested and working. Noticed that some of the validations are not used in `JsonSchemaProxyHandler.getValidValueOrDefault`, is this on purpose?

```javascript
const schemaValidate = (value) => JsonSchema.validate(value, {
    "allOf": [
        {"type": "number"},
        {
            "anyOf": [
                {"minimum": 10, "maximum": 100},
                {"minimum": -100, "maximum": -10}
            ]
        },
        {
            "oneOf": [
                {"multipleOf": 3},
                {"multipleOf": 5},
            ]
        },
        {
            "not": [
                {"multipleOf": 20}
            ]
        }
    ]
});

const jsValidate = (value) => typeof value === 'number' &&
    ((value >= 10 && value <= 100) || (value >= -100 && value <= -10)) &&
    (((value % 3 === 0) || (value % 5 === 0)) && (value % 15 !== 0)) &&
    value % 20 !== 0;

for (let i = -11; i < 25; i++) {
    try {
        validate(i);
        console.log(i, jsValidate(i) ? 'OK' : 'ERROR');
    } catch (e) {
        console.log(i, jsValidate(i) ? 'ERROR' : 'OK', `${e}`);
    }
}
```

```
-11 OK Error: 0 oneOf schemas matched
-10 OK
-9 OK Error: 0 anyOf schemas matched
-8 OK Error: 0 anyOf schemas matched
-7 OK Error: 0 anyOf schemas matched
-6 OK Error: 0 anyOf schemas matched
-5 OK Error: 0 anyOf schemas matched
-4 OK Error: 0 anyOf schemas matched
-3 OK Error: 0 anyOf schemas matched
-2 OK Error: 0 anyOf schemas matched
-1 OK Error: 0 anyOf schemas matched
0 OK Error: 0 anyOf schemas matched
1 OK Error: 0 anyOf schemas matched
2 OK Error: 0 anyOf schemas matched
3 OK Error: 0 anyOf schemas matched
4 OK Error: 0 anyOf schemas matched
5 OK Error: 0 anyOf schemas matched
6 OK Error: 0 anyOf schemas matched
7 OK Error: 0 anyOf schemas matched
8 OK Error: 0 anyOf schemas matched
9 OK Error: 0 anyOf schemas matched
10 OK
11 OK Error: 0 oneOf schemas matched
12 OK
13 OK Error: 0 oneOf schemas matched
14 OK Error: 0 oneOf schemas matched
15 OK Error: 2 oneOf schemas matched
16 OK Error: 0 oneOf schemas matched
17 OK Error: 0 oneOf schemas matched
18 OK
19 OK Error: 0 oneOf schemas matched
20 OK Error: not[0] schema matched
21 OK
22 OK Error: 0 oneOf schemas matched
23 OK Error: 0 oneOf schemas matched
24 OK
```

toasted-nutbread:
I think the reason `JsonSchemaProxyHandler.getValidValueOrDefault` doesn't some of the extra validations is because it was primarily designed for correcting non-conformant options values, and the options had a fairly simple schema which didn't use most of the schema features (minItems, number validation, etc.) These can be added at some point in the future.

Also, I'll probably add some of these tests as standalone unit tests in the future and add a generic validation script similar to [dictionary-validate.js](https://github.com/FooSoft/yomichan/blob/31fef733fac7bb943b20028d5f77b7e759ffde99/test/dictionary-validate.js).

toasted-nutbread:
Basic unit tests added in 174d9e7429a0f7a45538e56a594ce627239d80c3

siikamiika:
Thanks! Do you know if there's some way to make automated tests that run in a browser-like environment (selenium and such) to test things like popups?

toasted-nutbread:
I haven't looked into it that much, but I haven't found anything for it yet. Some of the tests can probably be done using node, such as term lookups, if there is a way to have simulate IndexedDB. rikaichamp does something similar to this, though I've not looked at it in great detail.

I have a test case HTML file that has a bunch of things that I have tested before, although the process is entirely manual. I'm thinking I may want to clean that up and add something like that to the repo, and new cases can be added as we go. That will be better than nothing, and if we find a good way to automate that at some point, we can try that. However, I'm willing to bet that it's difficult to do fully automated testing on web extensions.

Issue #322, opened 2020-01-11
[Feature Request] Custom dictionary with images


FrenchYann:
The other day, while reading the news, I thought it'd be nice to be able to see a map of japan's prefectures when looking one up.

So I made a proof of concept
![yomichan-image-dict-demo](https://user-images.githubusercontent.com/2552413/72205933-b419ae80-3488-11ea-8fd7-e018dd229635.png)

The two problems I ran into are:
- I couldn't find any documentation on the dictionary format itself (is there any? I'm still not quite sure what `sequenced` does) 
- I had to hack into the plugin's template to remove html escaping

My suggestion is, the dictionary format could be extended to disable html escaping on a per dictionary or per entry basis. Something like:
in `index.json`
```js
{
  "title":"example",
  "format":3,
  "revision":"jmdict4",
  "sequenced":true,
  "escape_html":false  // default is true
}
```
or
it `term_bank_#.json`
```js
[
   [
      expression, 
      reading, 
      definitionTags, 
      deinflectionRules, 
      score, 
      [
         "<img src=\"data:image/png;base64,iVBORw0K...\">"
      ], 
      sequence, 
      termTags,
      true   // don't escape html for this entry, should be backward compatible
   ], 
   ...
]
```
@toasted-nutbread is already messing with templates here #317 so if you find it to be a good idea, it should probably happen after this pull request is handled.

I think it could be a nice feature to not only learn prefectures or city of japan, but also other famous landmarks, celebrities, characters, specialties, dishes, plants, etc. situations for which a picture tells more than thousand words.

toasted-nutbread:
Few questions/comments:
* Are you planning on making the dictionary file yourself / is the data already something you have available for this?
* Allowing unescaped HTML to be injected would lead to issues with code injection, which poses a security flaw. If you only want to show an image (maybe with text description also), something like the following could be done:
  * Images (.png, .gif, .jpg, etc.) can be added directly to the .zip file
  * Each term can optionally have something like an `"image": "filename.jpg"` field
* Images should likely be stored in a separate database table for space concerns, and each term would only store the optional key pointer to that image.
* May have to add an additional Anki template option for the image.
* I believe `sequenced` has something to do with related terms, with the same reading or something. I know it's used for grouping scan/search results together.

Overall this seems doable so long as someone has the data to provide for the dictionary.

FrenchYann:
[...]

Issue #61, opened 2017-07-03
[feature request] Show pitch accent information


seanblue:
It would be nice if you could find a resource to include pitch accent information for words. For example, [Weblio](http://www.weblio.jp/content/%E5%85%88%E7%94%9F) shows something like `[0]` and `[1]` to indicate which mora has the downstep. And [OJAD](http://www.gavo.t.u-tokyo.ac.jp/ojad/search/index/sortprefix:accent/narabi1:kata_asc/narabi2:accent_asc/narabi3:mola_asc/yure:visible/curve:invisible/details:invisible/limit:20/word:%E5%85%88%E7%94%9F) shows a bar above high pitch mora and the downstep if there is one.

**Weblio**:
![image](https://user-images.githubusercontent.com/5884067/27803132-f6ae53b4-5ff5-11e7-8ebe-d364f4d2737f.png)

**OJAD**:
![image](https://user-images.githubusercontent.com/5884067/27803141-039b46c2-5ff6-11e7-92ca-5eea14491bf4.png)

seanblue:
Any thoughts on this?

FooSoft:
This (along with frequency lists) is something I am definitely adding support for after I work out the details about how to store this information in the extension and where to source it from. If you have any recommendations for quality offline sources of pitch accent information it would help. I will not support online sources since it's brittle, slow, and errorprone.

skliew:
I think the one in Rikaisama (http://rikaisama.sourceforge.net) is pretty good. It's stored in a SQLite database that is bundled with its installation.

seanblue:
@FooSoft I wonder if you could contact [OJAD ](http://www.gavo.t.u-tokyo.ac.jp/ojad) and ask for their raw data. They have pitch accents for a significant number of words (probably just about everything), plus professionally recorded audio for many verbs and adjectives, plus their conjugations.

They have a mail to link on their homepage via the "Inquiries" button. Given that the data is from a university, they might be willing to give out the data since it's for a free, open source project.

faneca:
Hi, I made a hack to make this (kinda) work, with some drawbacks (explained below), by using [Kanjium](https://github.com/mifunetoshiro/kanjium/)'s data ([this file](https://github.com/mifunetoshiro/kanjium/blob/master/data/source_files/raw/accents.txt), in particular).

[...]

Issue #297, opened 2019-12-02
Request to add Vietnamese kanjidic2 and jmdict


dreamofi:
Greetings!

I have just compiled a Vietnamese version for kanjidc2 and jmdict and would like to add them to Yomichan repo.
The files can be found here:
[jmdict_vietnamese](https://github.com/dreamofi/yomichan-Vietnamese-dictionary/raw/master/vietnameseDict/jmdict_vietnamese.zip)
[kanjidic_vietnamese](https://github.com/dreamofi/yomichan-Vietnamese-dictionary/raw/master/vietnameseDict/kanjidic_vietnamese.zip)

My repo can be found [here](https://github.com/dreamofi/yomichan-Vietnamese-dictionary/tree/master/)
With my limited capability, the dictionaries' format is not quite pretty, but I think it will be of some value to Vietnamese people learning Japanese.

Thank you for making a great addon for Japanese learner!

siikamiika:
Hi! Thanks for the contribution.

The convention with the dictionaries is that they aren't included with Yomichan (in repo or release), but rather the user installs them separately. @FooSoft hosts some free dictionaries at his homepage [here](https://foosoft.net/projects/yomichan/index.html#dictionaries) and [Yomichan Import](https://github.com/FooSoft/yomichan-import) is used to convert dictionaries to the Yomichan format.

Some comments:

- You seem to have created your own tool that does the same thing as Yomichan Import, which is fine, but it might be easier to handle future database changes in Yomichan if it was ported to Yomichan Import. I'm not necessarily asking you to do it.
- I installed the dictionary and it seems to have "\n" in places where there should be newlines. That should be an easy fix if the source data is consistently in this format.
- Your repository mentions that it's compiled from open source dictionaries, but I can't find a license for them. Yomichan already supports proprietary EPWING J-J dictionaries through Yomichan Import, but it's expected that the user has their own dictionary and does the conversion themselves. I think the license or a link to the original source should be mentioned somewhere when the dictionary data and not just the conversion tool is distributed.

After fixing at least the newline bug and licensing issues, I would approve linking to your repository from the readme.

FooSoft:
@siikamiika pretty much hit the nail on the head there 😄 

* My biggest concern is licensing. Even though a dictionary may be free, there can still be some strings attached; for example, in the case of JMDICT, I have to properly credit the original source, or Yomichan would technically not be in compliance.
[...]

PR #346, opened 2020-02-02
Dictionary schemas

toasted-nutbread:
Defines schemas for the dictionary JSON files that can be imported.

I am adding this for a few reasons:
* In a future commit, these should be used to validate data before importing, which should help to prevent unwanted data from getting into the database.
* This can be used as a sort of documentation of the file format, in case anyone wants to generate their own dictionaries from a custom data set (#322). Will also serve to document future changes and additions to the types of data supported (e.g. pitch accents #61).

Note that the descriptions of everything are based on how I am seeing the values be used in the code. If anyone has any better descriptions for anything, feel free to suggest.

Also adds a validation script which can be run using node. I have validated the following dictionaries:
```
innocent_corpus.zip
jmdict_dutch.zip
jmdict_english.zip
jmdict_french.zip
jmdict_german.zip
jmdict_hungarian.zip
jmdict_russian.zip
jmdict_slovenian.zip
jmdict_spanish.zip
jmdict_swedish.zip
jmnedict.zip
kanjidic_english.zip
kanjidic_french.zip
kanjidic_portuguese.zip
kanjidic_spanish.zip
kanjium_pitchaccent.zip
kireicake.zip
NHK日本語発音アクセント辞典.zip
アニメとドラマの頻度リスト.zip
三省堂　スーパー大辞林.zip
大辞泉.zip
広辞苑第六版 • 付属資料.zip
故事ことわざの辞典.zip
明鏡国語辞典.zip
研究社　新和英大辞典　第５版.zip
```

There were only a few weird cases across all the dictionaries where the data was not what I expected. Particularly, there were a few cases where the main expression was an empty string, and a few times `null` was used for some tags instead of an empty string. The schemas are relaxed enough to handle this.

This depends on #345 being merged first.

### DIFF
--- a/ext/bg/data/dictionary-index-schema.json
+++ b/ext/bg/data/dictionary-index-schema.json
@@ -1,0 +1,40 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "type": "object",
+    "description": "Index file containing information about the data contained in the dictionary.",
+    "required": [
+        "title",
+        "revision"
+    ],
+    "properties": {
+        "title": {
+            "type": "string",
+            "description": "Title of the dictionary."
+        },
+        "revision": {
+            "type": "string",
+            "description": "Revision of the dictionary. This value is only used for displaying information."
+        },
+        "sequenced": {
+            "type": "boolean",
+            "default": false,
+            "description": "Whether or not this dictionary can be used as the primary dictionary. Primary dictionaries typically contain term/expression definitions."
+        },
+        "format": {
+            "type": "integer",
+            "description": "Format of data found in the JSON data files."
+        },
+        "version": {
+            "type": "integer",
+            "description": "Alias for format."
+        }
+    },
+    "anyOf": [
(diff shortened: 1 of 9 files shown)


siikamiika:
Tested this on some of my dictionaries including the ones posted in https://github.com/FooSoft/yomichan/issues/297, and they seemed to validate. I also edited one term bank in JMDict by removing a sequence, and the schema catched it with `JsonSchemaValidationError: Array length too short`. A `null` value was also catched.

siikamiika:
The schemas themselves also make sense. I was first confused by `kanji-bank-v1`, because I've never worked with the older format.

toasted-nutbread:
> The schemas themselves also make sense. I was first confused by `kanji-bank-v1`, because I've never worked with the older format.

Me neither; the schemas are based on the import process for using v1. I didn't look into what v2 was, but my guess is that it was probably similar to v3 minus some of the array entries.

PR #350, opened 2020-02-06
query parser html templates

siikamiika:
In the spirit of https://github.com/FooSoft/yomichan/pull/317.

Removes the need to rebuild any templates while developing, but Handlebars is still used for Anki templates.

The `TemplateHandler` class could also be used in `DisplayGenerator`, but I won't do that in this PR.

https://github.com/FooSoft/yomichan/pull/279


siikamiika:
Maybe this is too much magic, but I'm also thinking if the `TemplateHandler` should have a `Proxy` that converts `camelCase` property access to `kebab-case` on the fly :smile: Then fragments could also use a naming convention such as `term-reason-fragment-template` --> `termReasonFragment`.

ext/mixed/js/template-handler.js
```diff
@@ -0,0 +1,53 @@
...
+    constructor(html) {
+        this._templates = new Map();
+        this._html = html;
+        this._doc = null;
+
+        this._initialize();
+    }
+
+    _initialize() {
+        this._doc = new DOMParser().parseFromString(this._html, 'text/html');
```
toasted-nutbread: I would just leave `this._doc` as a temp var `doc` since it's not used afterwards, and similar for `_html`. Everything could just be put into the constructor also, it doesn't really benefit much to abstract it into a private `_initialize` function which isn't `async` and is of very little complexity.

toasted-nutbread:
> Maybe this is too much magic, but I'm also thinking if the `TemplateHandler` should have a `Proxy` that converts `camelCase` property access to `kebab-case` on the fly 😄 Then fragments could also use a naming convention such as `term-reason-fragment-template` --> `termReasonFragment`.
> 
> edit: it could be quite confusing because it creates a copy every time with `importNode` but looks like a property

Yeah, I would avoid having getters which are constructing new objects upon access. I also think it's fine to keep kebab and strings case since we already have to do that for CSS selectors, and this is effectively a CSS selector abstraction.

siikamiika:
Something I thought while doing https://github.com/FooSoft/yomichan/pull/350/commits/9c98c631afde843f8b5ed4c54fc1f4ff0ec3a00a: there are different ways ensure that an object's initialization promise has been fulfilled. In `DisplayGenerator` there's the method `isInitialized` that returns a boolean, but `initialize` is called in the same `async` function in `Display`. There could actually be a bug because I don't see `_isInitialized` being set to `true` anywhere.

Anyway, I think it would make the code easier to reason about if the object would be guaranteed to be initialized when some of its methods are called from another object that is composed of or inherits the object.

I might be missing some performance issues, though. Maybe there's some room for optimization with `Promise.all` when the async dependencies don't depend on each other.

toasted-nutbread:
> There could actually be a bug because I don't see `_isInitialized` being set to `true` anywhere.

Probably true

> Anyway, I think it would make the code easier to reason about if the object would be guaranteed to be initialized when some of its methods are called from another object that is composed of or inherits the object.

It could be possible to make the async prepare functions into private implementation details which the constructor automatically calls, and then it's up to the class to ensure that it awaits the prepare promise before doing any work that depends on it. This would include when the class's public API functions are called.

One downside is that this might incur additional async, as some public functions will now be required to await preparation. This is the case for how `DisplayGenerator` was set up; it only has one async function (`initialize`) so that all of the other internal functions can be non-async. The responsibility of calling the initialize method is passed to the user of the class.

siikamiika:
> It could be possible to make the async prepare functions into private implementation details which the constructor automatically calls, and then it's up to the class to ensure that it awaits the prepare promise before doing any work that depends on it. This would include when the class's public API functions are called.

I thought of that, but it feels kind of overkill to cascade the `async` unless it's made default everywhere. Do you think it would make sense to switch to a single method name like `async prepare` or `async initialize` and always `await` it before inheriting it or using it to compose a class (or do they have a different meaning)? Example what I mean here:
https://github.com/FooSoft/yomichan/blob/17181eb97efcef6a4221397cfe6030b880792e32/ext/fg/js/float.js#L96-L109

What could be done is add `displayGenerator.initialize()` here:
https://github.com/FooSoft/yomichan/blob/17181eb97efcef6a4221397cfe6030b880792e32/ext/mixed/js/display.js#L245-L248

`DisplaySearch` also calls `Display.initialize` in its `prepare`:

https://github.com/FooSoft/yomichan/blob/17181eb97efcef6a4221397cfe6030b880792e32/ext/bg/js/search.js#L48-L51

toasted-nutbread:
I think that if we change things up, we should treat `prepare` as a public setup function which, if present on a class definition, _must_ be called and `await`ed before the class can be properly used. (An exception might be something like non-async `isPrepared`.) This pattern already exists and is older than anything I have added.

The `initialize` methods should likely be removed or renamed to avoid confusion, and made "private" if necessary.

Probably would be good to track this task in https://github.com/FooSoft/yomichan/projects/5 as an issue.

Issue #356, opened 2020-02-12
prepare classes having async dependencies before using them


siikamiika:
> It could be possible to make the async prepare functions into private implementation details which the constructor automatically calls, and then it's up to the class to ensure that it awaits the prepare promise before doing any work that depends on it. This would include when the class's public API functions are called.

I thought of that, but it feels kind of overkill to cascade the `async` unless it's made default everywhere. Do you think it would make sense to switch to a single method name like `async prepare` or `async initialize` and always `await` it before inheriting it or using it to compose a class (or do they have a different meaning)? Example what I mean here:
https://github.com/FooSoft/yomichan/blob/17181eb97efcef6a4221397cfe6030b880792e32/ext/fg/js/float.js#L96-L109

What could be done is add `await displayGenerator.initialize()` here:
https://github.com/FooSoft/yomichan/blob/17181eb97efcef6a4221397cfe6030b880792e32/ext/mixed/js/display.js#L245-L248

`DisplaySearch` also calls `Display.initialize` in its `prepare`:

https://github.com/FooSoft/yomichan/blob/17181eb97efcef6a4221397cfe6030b880792e32/ext/bg/js/search.js#L48-L51

_Originally posted by @siikamiika in https://github.com/FooSoft/yomichan/pull/350#issuecomment-584535338_

siikamiika:
I think that if we change things up, we should treat `prepare` as a public setup function which, if present on a class definition, _must_ be called and `await`ed before the class can be properly used. (An exception might be something like non-async `isPrepared`.) This pattern already exists and is older than anything I have added.

The `initialize` methods should likely be removed or renamed to avoid confusion, and made "private" if necessary.

Probably would be good to track this task in https://github.com/FooSoft/yomichan/projects/5 as an issue.

_Originally posted by @toasted-nutbread in https://github.com/FooSoft/yomichan/pull/350#issuecomment-584980110_

siikamiika:
This is now mostly done in all core classes. There could still be some places that don't follow this style, but since the vast majority does, I think it's enough to resolve the issue.

Issue #353, opened 2020-02-09
Extension security


toasted-nutbread:
I was recently going through Mozilla's [security best practices for extensions](https://extensionworkshop.com/documentation/develop/build-a-secure-extension/) page and noticed a few things that may affect Yomichan.

* **Use the standard extension content security policy (CSP)**
  I've mistakenly tried to adjust this before, but Handlebars uses `eval` (really `Function` constructor) to generate content. Can't do anything about this at this point, but just mentioning it here anyway. #203
* **Add eslint-plugin-no-unsanitized to ESLint**
  This is something that can probably be set up after #350 is added. #352 can also be subsequently updated.
* **Don’t inject moz-extension paths directly**
  Yomichan currently has a fingerprinting risk due to the popup `<iframe>` element.

siikamiika:
> Don’t inject moz-extension paths directly

Agreed. It's good that the `<iframe>` isn't injected before text is scanned, though. I also think that Chrome is not affected by this, because they use the same ID in Chrome Web Store (other than fingerprinting by the fact of having Yomichan installed). What about other identifying data (however small) such as `class` or `data-yomichan-theme`?

toasted-nutbread:
> What about other identifying data (however small) such as `class` or `data-yomichan-theme`?

I don't see that as being as big of a deal, since it's not truly unique per user. Unfortunately, Mozilla's page doesn't really describe how to hide the assets. They mention it's possible to use data urls or srcdoc, but the host webpage can read both. If the iframe is treated as a empty temp frame which Yomichan subsequently injects scripts into, then Yomichan would need a way of correlating that iframe element to the extension frame ID. I ran into this same issue while trying to implement better popup iframe support, but I didn't find a good solution. At least not without making assumptions on how some of the web extension APIs work which is not covered in the documentation.

I also mentioned this briefly in #292, but a closed shadow DOM may help with this, as scripts won't  be able to read the DOM. This could also help isolate the Yomichan styles. (There are potentially issues with [`closed` being able to be bypassed](https://blog.revillweb.com/open-vs-closed-shadow-dom-9f3d7427d1af?gi=23024b71b9d6), but we may not have this issue since the context of the web extension scripts is more secure.)

Also: it does seem possible to manually call `postMessage` from the host webpage on the iframe to trigger the iframe APIs. This may be able to be improved by changing `targetOrigin` to something other than `'*'`, but this will take some experimentation.

siikamiika:
[...]

PR #357, opened 2020-02-13
Simplify display prepare

siikamiika:
I couldn't remove the `isPrepared` calls from `Display` yet. That would require some additional handling in `Popup`. But this simplifies `DisplayGenerator`.

 #356

### DIFF
--- a/ext/fg/js/float.js
+++ b/ext/fg/js/float.js
@@ -45,6 +45,8 @@ class DisplayFloat extends Display {
         }
 
         this.setContentScale(scale);
+
+        window.parent.postMessage('initialized', '*');
     }
 
     onError(error) {
--- a/ext/fg/js/float.js
+++ b/ext/fg/js/float.js
@@ -46,7 +46,7 @@ class DisplayFloat extends Display {
 
         this.setContentScale(scale);
 
-        window.parent.postMessage('initialized', '*');
+        apiForward('popupSetDisplayInitialized');
     }
 
     onError(error) {
--- a/ext/fg/js/float.js
+++ b/ext/fg/js/float.js
@@ -33,6 +33,20 @@ class DisplayFloat extends Display {
         window.addEventListener('message', (e) => this.onMessage(e), false);
     }
 
+    async prepare(options, popupInfo, url, childrenSupported, scale) {
+        await super.prepare(options);
+
+        const {id, depth, parentFrameId} = popupInfo;
+        this.optionsContext.depth = depth;
+        this.optionsContext.url = url;
+
+        if (childrenSupported) {
+            popupNestedInitialize(id, depth, parentFrameId, url);
+        }
+
+        this.setContentScale(scale);
+    }
+
     onError(error) {
         if (this._orphaned) {
             this.setContent('orphaned');
@@ -92,20 +106,6 @@ class DisplayFloat extends Display {
     setContentScale(scale) {
         document.body.style.fontSize = `${scale}em`;
     }
-
-    async initialize(options, popupInfo, url, childrenSupported, scale) {
-        await super.initialize(options);
-
-        const {id, depth, parentFrameId} = popupInfo;
-        this.optionsContext.depth = depth;
-        this.optionsContext.url = url;
-
-        if (childrenSupported) {
-            popupNestedInitialize(id, depth, parentFrameId, url);
-        }
-
-        this.setContentScale(scale);
-    }
 }
 
 DisplayFloat._onKeyDownHandlers = new Map([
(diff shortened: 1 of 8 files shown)


siikamiika:
I noticed that the popup will be empty for the first scan. I don't yet know why it happens, but `Popup._createInjectPromise` looks interesting.

siikamiika:
This fixes the issue, but now the `isInjected` and `setInitialized` distinction is confusing https://github.com/FooSoft/yomichan/pull/357/commits/c0225f1f844376e5f61222a4d8186a2c914026eb

siikamiika:
I read the code a bit and realized that it can actually be desirable not to prepare some components right away. Also related to #353, but it could be wasteful to inject the popup on every page before scanning any text. Not sure yet if it makes sense to check in `Frontend` if `Popup` is injected or keep the checks in `Popup`.

ext/fg/js/popup.js
```diff
@@ -160,7 +154,7 @@ class Popup {
     }
 
     isVisibleSync() {
-        return this._isInjected && (this._visibleOverride !== null ? this._visibleOverride : this._visible);
+        return (this._visibleOverride !== null ? this._visibleOverride : this._visible);
```
siikamiika: This should work because `_visible` won't be true before injection has completed

siikamiika: Tested that screenshot still works

ext/fg/js/popup.js
```diff
@@ -341,10 +333,9 @@ class Popup {
     }
 
     _invokeApi(action, params={}) {
-        if (!this._isInjectedAndLoaded) {
-            throw new Error('Frame not loaded');
+        if (this._container.contentWindow) {
```
siikamiika: This is needed because `Frontend.onSearchClear` calls `Popup.clearAutoPlayTimer` before the popup has initialized. The check could also be in `clearAutoPlayTimer` but it wouldn't make sense without another flag. Could also mean that the call to `onSearchClear` is unnecessary.

siikamiika: The call comes from here:
https://github.com/FooSoft/yomichan/blob/af89659751f72ae282f2c81d797cc224d1077b0a/ext/bg/js/search.js#L250

siikamiika:
Some timings on Chrome for https://github.com/FooSoft/yomichan/pull/357/commits/d7e1ef01d8af4a315a31364eb5138e24a132ea1e (did a few runs and chose the best value, very unscientific):

```
Display dependencies
--------------------

prepare Display dependencies one by one took 9.23999981ms  [Yomichan]
  Sample[0] took 9.23999981ms (100.0%)  [prepare Display dependencies one by one]

prepare Display dependencies Promise.all took 4.44499985ms  [Yomichan]
  Sample[0] took 4.44499985ms (100.0%)  [prepare Display dependencies Promise.all]

DisplaySearch dependencies
--------------------
(the below timings both have Display dependencies with Promise.all)

prepare DisplaySearch dependencies one by one took 11.74999983ms  [Yomichan]
  Sample[0] took 11.74999983ms (100.0%)  [prepare DisplaySearch dependencies one by one]

prepare DisplaySearch dependencies Promise.all took 6.24999986ms  [Yomichan]
  Sample[0] took 6.24999986ms (100.0%)  [prepare DisplaySearch dependencies Promise.all]
```

ext/fg/js/float.js
```diff
@@ -45,6 +45,8 @@ class DisplayFloat extends Display {
         }
 
         this.setContentScale(scale);
+
+        window.parent.postMessage('initialized', '*');
```
toasted-nutbread: Indicating to the popup class that the iframe is fully read is a good idea, but there are probably a few issues with this implementation:

1. Multiple popups can be on the same page due to the nested popup and proxy systems. A unique ID may need to be generated for each popup to make sure the messages are handled by the appropriate location.

2. The message is being posted to the root HTML page. This means that it is not being sent exclusively to the extension, and that it could affect the page the script is running on. For example, if a website is has a handler for an 'initialized' method, this could interfere with its operation.

I haven't tested these claims myself, but the `targetOrigin` being `'*'` likely has the same concerns as what was noted in #353. It may be possible to just have the float frame post a message directly to its tab using [`sendMessage`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/sendMessage) to avoid the concerns with `postMessage`. Still will probably need a unique identifier.

siikamiika: > It may be possible to just have the float frame post a message directly to its tab using [`sendMessage`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/sendMessage) to avoid the concerns with `postMessage`. Still will probably need a unique identifier.

I actually tested `apiForward` that uses this but it only worked for the first popup. Should have probably added the method to `PopupProxy` and `PopupProxyHost` too. They use some sort of ID, maybe that handles sending the message to the correct popup?

toasted-nutbread: Yeah, the existing id of the popup/proxy popup should work.

siikamiika: Should be fixed in https://github.com/FooSoft/yomichan/pull/357/commits/810a7e7d92d15412974810702d954de60453dd31

ext/mixed/js/display.js
```diff
@@ -357,18 +355,14 @@ class Display {
...
 
         if (context.focus !== false) {
             window.focus();
         }
 
-        if (!this.displayGenerator.isInitialized()) {
-            await this.displayGenerator.initialize();
-            if (this.setContentToken !== token) { return; }
-        }
+        if (this.setContentToken !== token) { return; }
```
siikamiika: Just noticed that this is probably not needed unless there's something async before coming into the method. I'll look at it later. The `promiseTimeout` below still needs the token check.

siikamiika: Fixed

PR #358, opened 2020-02-14
General refactoring

toasted-nutbread:
Wanted to get a sanity check on these changes, and it felt like overkill to submit each of these as an individual PR, so this PR contains a bunch of miscellaneous refactorings. Easiest to read on a commit-by-commit basis since the scope of each is different.

PR #362, opened 2020-02-15
More type refactoring

toasted-nutbread:
Kind of a follow-up on #358, but this change is mostly updating more objects to use `Set` or `Map`. There should be some slight improvements to some of the functions in translator.js also.

### DIFF
--- a/ext/bg/js/dictionary.js
+++ b/ext/bg/js/dictionary.js
@@ -177,11 +177,17 @@ function dictTermsMergeBySequence(definitions, mainDictionary) {
 
 function dictTermsMergeByGloss(result, definitions, appendTo=null, mergedIndices=null) {
     const definitionsByGloss = appendTo !== null ? appendTo : new Map();
+
+    const resultExpressionsMap = result.expressions;
+    const resultExpressionSet = result.expression;
+    const resultReadingSet = result.reading;
+    const resultSource = result.source;
+
     for (const [index, definition] of definitions.entries()) {
         const {expression, reading} = definition;
 
         if (mergedIndices !== null) {
-            const expressionMap = result.expressions.get(expression);
+            const expressionMap = resultExpressionsMap.get(expression);
             if (
                 typeof expressionMap !== 'undefined' &&
                 typeof expressionMap.get(reading) !== 'undefined'
@@ -200,7 +206,7 @@ function dictTermsMergeByGloss(result, definitions, appendTo=null, mergedIndices
                 reading: new Set(),
                 definitionTags: [],
                 glossary: definition.glossary,
-                source: result.source,
+                source: resultSource,
                 reasons: [],
                 score: definition.score,
                 id: definition.id,
@@ -212,8 +218,8 @@ function dictTermsMergeByGloss(result, definitions, appendTo=null, mergedIndices
         glossDefinition.expression.add(expression);
         glossDefinition.reading.add(reading);
 
-        result.expression.add(expression);
-        result.reading.add(reading);
+        resultExpressionSet.add(expression);
+        resultReadingSet.add(reading);
 
         for (const tag of definition.definitionTags) {
             if (!glossDefinition.definitionTags.find((existingTag) => existingTag.name === tag.name)) {
@@ -235,10 +241,10 @@ function dictTermsMergeByGloss(result, definitions, appendTo=null, mergedIndices
                     ...
                 ]);
             */
-            let readingMap = result.expressions.get(expression);
+            let readingMap = resultExpressionsMap.get(expression);
             if (typeof readingMap === 'undefined') {
                 readingMap = new Map();
-                result.expressions.set(expression, readingMap);
+                resultExpressionsMap.set(expression, readingMap);
             }
 
             let termTagsMap = readingMap.get(reading);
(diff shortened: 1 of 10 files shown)


ext/bg/js/dictionary.js
```diff
@@ -132,26 +132,27 @@ function dictTermsGroup(definitions, dictionaries) {
...
         results.push({
-            definitions: groupDefs,
+            definitions: groupDefinitions,
             expression: firstDef.expression,
             reading: firstDef.reading,
             furiganaSegments: firstDef.furiganaSegments,
             reasons: firstDef.reasons,
             termTags: firstDef.termTags,
-            score: groupDefs.reduce((p, v) => v.score > p ? v.score : p, Number.MIN_SAFE_INTEGER),
+            score: groupDefinitions.reduce((p, v) => v.score > p ? v.score : p, Number.MIN_SAFE_INTEGER),
```
siikamiika: I don't remember any more why this `Number.MIN_SAFE_INTEGER` is here, but you removed it from another place in an earlier commit: https://github.com/FooSoft/yomichan/commit/5541aae2012785894fbb52b2767cce5be9a6f5ba

ext/bg/js/dictionary.js
```diff
@@ -123,7 +123,7 @@ function dictTermsCompressTags(definitions) {
 }
 
 function dictTermsGroup(definitions, dictionaries) {
-    const groups = {};
+    const groups = new Map();
     for (const definition of definitions) {
         const key = [definition.source, definition.expression];
         key.push(...definition.reasons);
```
siikamiika: Also unrelated to this PR, but should this be sorted?

ext/bg/js/dictionary.js
```diff
@@ -58,10 +60,11 @@ function dictTermsSort(definitions, dictionaries=null) {
...
         let i;
         if (dictionaries !== null) {
-            i = (
-                ((dictionaries[v2.dictionary] || {}).priority || 0) -
-                ((dictionaries[v1.dictionary] || {}).priority || 0)
-            );
+            const dictionaryInfo1 = dictionaries.get(v1.dictionary);
+            const dictionaryInfo2 = dictionaries.get(v2.dictionary);
+            const priority1 = typeof dictionaryInfo1 !== 'undefined' ? dictionaryInfo1.priority || 0 : 0;
+            const priority2 = typeof dictionaryInfo2 !== 'undefined' ? dictionaryInfo2.priority || 0 : 0;
```
siikamiika: I think the `priority || 0` isn't needed here if `dictEnabledSet` is the only source of dictionaries for this function because it's already added there

ext/bg/js/translator.js
```diff
@@ -70,8 +70,8 @@ class Translator {
         return {sequencedDefinitions, defaultDefinitions};
     }
 
-    async getMergedSecondarySearchResults(text, expressionsMap, secondarySearchTitles) {
-        if (secondarySearchTitles.length === 0) {
+    async getMergedSecondarySearchResults(text, expressionsMap, secondarySearchDictionaries) {
+        if (secondarySearchDictionaries.length === 0) {
```
siikamiika: `secondarySearchDictionaries.size`

ext/bg/js/dictionary.js
```diff
@@ -251,7 +253,7 @@ function dictTermsMergeByGloss(result, definitions, appendTo=null, mergedIndices
     }
 
     return definitionsByGloss;
-}
+  }
```
siikamiika: I guess this was an accident :smile:

ext/bg/js/dictionary.js
```diff
@@ -177,11 +177,17 @@ function dictTermsMergeBySequence(definitions, mainDictionary) {
 
 function dictTermsMergeByGloss(result, definitions, appendTo=null, mergedIndices=null) {
     const definitionsByGloss = appendTo !== null ? appendTo : new Map();
+
+    const resultExpressionsMap = result.expressions;
```
siikamiika: Can't comment on the exact line, but I think the data layout comment could also be updated to `resultExpressionsMap` instead of `result.expressions`

ext/bg/js/dictionary.js
```diff
@@ -262,16 +286,15 @@ function dictTermsMergeByGloss(result, definitions, appendTo=null, mergedIndices
...
     for (const definition of definitionsByGloss.values()) {
-        definition.only = [];
-        if (!utilSetEqual(definition.expression, resultExpressionSet)) {
-            for (const expression of utilSetIntersection(definition.expression, resultExpressionSet)) {
-                definition.only.push(expression);
-            }
+        const only = [];
+        const expressionSet = definition.expression;
+        const readingSet = definition.reading;
+        definition.only = only;
```
siikamiika: Doesn't matter in practice in JS, but could be more clear if `only` was added to `definition` after it's built

ext/bg/js/dictionary.js
toasted-nutbread: It's used as the default value in case the array is empty. It was removed in the reference commit since there was guaranteed to be a value.

ext/bg/js/dictionary.js
toasted-nutbread: I'm not sure; there's a `dictTermsSort` call at the end, but I don't know if it's required or what its exact purpose is (I haven't looked into it either though).

ext/bg/js/translator.js
toasted-nutbread: I was almost positive I had missed one of these, but I didn't see it when doing a check. Thanks for noticing.

PR #363, opened 2020-02-16
Eslint stylistic rules

toasted-nutbread:
Rules for whitespaces. The added rules are designed to match Yomichan's current style. Only minor fixes had to be performed on the codebase to be conformant.

The one additional rule I was looking at adding was [`"space-infix-ops": "error"`](https://eslint.org/docs/rules/space-infix-ops) which requires spaces around inner operators:

```js
const x=0; // Invalid
const y = 1+2; // Invalid
// ...
const x = 0; // Valid
const y = 1 + 2; // Valid
```

However, this also has the effect of requiring spaces on default values for function arguments and destructuring:

```js
const {x=32} = {/* ... */};
function (arg1=null) { /* ... */ }
```

These are the only places in the codebase where operators aren't spaced, but it's ~30 locations, so I don't know if that's a rule that should be added.


ext/bg/js/search-query-parser.js
```diff
@@ -59,7 +59,7 @@ class QueryParser extends TextScanner {
 
         this.search.setContent('terms', {definitions, context: {
             focus: false,
-            disableHistory: cause === 'mouse' ? true : false,
+            disableHistory: cause === 'mouse',
```
siikamiika: I've seen `(...)` used for some places similar to this, should it be enforced too (if posssible with ESLint)?

ext/bg/js/database.js
```diff
@@ -585,7 +585,7 @@ class Database {
             tagDataLoaded(summary, bank, ++bankTotalCount, bankLoadedCount++);
         }
 
-        const loadBank = async (summary, namer, count, callback) => {
+        const loadBank = async (namer, count, callback) => {
```
siikamiika: Is there a rule that allows specifying variables used from the parent scope like the `/*global name1, name2*/`, similar to [PHP anonymous function](https://www.php.net/manual/en/functions.anonymous.php) `use`?

siikamiika:
> The one additional rule I was looking at adding was [`"space-infix-ops": "error"`](https://eslint.org/docs/rules/space-infix-ops) which requires spaces around inner operators

That's a shame, I think the current style makes default values more obvious (but some of it is probably Python bias). Found some related issues at ESLint Github and it seems like there hasn't been enough interest for adding support: https://github.com/eslint/eslint/issues/3587 https://github.com/eslint/eslint/issues/7489

ext/bg/js/search-query-parser.js
toasted-nutbread: There is [`no-extra-parens`](https://eslint.org/docs/rules/no-extra-parens), which appears to do the opposite.

The reason why I sometimes do that is because, IMO, it more clearly delimits expressions, particularly multi-part ones (e.g. `(a === 3 && b === 4)`). It also separates places where there are similar operators (e.g. `let x = (y === 32)`).

However, I'm often not consistent about this, and I don't really have a preference about using either since the benefits are _very_ minor. I do think `no-extra-parens` would probably be overkill though; minor stylistic differences are fine so long as they don't harm readability or understandability. For example, I have seen lint rules which prefer to not have parentheses anywhere, _including_ around logic expressions with mixed `&&` and `||`. Example: `x && y || z` is less clear than `(x && y) || z`.

ext/bg/js/database.js
toasted-nutbread: None that I'm aware of, and I think it would also be overkill. Closure capture is such a fundamental concept that I don't think it should be explicitly stated; this must be done in C++ and is honestly quite tiresome. IDEs will often also be able to highlight uses of a variable, so this can also provide clarity about variable use.

ext/bg/js/search-query-parser.js
siikamiika: Yeah, it can be very contextual when it makes sense to add the extra parentheses, not something that a linter rule will understand. There are also code formatters like [Black](https://github.com/psf/black) (Python) that automate the part of fixing lint errors, but having tried that there are often cases when an alternative way of formatting would make it more clear for future readers. The benefit is that you don't have to spend time thinking of formatting.

ext/bg/js/database.js
siikamiika: I also thought the same when I started doing PHP last year, but now I'm used to it. Most programming languages seem to capture automatically, even Rust.

Issue #29, opened 2017-01-02
[feature request] iframe support


jpc-ae:
There are several sites, like google translator toolkit, where different parts of the page are broken up into iframes. It would be nice to be able to check text in all the frames on sites that still use them.

FooSoft:
Could you link me to a specific page(s) that does not work?

jpc-ae:
Well, it might be a niche use-case, but if you go to https://translate.google.com/toolkit/ and upload some Japanese text to work on, then open it, the left pane in the workbench (the opened file) isn't being parsed

siikamiika:
Twitter widget: http://dic.nicovideo.jp/

FooSoft:
So this is still appears to be a problem for the twitter widget, but the latest release solves IFRAME issues in general. I think the problem is that the Twitter IFRAME is procedurally generated so that the content script is not injected into it? Need to investigate this further.

seanblue:
I'm still seeing some iframe issues. Take this [Amazon page](https://www.amazon.co.jp/%E6%96%B0%E4%B8%96%E7%95%8C%E3%82%88%E3%82%8A-%E4%B8%8A-%E8%AC%9B%E8%AB%87%E7%A4%BE%E6%96%87%E5%BA%AB-%E8%B2%B4%E5%BF%97%E7%A5%90%E4%BB%8B-ebook/dp/B009GXME5G/). Scanning the text in the description (starting with １０００年後の日本。) don't show the popup.

I've seen the popup not work on other sites, but I'm not sure if they were iframe issues and I don't remember the sites off hand.

siikamiika:
@seanblue For me the popup shows up, but it's almost unusable because the iframe is so small. It would be possible to open a popup on the root page with the framework that was added in recursive popups (https://github.com/FooSoft/yomichan/pull/185), but there could be some problems positioning the popup according to https://github.com/FooSoft/yomichan/pull/185#issuecomment-522788721.

toasted-nutbread:
I have looked into supporting iframes after the recursive popups were added, and the only issue that really prevents it is that there is no way to determine which ```<iframe>``` corresponds to which ```frameId```. This makes it very difficult to determine what the correct offset for the popup should be on the root page. I suppose it would be possible to use "Full width" display mode to avoid this issue, but I think it's worth trying to exhaust any other potential methods of determining popup position before it comes to this.

M-Azuma:
I seem to have the same problem.
[...]

PR #364, opened 2020-02-16
popup proxy for iframe

siikamiika:
My plan is to empirically detect the position of iframes on the page to position the popup. It probably won't be perfect, but a lot better than the current iframe support. If I can't get it to work, I'll make it fallback to full width popups.

Opening the PR to get comments on the general structure of this solution, but don't spend too much time reviewing the code yet.

### DIFF
--- a/ext/fg/js/frontend-initialize.js
+++ b/ext/fg/js/frontend-initialize.js
@@ -16,19 +16,25 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
 
-/*global PopupProxyHost, PopupProxy, Frontend*/
+/*global PopupProxyHost, PopupProxy, PopupIframe, Frontend, apiFrameInformationGet*/
 
 async function main() {
     const data = window.frontendInitializationData || {};
     const {id, depth=0, parentFrameId, ignoreNodes, url, proxy=false} = data;
+    const isPageIframe = !proxy && (window !== window.parent);
 
-    let popupHost = null;
-    if (!proxy) {
-        popupHost = new PopupProxyHost();
+    let popup;
+    if (isPageIframe) {
+        const {frameId} = await apiFrameInformationGet();
+        popup = new PopupIframe(frameId);
+    } else if (proxy) {
+        popup = new PopupProxy(depth + 1, id, parentFrameId, url);
+    } else {
+        const popupHost = new PopupProxyHost();
         await popupHost.prepare();
+        popup = popupHost.createPopup(null, depth);
     }
 
-    const popup = proxy ? new PopupProxy(depth + 1, id, parentFrameId, url) : popupHost.createPopup(null, depth);
     const frontend = new Frontend(popup, ignoreNodes);
     await frontend.prepare();
 }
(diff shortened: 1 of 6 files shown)


toasted-nutbread:
For reference, here is some code that I worked on:

https://github.com/toasted-nutbread/yomichan/tree/iframe-popups

Also, see this specific comment: https://github.com/FooSoft/yomichan/issues/29#issuecomment-568207279, which is a race issue with frame load order. I primarily notice it in Chrome.

siikamiika:
Thanks, seems like the general direction is about the same, but you reused `PopupProxy`. I'll see if I can do something about the race issue. Do you know if it's related to why on some pages the iframes don't work at all?

toasted-nutbread:
I have personally not seen those cases, so I'm not sure. Nothing immediately strikes me as to why that would currently happen. The reason the race condition is an issue for the way I set up the iframe branch is because the iframe tries to communicate with the root frame (id 0), but it sends the message before the root frame has set up the listener.

siikamiika:
> I have personally not seen those cases, so I'm not sure.

Try on [this Amazon page](https://www.amazon.co.jp/%E6%96%B0%E4%B8%96%E7%95%8C%E3%82%88%E3%82%8A-%E4%B8%8A-%E8%AC%9B%E8%AB%87%E7%A4%BE%E6%96%87%E5%BA%AB-%E8%B2%B4%E5%BF%97%E7%A5%90%E4%BB%8B-ebook/dp/B009GXME5G/) from #29 (with a version before #357). It works on Firefox for me, but not on Chrome. The twitter widget [here](https://dic.nicovideo.jp/) fails on both browsers.

toasted-nutbread:
The twitter one probably doesn't work because there are some about:blank shenanigans going on, and Yomichan doesn't run on about:blank. We will have to add [`"match_about_blank": true`](https://developer.chrome.com/extensions/content_scripts) to support this.

siikamiika:
> The twitter one probably doesn't work because there are some about:blank shenanigans going on, and Yomichan doesn't run on about:blank. We will have to add [`"match_about_blank": true`](https://developer.chrome.com/extensions/content_scripts) to support this.

Wow, nice find! This actually fixes both of the sites on Chrome and Firefox.

I've added a guessing framework which seems to work fine as long as there aren't nested iframes. I'll probably make it fall back to full width popup.

siikamiika:
This changes current behavior: https://github.com/FooSoft/yomichan/pull/364/commits/d6ba291d0ce341fa570b73499ea90603fc449906
The popup size won't be kept after scanning again.

siikamiika:
@toasted-nutbread Are there any considerations such as permissions for adding `match_about_blank`? You can push it to master if you want.

toasted-nutbread:
> @toasted-nutbread Are there any considerations such as permissions for adding `match_about_blank`? You can push it to master if you want.

I'll check and see if it reports any permissions differences on install, but I don't think it does. Other than that, I don't think there should be any issues. I'll do a little bit of testing to confirm and add it in, I'm working on something in that area anyway.

toasted-nutbread:
Didn't notice any differences in the permissions pages or popups during installation. f0bd61617785e503ffd0649ae60502469eb75876

PR #366, opened 2020-02-16
Popup init fix

### LINKED ISSUE #360: setDisplayInitialized failing due to iframes
Getting this error when iframes are on the page:

```
Error in event handler: Error: Override me
    at Popup.setDisplayInitialized (/fg/js/popup.js:128:15)
    at /fg/js/frontend.js:248:59
    at Frontend.onRuntimeMessage (/fg/js/frontend.js:85:24)
```

toasted-nutbread:
Fixes #360

### DIFF
--- a/ext/fg/js/float.js
+++ b/ext/fg/js/float.js
@@ -29,12 +29,16 @@ class DisplayFloat extends Display {
         };
 
         this._orphaned = false;
+        this._prepareInvoked = false;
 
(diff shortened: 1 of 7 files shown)


siikamiika:
Thanks for the fix, I couldn't figure out how to make it work again without revert and that's why I started doing an iframe popup proxy instead. I'll continue to work on #364 as a new feature and not a bugfix after rebasing on this.

PR #367, opened 2020-02-16
Defer content script css injection

toasted-nutbread:
Creating this as a PR now to allow some testing to see if this works. I have noticed that this causes about:blank pages to not work, so it's not ready to be merged unless that can be fixed.


toasted-nutbread:
6a5a0d3d3095b32d00ea716377ec9e540d4f06e6 was the issue

PR #369, opened 2020-02-18
Dictionary validation

toasted-nutbread:
Validates dictionaries using schemas before importing the data.

Will be used for #322, #61.


siikamiika:
Tested this a bit with valid and intentionally broken dictionaries. It seemed to work good. I think it would be more helpful if the error messages were more verbose than `Invalid enum value` and include the attempted value, and also tell the json file that triggered the error. But it does its job.

toasted-nutbread:
> I think it would be more helpful if the error messages were more verbose than `Invalid enum value` and include the attempted value, and also tell the json file that triggered the error.

Updated the error messages to have more info. Examples:

```
Dictionary has invalid data in 'index.json' for value 'dictionary.format', validated against 'schema.properties.format': Invalid enum value
Dictionary has invalid data in 'kanji_bank_1.json' for value 'dictionary[0]', validated against 'schema.additionalItems': Array length too short
Dictionary has invalid data in 'kanji_meta_bank_1.json' for value 'dictionary', validated against 'schema': Value type object does not match schema type array
Dictionary has invalid data in 'tag_bank_1.json' for value 'dictionary[0][0]', validated against 'schema.additionalItems.items[0]': Value type object does not match schema type string
Dictionary has invalid data in 'term_bank_1.json' for value 'dictionary[0][5]', validated against 'schema.additionalItems.items[5]': Value type boolean does not match schema type array        
Dictionary has invalid data in 'term_meta_bank_1.json' for value 'dictionary', validated against 'schema': Value type boolean does not match schema type array
```

Also added some tests to test some basic invalid dictionaries.

PR #371, opened 2020-02-19
Database tests

toasted-nutbread:
Some basic unit tests for database.js. Intended to be merged before #369, and then #369 should subsequently be updated to ensure it is conformant.


siikamiika:
Nice, thanks for adding the test data :+1:

toasted-nutbread:
One thing to note: fake-indexeddb doesn't work with a huge amount of data, so something big like the innocent_corpus dictionary isn't able to be imported for testing

PR #375, opened 2020-02-22
Document tests

toasted-nutbread:
Tests for some of the functions in [document.js](https://github.com/FooSoft/yomichan/blob/master/ext/fg/js/document.js). This demonstrates that it can be possible to test some of the DOM-based functionality without requiring a web browser. This will be useful for #286 and future related changes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,16 +24,46 @@
                 "js-tokens": "^4.0.0"
             }
         },
+        "abab": {
+            "version": "2.0.3",
+            "resolved": "https://registry.npmjs.org/abab/-/abab-2.0.3.tgz",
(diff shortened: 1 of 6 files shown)


siikamiika:
This seems very useful, thanks! I tested adding some basic errors to `source.js` and the tests caught them. My only concern is that the tests might need large changes when an overhaul like https://github.com/FooSoft/yomichan/pull/286 is done to scanning, but it seems worth the trouble because scanning is a core feature.

toasted-nutbread:
That is expected; I will definitely have to update the tests when #286 and similar are integrated. The main intent is so that the results aren't inadvertently modified when changing something which shouldn't affect it.

Issue #378, opened 2020-02-26
Export from a single dictionary


psychodelicwasabi:
Would there be a way to generate the Anki cards with the definitions of a single dictionary only? I have multiple dictionaries, but don't need all those definitions in the card. So, in the end, I have to erase the other definitions when making cards using Yomichan, which becomes time consuming. Is there a way to play with handlebars or something to make that happen? Also, please, when answering, make it extremely simple, I really can't code at all.

Ideally, I'd also like to get rid of the reading and etc that comes with the definition, but I'm not sure that can be done. Here's an image with all the information I don't need and end up cleaning in the card-making process underlined. Thank you for the attention.
![image](https://user-images.githubusercontent.com/39304471/75307483-32a49280-5822-11ea-9c39-98eaaab0eee0.png)

toasted-nutbread:
Answer to first half: this is not possible in the current version of Yomichan, but this ability to could be added. Taking your idea of using the Handlebars templates, this would require the ability to include definitions based on a condition, in this case the condition would be checking to see if the name of the dictionary is `"三省堂　スーパー大辞林"` for the definitions. Unfortunately, it seems that Handlebars doesn't have built-in functionality for this comparison, so this would have to be added.

Answer to second half: this would likely be difficult to consistently do in a handlebars template since there doesn't seem to be a good pattern for what you want to remove. Consider another definition for `"口"`:
![image](https://user-images.githubusercontent.com/11037431/75498091-3f48f800-5994-11ea-9b6d-a29643573037.png)
The pattern in this case is different.

This is honestly more of a problem with the dictionary data than with what Yomichan's doing with it; I think a few of the Japanese=>Japanese dictionaries have similar formatting issues.

Geniusssmit:
I'm dealing with the same issue, is there any way to only export from 1 dictionary now?

toasted-nutbread:
You are now able to do these comparison! The trick is finding the right place to put them, and what field to compare against, but here is an example:

```handlebars
{{#*inline "test"}}
    {{~#if (op "===" definition.dictionary "JMdict (English)")~}}
        Matches
    {{~#else~}}
        Doesn't match
    {{~/if~}}
{{/inline}}
```

Note: the example above is not using per-definition dictionary checks, which may be necessary depending on which presentation mode is being used. You would likely need to modify the `glossary` inline definition and add these checks to it.

There is also some basic documentation for the Handlebars helpers now:
[...]

Issue #190, opened 2019-08-27
No "Add to Anki" button on Firefox


jacobalbano:
The button to add terms to Anki has suddenly stopped appearing for me. I've tried restarting Firefox, Anki, Yomichan, and even my computer but nothing has caused it to come back.

![image](https://user-images.githubusercontent.com/3813170/63738256-321d4600-c857-11e9-8883-d4655ebec9ca.png)

When I navigate to http://localhost:8765/. I do see the `AnkiConnect v.6` message. Creating a new deck in Anki and then opening Yomichan settings shows the deck in the drop-down:
![image](https://user-images.githubusercontent.com/3813170/63738377-8e806580-c857-11e9-9348-b912d0c50919.png)

I've seen this issue come up for a few people in the Japanese-learning community I'm part of, but I've never had a problem before. I'd like to know if there's any way I can troubleshoot this.

siikamiika:
@jacobalbano Have you edited your Anki templates in the past? The error you're experiencing is probably caused the new feature letting you embed screenshots of the page to your flashcards, because edited templates don't get updated with the new default. If your template is missing screenshot, it will not work (this could probably be fixed somehow but I haven't investigated yet).

To fix the issue, you can either reset your templates to default or add the following there:
```
{{#*inline "screenshot"}}
    <img src="{{definition.screenshotFileName}}" />
{{/inline}}
```

jacobalbano:
@siikamiika That fixed it, thanks. I doubt I'm the only person to have edited the templates though; any chance of a fix that doesn't silently fail when a field is missing? I don't plan to use the screenshot feature and it seems that the template should only be *required* to include fields that you want to capture.

FooSoft:
The settings could be versioned, fixing up this issue in the process... however I imagine that at this point it might be moot since the current vetsion has been out for a little while already.

PR #380, opened 2020-02-28
Load default Anki field templates from a file

toasted-nutbread:
While looking into #378, I realized that it's somewhat annoying to look at the default field templates with syntax highlighting. To do this, the defaults must be copied either from options.js or from the settings page, both of which is an extra step. Moving the handlebars templates into a separate file allows them to be viewed in isolation more easily, and also follows the pattern used to have JSON data in separate files.

This currently does affect the settings upgrade process for _very_ old settings, from around version 1.5.2 (2017-11-08). However, IMO the conditions to cause this should be sufficiently rare that it's better to just omit the additional code/complexity it would take to maintain the same upgrade: user is upgrading from 1.5.2 to 20.2.24.0+ (rare) and has modified their Anki templates (rare). And the effect is just that the Anki templates are reverted to the default value.

Affected upgrade:
https://github.com/FooSoft/yomichan/commit/8e29da0c6bd0b80dc6c9e37a525a37258518c293#diff-4c659e681ee5d121dda04689ecc50d35L60-R63

Upgrade originally introduced in:
https://github.com/FooSoft/yomichan/commit/5f6830c7dd88520ebf41c24ab51e2c3240039939#diff-4c659e681ee5d121dda04689ecc50d35R268-R276

### DIFF
--- a/ext/bg/data/default-anki-field-templates.handlebars
+++ b/ext/bg/data/default-anki-field-templates.handlebars
@@ -1,0 +1,161 @@
+{{#*inline "glossary-single"}}
+    {{~#unless brief~}}
+        {{~#if definitionTags~}}<i>({{#each definitionTags}}{{name}}{{#unless @last}}, {{/unless}}{{/each}})</i> {{/if~}}
+        {{~#if only~}}({{#each only}}{{{.}}}{{#unless @last}}, {{/unless}}{{/each}} only) {{/if~}}
+    {{~/unless~}}
+    {{~#if glossary.[1]~}}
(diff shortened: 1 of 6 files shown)


siikamiika:
I agree that it's become very rare that there would be someone with such an old version. One reason why an upgrade could be postponed is a permission request, but two years should be enough time.

It could still make sense to have a method for conditionally upgrading templates without having templates as JS strings at some point, such as when a new partial is added. I remember there were some problems when screenshot support was added https://github.com/FooSoft/yomichan/issues/190.

Issue #379, opened 2020-02-27
[Feature Request] Hotkey to skip to first entry in second dictionary


breead:
I use several dictionaries, with 新明解 being the primary one. However there are lots of words for which I also want to check JMdict (due to entries not in other dictionaries, commonly slang/expressions). If there are a lot of cards (I have result grouping set to split because it helps with making anki cards), scrolling and finding exactly the first JMdict entry can take a little while, and so I think a hotkey to scroll there automatically would be really useful. 
![image](https://user-images.githubusercontent.com/10461870/75476649-4293b200-599b-11ea-830b-c902cf20c09a.png)
If anybody could implement this feature I would be very grateful, it'd be a massive QoL improvement for quickly using several dictionaries. (Never written a github issue before, sorry if this is missing something)

Issue #368, opened 2020-02-17
Query tab Frontend instance structure from Backend for communication


siikamiika:
There are many ways to communicate between windows: you can `postMessage` to a window, `chrome.runtime.sendMessage/onMessage`, connect to port with `chrome.runtime.connect`, maybe I forgot something else.

They all work a bit differently, but especially window communication can be very confusing when used with multiple frames or nested frames. I've seen `apiForward` used instead probably because it only sends the message to the extension in the current tab, but that doesn't solve the issue of sending to the correct frame of the extension.

After seeing where https://github.com/FooSoft/yomichan/issues/353 is going, I'm thinking if it would be helpful to make `Backend` manage all the different instances of `Frontend` and their types (root page instance, nested proxies, iframe proxy https://github.com/FooSoft/yomichan/pull/364) along with a unique ID so that you could send messages to specific `Frontend` instances. For one, it would solve the issue I mentioned in https://github.com/FooSoft/yomichan/issues/353#issuecomment-587116442. I think it would also make debugging easier if you had to target a specific ID with every call.

Example structure (123 is tab ID but you could query the current tab):
```json
{
  "123": [
    {"id": "581ce391b4c19ef5e2adc8745e343820", "type": "root", "parent": null},
    {"id": "555c5664b055b58d551b4e04d6847454", "type": "iframe", "parent": null},
    {"id": "cf3187af33d0055b6355b811fedc73e9", "type": "nested", "parent": "581ce391b4c19ef5e2adc8745e343820"},
    {"id": "569dfb2c6043c90f12b04487ba14f830", "type": "nested", "parent": "cf3187af33d0055b6355b811fedc73e9"}
  ]
}
```

toasted-nutbread:
I have thought of this a few times, and the main issue that comes up is the added complexity and permissions requirements. The background now has to know about all tabs and frames and to assign IDs. In order to not have memory leaks, the backend also needs to observe for when frames are removed in order to clean up data. This requires the [webNavigation](https://developer.chrome.com/extensions/webNavigation) permission, which is otherwise not needed (so far).

siikamiika:
I agree it would make things more complicated. Would be nicer if this could be handled within a tab, but I remember seeing in https://github.com/FooSoft/yomichan/issues/29#issuecomment-568207279 that frames can load before the root page, so either the frames in one tab have to share their state in a peer-to-peer way or there has to be a master that's up before the frames.

[...]

PR #381, opened 2020-02-29
get all matching profiles

siikamiika:
This doesn't change anything yet, but allows getting a list of matching profiles that can be used to cycle between profiles with a hotkey or a button later.

### DIFF
--- a/ext/bg/js/backend.js
+++ b/ext/bg/js/backend.js
@@ -212,34 +212,44 @@ class Backend {
         }
     }
 
-    async getOptions(optionsContext) {
+    async getAllMatchingOptions(optionsContext) {
         if (this.isPreparedPromise !== null) {
             await this.isPreparedPromise;
         }
-        return this.getOptionsSync(optionsContext);
+        return this.getAllMatchingOptionsSync(optionsContext);
+    }
+
+    async getOptions(optionsContext) {
+        return (await this.getAllMatchingOptions(optionsContext))[0];
+    }
+
+    getAllMatchingOptionsSync(optionsContext) {
+        return this.getAllMatchingProfilesSync(optionsContext).map((p) => p.options);
     }
 
     getOptionsSync(optionsContext) {
-        return this.getProfileSync(optionsContext).options;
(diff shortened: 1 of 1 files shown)


toasted-nutbread:
Should any of these functions be marked as private?

siikamiika:
I'm not sure about all of them yet, and maybe even the async version could be dropped (https://github.com/FooSoft/yomichan/issues/356). At least the generator is only used privately.

This is groundwork for a solution to issues like https://github.com/FooSoft/yomichan/issues/378 and https://github.com/FooSoft/yomichan/issues/379, but it's not the only possible way to go. Another one would be to add a global setting (such as a string) that can be used as a profile condition and switched from `browser_action`, hotkeys or a button inside the popup.

The advantage with this approach is that you can always have profiles with duplicate conditions and the one sorted first (as far as I understand profiles) will always be used, but now you could see all of them. The disadvantage is that the profile switch is not persistent, but I'd imagine that an easy workaround would be rearranging the profiles.

Yet another issue to solve before live profile switching can be implemented is properly applying all settings to the popup when the profile changes. Full refresh could be one solution.

toasted-nutbread:
> I'm not sure about all of them yet, and maybe even the async version could be dropped (#356).

This might be doable; I would be sure to test if there are any obvious errors that occur during startup for example (reload the extension). I know some can already occur if content pages are trying to use the background APIs but the background isn't fully initialized yet, but that's more of a separate issue.

> At least the generator is only used privately.

Just mark anything that isn't intended to be used externally with `_`.

siikamiika:
Closing this PR because I'm working on another branch that includes something similar. I stopped working on it for now because I realized that I need reliable communication from `DisplayFloat` to `Frontend` to be able to set the selected profile in both. I'll explore the peer-to-peer option with logic similar to what's done with the `Yomichan` class in https://github.com/FooSoft/yomichan/pull/386 instead of storing everything in Backend. https://github.com/FooSoft/yomichan/issues/368

https://github.com/siikamiika/yomichan/tree/switch-matching-profiles

Issue #359, opened 2020-02-15
Preferred callback style


toasted-nutbread:
## Background
I am looking to change most of the static class callback maps into instance member callback maps for style consistency and to allow static analysis and automatic refactoring tools to have stronger type inferencing information.

For example, I am looking to convert these:
https://github.com/FooSoft/yomichan/blob/89cb98e4d14ad4980133f8e8e44b15d56c3ebea5/ext/bg/js/backend.js#L843-L868

Into something that looks more like this (without the `FrontendApiReceiver` constructor of course, just the `Map`):
https://github.com/FooSoft/yomichan/blob/89cb98e4d14ad4980133f8e8e44b15d56c3ebea5/ext/fg/js/popup-proxy-host.js#L36-L48

The intent is to replace `self` with `this`. `self` cannot be type-inferenced whereas `this` can be. In general, the overhead of having a per-instance `Map` rather than a global one should be very small since most of the classes that are using this pattern are instantiated only once or a very small number of times. I was also the one to initially change to the static `Map` pattern, but after using it for a while, I've found that having the `Map` as an instance member is probably better.

---

## Callback Style
That change will be simple enough to make, but as I was looking into it, I noticed something: there are two styles for callback functions, and I am wondering if we should pick one or the other for consistency and if it's worth it to do so.

The two styles are as follows:

### Type 1
`something.addCallback(this.someFunction.bind(this));`
https://github.com/FooSoft/yomichan/blob/89cb98e4d14ad4980133f8e8e44b15d56c3ebea5/ext/bg/js/backend.js#L71
https://github.com/FooSoft/yomichan/blob/89cb98e4d14ad4980133f8e8e44b15d56c3ebea5/ext/fg/js/frontend.js#L54-L55

### Type 2
`something.addCallback(() => this.someFunction());`
https://github.com/FooSoft/yomichan/blob/89cb98e4d14ad4980133f8e8e44b15d56c3ebea5/ext/bg/js/backend.js#L66
https://github.com/FooSoft/yomichan/blob/89cb98e4d14ad4980133f8e8e44b15d56c3ebea5/ext/fg/js/popup.js#L340
https://github.com/FooSoft/yomichan/blob/89cb98e4d14ad4980133f8e8e44b15d56c3ebea5/ext/fg/js/frontend.js#L58-L60

### Comparison
[...]

PR #385, opened 2020-03-01
Pitch accents

toasted-nutbread:
Adds support for #61.

The dictionary generation script can be found here:
https://github.com/toasted-nutbread/yomichan-pitch-accent-dictionary

A build of the dictionary can be found here:
https://github.com/toasted-nutbread/yomichan-pitch-accent-dictionary/releases/tag/1.0.0

There are three different modes for displaying pitch accent data, each of which can be enabled/disabled individually.
* **Show downstep notation for pitch accents**
  Shows a bar over characters with a higher pitch and, if present, a small down bar where the pitch drop is.
* **Show position notation for pitch accents**
  Shows the numeric indication for the accent position. E.g. [1]
* **Show graph for pitch accents**
  Shows a graph of points indicating the accent positions of the mora. Currently disabled by default.

Here are some good terms for testing:
| Term | Reason |
| ----- |:--------- |
| 端 | Single kanji term with three readings with varying accent positions. |
| に | Single term with many sequenced terms in merged mode. |
| ２つ | Term which has multiple pitch accent positions with tags. |
| 博士 | Term with two kanji which has multiple readings with the same accent position. |
| し | Term with single pitch accent definition and no sequenced terms. |
| しっ | Term with no pitch accent definition. |

Example image showing all 3 modes:
image

### DIFF
--- a/ext/mixed/js/display-generator.js
+++ b/ext/mixed/js/display-generator.js
@@ -294,7 +294,7 @@ class DisplayGenerator {
         node.dataset.pitchesMulti = 'true';
         node.dataset.pitchesCount = `${dictionaryPitches.length}`;
 
-        const tag = this.createTag({notes: '', name: dictionary, category: 'dictionary'});
+        const tag = this.createTag({notes: '', name: dictionary, category: 'pitch-accent-dictionary'});
         node.querySelector('.term-pitch-accent-group-tag-list').appendChild(tag);
 
         const n = node.querySelector('.term-pitch-accent-list');
--- a/ext/mixed/js/display-generator.js
+++ b/ext/mixed/js/display-generator.js
@@ -301,22 +301,28 @@ class DisplayGenerator {
         }
     }
 
-    static _appendMultiple(container, createItem, detailsArray, fallback=[]) {
+    static _appendMultiple(container, createItem, detailsIterable, fallback=[]) {
         if (container === null) { return 0; }
 
-        const isArray = Array.isArray(detailsArray);
-        if (!isArray) { detailsArray = fallback; }
+        const multi = (
+            detailsIterable !== null &&
+            typeof detailsIterable === 'object' &&
+            typeof detailsIterable[Symbol.iterator] !== 'undefined'
+        );
+        if (!multi) { detailsIterable = fallback; }
 
-        container.dataset.multi = `${isArray}`;
-        container.dataset.count = `${detailsArray.length}`;
-
-        for (const details of detailsArray) {
+        let count = 0;
+        for (const details of detailsIterable) {
             const item = createItem(details);
             if (item === null) { continue; }
             container.appendChild(item);
+            ++count;
         }
 
-        return detailsArray.length;
+        container.dataset.multi = `${multi}`;
+        container.dataset.count = `${count}`;
+
+        return count;
     }
 
     static _appendFurigana(container, segments, addText) {
--- a/ext/mixed/js/display-generator.js
+++ b/ext/mixed/js/display-generator.js
@@ -43,11 +43,17 @@ class DisplayGenerator {
 
         const expressionMulti = Array.isArray(details.expressions);
         const definitionMulti = Array.isArray(details.definitions);
+        const expressionCount = expressionMulti ? details.expressions.length : 1;
+        const definitionCount = definitionMulti ? details.definitions.length : 1;
 
         node.dataset.expressionMulti = `${expressionMulti}`;
         node.dataset.definitionMulti = `${definitionMulti}`;
-        node.dataset.expressionCount = `${expressionMulti ? details.expressions.length : 1}`;
-        node.dataset.definitionCount = `${definitionMulti ? details.definitions.length : 1}`;
+        node.dataset.expressionCount = `${expressionCount}`;
+        node.dataset.definitionCount = `${definitionCount}`;
+
+        bodyContainer.dataset.sectionCount = `${
+            (definitionCount > 0 ? 1 : 0)
+        }`;
 
         const termTags = details.termTags;
         let expressions = details.expressions;
--- a/ext/mixed/js/display-generator.js
+++ b/ext/mixed/js/display-generator.js
@@ -304,7 +304,7 @@ class DisplayGenerator {
     }
 
     createPitch(details) {
-        const {expressions, reading, position, tags} = details;
+        const {exclusiveExpressions, reading, position, tags} = details;
         const morae = jp.getKanaMorae(reading);
 
         const node = this._templateHandler.instantiate('term-pitch-accent');
@@ -319,7 +319,7 @@ class DisplayGenerator {
         DisplayGenerator._appendMultiple(n, this.createTag.bind(this), tags);
 
         n = node.querySelector('.term-pitch-accent-expression-list');
-        DisplayGenerator._appendMultiple(n, this.createPitchExpression.bind(this), expressions);
+        DisplayGenerator._appendMultiple(n, this.createPitchExpression.bind(this), exclusiveExpressions);
 
         n = node.querySelector('.term-pitch-accent-characters');
         for (let i = 0, ii = morae.length; i < ii; ++i) {
@@ -481,6 +481,7 @@ class DisplayGenerator {
     static _getPitchInfos(definition) {
         const results = new Map();
 
+        const allExpressions = new Set();
         const expressions = definition.expressions;
         const sources = Array.isArray(expressions) ? expressions : [definition];
         for (const {pitches: expressionPitches, expression} of sources) {
@@ -498,7 +499,19 @@ class DisplayGenerator {
                         dictionaryResults.push(pitchInfo);
                     }
                     pitchInfo.expressions.add(expression);
+                    allExpressions.add(expression);
                 }
+            }
+        }
+
+        for (const dictionaryResults of results.values()) {
+            for (const result of dictionaryResults) {
+                const exclusiveExpressions = [];
+                const resultExpressions = result.expressions;
+                if (!areSetsEqual(resultExpressions, allExpressions)) {
+                    exclusiveExpressions.push(...getSetIntersection(resultExpressions, allExpressions));
+                }
+                result.exclusiveExpressions = exclusiveExpressions;
             }
         }
 
--- a/ext/mixed/js/display-generator.js
+++ b/ext/mixed/js/display-generator.js
@@ -25,6 +25,7 @@
 class DisplayGenerator {
     constructor() {
         this._templateHandler = null;
+        this._termPitchAccentStaticTemplateIsSetup = false;
     }
 
     async prepare() {
@@ -337,6 +338,10 @@ class DisplayGenerator {
             n.appendChild(n1);
         }
 
+        if (morae.length > 0) {
+            this.populatePitchGraph(node.querySelector('.term-pitch-accent-graph'), position, morae);
+        }
+
         return node;
     }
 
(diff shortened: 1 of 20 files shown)


siikamiika:
Looks very nice, good job! I'll test this more closely some day next week.

ext/mixed/js/display-generator.js
```diff
@@ -258,6 +276,117 @@ class DisplayGenerator {
...
+        node.dataset.tagCount = `${tags.length}`;
+
+        let n = node.querySelector('.term-pitch-accent-position');
+        n.textContent = `${position}`;
+
+        n = node.querySelector('.term-pitch-accent-tag-list');
+        DisplayGenerator._appendMultiple(n, this.createTag.bind(this), tags);
+
+        n = node.querySelector('.term-pitch-accent-expression-list');
+        DisplayGenerator._appendMultiple(n, this.createPitchExpression.bind(this), expressions);
```
siikamiika: Instead of `expressions`, I think this could have something similar to `only` here:
https://github.com/FooSoft/yomichan/blob/46fee07d36b3966af9bacca8c6253b044bde07ee/ext/bg/js/dictionary.js#L282-L293
Now there are issues when either multiple expressions have the same pitch accent:

Or only graphs are enabled:

siikamiika:
Just a minor cosmetic/usability issue, but I think the titles can take quite a lot of vertical space with the default sized popup. Maybe an option to disable them would be useful.

siikamiika:
The pitch accents don't respect the "compact glossaries" setting. I think they look good as is and reusing the setting could be undesirable, so my solution would be to either leave it or add a separate setting. Minor issue.

ext/mixed/js/display-generator.js
```diff
@@ -258,6 +276,117 @@ class DisplayGenerator {
         return node;
     }
 
+    createPitches(details) {
+        if (!this._termPitchAccentStaticTemplateIsSetup) {
+            this._termPitchAccentStaticTemplateIsSetup = true;
+            const t = this._templateHandler.instantiate('term-pitch-accent-static');
+            document.head.appendChild(t);
+        }
```
siikamiika: Is this worth the optimization or are there other concerns? Maybe it could be added in `prepare`.

ext/mixed/js/display-generator.js
```diff
@@ -346,4 +481,85 @@ class DisplayGenerator {
...
+    static _jpCreateSmallKanaSet() {
+        return new Set(Array.from('ぁぃぅぇぉゃゅょゎァィゥェォャュョヮ'));
+    }
+
+    static _jpIsMoraPitchHigh(moraIndex, pitchAccentPosition) {
+        return pitchAccentPosition === 0 ? (moraIndex > 0) : (moraIndex < pitchAccentPosition);
+    }
 }
+
+DisplayGenerator._smallKanaSet = DisplayGenerator._jpCreateSmallKanaSet();
```
siikamiika: There are now even more Japanese functions in `DisplayGenerator`. I remember that we let `_isCharacterKanji` slide because it was so simple, but this is more complex. Might be worth it to include `japanese.js` as a dependency now.

Also, does adding `_smallKanaSet` to the class have the same issues as callbacks (https://github.com/FooSoft/yomichan/issues/359)?

seanblue:
I currently see all the readings for 端 as separate entries. Does this PR change that view to combine them as shown in your pictures or is that a setting? If a setting, how will this look with separate entries?

toasted-nutbread:
> 
> 
> I currently see all the readings for 端 as separate entries. Does this PR change that view to combine them as shown in your pictures or is that a setting? If a setting, how will this look with separate entries?

It's the "Result grouping" setting when set to "Group results by main dictionary entry".

Looks like this in the other modes:
image

ext/mixed/js/display-generator.js
toasted-nutbread: In cases where pitch accent dictionaries aren't installed, it won't be added. I suppose it's not a big deal either way if it's unused, so it could go into prepare.

ext/mixed/js/display-generator.js
toasted-nutbread: > There are now even more Japanese functions in `DisplayGenerator`. I remember that we let `_isCharacterKanji` slide because it was so simple, but this is more complex. Might be worth it to include `japanese.js` as a dependency now.

There should probably be a new japanese.js for the frontend now, yeah.

> Also, does adding `_smallKanaSet` to the class have the same issues as callbacks (#359)?

Not really an issue per se, but it could be updated to use the same convention. This change was initially made before that, so that's partially why this style was used.

ext/mixed/js/display-generator.js
toasted-nutbread: > Instead of `expressions`, I think this could have something similar to `only` 

That's effectively already what it's doing

> [img1]

Is the issue here that `背、脊：` is displayed twice?

> [img2]

Is the issue here that there is no reading displayed?

toasted-nutbread:
> Just a minor cosmetic/usability issue, but I think the titles can take quite a lot of vertical space with the default sized popup. Maybe an option to disable them would be useful.

The reason I added them headings is so that the pitch accents and definitions don't blend together. Do you (or anyone else following) think it would be unclear if the headings weren't there? They were also added since there were some spacing issues with the overlines for the downstep notation, but that can probably be adjusted without headings.

ext/mixed/js/display-generator.js
siikamiika: The problem is that it only works when all expressions differ in both reading and kanji. In the 背 example the pitch expression pointers are not needed because both pitch accents apply to both results, and in the 端 example the logic thinks that everything is fine although there are three different readings.

ext/mixed/js/display-generator.js
siikamiika: > There should probably be a new japanese.js for the frontend now, yeah.

That's cool too if there's not much overlap.

siikamiika:
> Do you (or anyone else following) think it would be unclear if the headings weren't there?

I think the dictionary name tag gives some hint already, but I like how it currently looks on the search page where there's enough vertical space. I tested with
```css
.term-entry-body-section-header {
  display: none;
}
```
and it still seems quite clear.

seanblue:
> Do you (or anyone else following) think it would be unclear if the headings weren't there?

I don't think the headings are necessary. You have to add a pitch accent dictionary explicitly, so at that point I think you expect the extra information to show up. And I agree that the tags should make it clear if for some reason there was confusion.

siikamiika:
There's a potential gotcha in just removing the headings when you have multiple pitch accent dictionaries, pitch accents and definitions use a separate ordered list and it could look weird:

One solution would be to make pitch accents a subset of regular definitions, but that would be a major restructuring in the code and then you would have other problems such as sorting pitch accents before definitions. It's currently an advanced feature to sort dictionaries.

toasted-nutbread:
> There's a potential gotcha in just removing the headings when you have multiple pitch accent dictionaries, pitch accents and definitions use a separate ordered list and it could look weird:

It may also be possible to just show the index for both at all times, and offset the starting index of the standard dictionary entries. Not sure if this would be confusing though; could do something like change the color of the dictionary tag to more clearly indicate it's for a separate category.

ext/mixed/js/display-generator.js
toasted-nutbread: I'll look into making this implementation more consistent with the `only` implementation, but there are potentially ambiguous results with that method as well:

端/はた: Which kanji does the (はた only) apply to?
image

siikamiika: If you mean the JMdict's definition "third person", it applies to every kanji. The kanji is left out to keep it brief. It tries to emulate the original JMdict format which looks like this (from [my abandoned project](https://github.com/siikamiika/mecab-translate)):


I've also been thinking of a display format where the readings are grouped under each kanji, but that could be problematic with cross browser ruby text and Anki. Better to fix the data structure first anyway https://github.com/FooSoft/yomichan/issues/337.

siikamiika:
> It may also be possible to just show the index for both at all times, and offset the starting index of the standard dictionary entries. Not sure if this would be confusing though; could do something like change the color of the dictionary tag to more clearly indicate it's for a separate category.

Maybe it's best to keep it as simple as possible, this could also require significant changes to the Anki templates. I think a separator with separate indices and possibly different dictionary tag color are fine.

Speaking of Anki, it doesn't seem to be possible to export pitch accents to Anki right now. Do you know if it's possible to render SVG in the browser as PNG easily and how the downstep notation styles can be exported?

toasted-nutbread:
> Speaking of Anki, it doesn't seem to be possible to export pitch accents to Anki right now. Do you know if it's possible to render SVG in the browser as PNG easily

It's probably possible to export as a PNG using a `<canvas>` to render, and I think Anki also just supports using `<svg>` tags.

> and how the downstep notation styles can be exported?

Would probably have to just be different `{tags}` for the possible styles.

Anki support should probably be part of a follow-up PR so as to limit the scope of this PR.

siikamiika:
> > Speaking of Anki, it doesn't seem to be possible to export pitch accents to Anki right now. Do you know if it's possible to render SVG in the browser as PNG easily
> 
> It's probably possible to export as a PNG using a `<canvas>` to render, and I think Anki also just supports using `<svg>` tags.
> 
> > and how the downstep notation styles can be exported?
> 
> Would probably have to just be different `{tags}` for the possible styles.

Nice :+1: 

> Anki support should probably be part of a follow-up PR so as to limit the scope of this PR.

Sure, then I think the only issue is removing the headings and deciding how to handle indexing. Right now a horizontal bar separator seems like the best option to me.

ext/mixed/js/display-generator.js
toasted-nutbread: Updated in 0d80fcdf86745da133e4510eeea809a4eeafe120

toasted-nutbread:
Styles have been updated and the disambiguation notation is now more consistent with the existing "only" implementation.

ext/mixed/js/display-generator.js
```diff
@@ -498,10 +499,22 @@ class DisplayGenerator {
...
         }
 
+        for (const dictionaryResults of results.values()) {
+            for (const result of dictionaryResults) {
+                const exclusiveExpressions = [];
+                const resultExpressions = result.expressions;
+                if (!areSetsEqual(resultExpressions, allExpressions)) {
+                    exclusiveExpressions.push(...getSetIntersection(resultExpressions, allExpressions));
+                }
+                result.exclusiveExpressions = exclusiveExpressions;
```
siikamiika: I'd update it like this to handle readings as well:
```diff
diff --git a/ext/mixed/js/display-generator.js b/ext/mixed/js/display-generator.js
index 8177c31c..b6469d62 100644
--- a/ext/mixed/js/display-generator.js
+++ b/ext/mixed/js/display-generator.js
@@ -482,10 +482,13 @@ class DisplayGenerator {
         const results = new Map();

         const allExpressions = new Set();
+        const allReadings = new Set();
         const expressions = definition.expressions;
         const sources = Array.isArray(expressions) ? expressions : [definition];
         for (const {pitches: expressionPitches, expression} of sources) {
+            allExpressions.add(expression);
             for (const {reading, pitches, dictionary} of expressionPitches) {
+                allReadings.add(reading);
                 let dictionaryResults = results.get(dictionary);
                 if (typeof dictionaryResults === 'undefined') {
                     dictionaryResults = [];
@@ -499,7 +502,6 @@ class DisplayGenerator {
                         dictionaryResults.push(pitchInfo);
                     }
                     pitchInfo.expressions.add(expression);
-                    allExpressions.add(expression);
                 }
             }
         }
@@ -508,9 +510,13 @@ class DisplayGenerator {
             for (const result of dictionaryResults) {
                 const exclusiveExpressions = [];
                 const resultExpressions = result.expressions;
+                const resultReadings = new Set([result.reading]);
                 if (!areSetsEqual(resultExpressions, allExpressions)) {
                     exclusiveExpressions.push(...getSetIntersection(resultExpressions, allExpressions));
                 }
+                if (!areSetsEqual(resultReadings, allReadings)) {
+                    exclusiveExpressions.push(...getSetIntersection(resultReadings, allReadings));
+                }
                 result.exclusiveExpressions = exclusiveExpressions;
             }
         }
```

toasted-nutbread: > I'd update it like this to handle readings as well:

The readings are redundant when the downstep notation is also visible. Does it make sense to not display them for this case? (I think this is why I initially didn't add this, because I forgot that there were options to turn that off.)

image

siikamiika: Yeah, I think it's good to add that exception.

toasted-nutbread: Updated.

ext/mixed/js/display-generator.js
siikamiika: Found this old review comment after reviewing the history, feel free to move it there now or leave as is because it's just a minor code style issue.

PR #386, opened 2020-03-01
Simplify backend prepare

siikamiika:
https://github.com/FooSoft/yomichan/commit/2abf46b6fa9d6c03be0d98045e111f2d8e1e41d5 does to `Backend` what is discussed in https://github.com/FooSoft/yomichan/issues/356 (and https://github.com/FooSoft/yomichan/pull/381). This time all of the event handlers are also made `async` or `Promise` for consistency, but they don't check if the class is ready every time.

https://github.com/FooSoft/yomichan/commit/e6e5f23cf8481db31b94c18244f404dd3374ad90 fixes an issue where Chrome would display an alert when it's started so that Yomichan search page opens first, and possibly other cases where `Frontend` is starting to prepare its dependencies before `Backend` is prepared.

### DIFF
--- a/ext/bg/js/settings/backup.js
+++ b/ext/bg/js/settings/backup.js
@@ -362,10 +362,10 @@ async function _onSettingsResetConfirmClick() {
 
 // Setup
 
-window.addEventListener('DOMContentLoaded', () => {
+function backupInitialize() {
     document.querySelector('#settings-export').addEventListener('click', _onSettingsExportClick, false);
     document.querySelector('#settings-import').addEventListener('click', _onSettingsImportClick, false);
     document.querySelector('#settings-import-file').addEventListener('change', _onSettingsImportFileChange, false);
     document.querySelector('#settings-reset').addEventListener('click', _onSettingsResetClick, false);
     document.querySelector('#settings-reset-modal-confirm').addEventListener('click', _onSettingsResetConfirmClick, false);
-}, false);
+}
--- a/ext/bg/js/settings/backup.js
+++ b/ext/bg/js/settings/backup.js
@@ -120,7 +120,7 @@ async function _onSettingsExportClick() {
 // Importing
 
 async function _settingsImportSetOptionsFull(optionsFull) {
-    return utilIsolate(await utilBackend().setFullOptions(
+    return utilIsolate(utilBackend().setFullOptions(
         utilBackgroundIsolate(optionsFull)
     ));
 }
--- a/ext/mixed/js/core.js
+++ b/ext/mixed/js/core.js
@@ -269,16 +269,25 @@ const yomichan = (() => {
         constructor() {
             super();
 
+            this._isBackendPreparedResolve = null;
+            this._isBackendPreparedPromise = new Promise((resolve) => (this._isBackendPreparedResolve = resolve));
+
             this._messageHandlers = new Map([
+                ['backendPrepared', this._onBackendPrepared.bind(this)],
                 ['getUrl', this._onMessageGetUrl.bind(this)],
                 ['optionsUpdated', this._onMessageOptionsUpdated.bind(this)],
                 ['zoomChanged', this._onMessageZoomChanged.bind(this)]
             ]);
 
             chrome.runtime.onMessage.addListener(this._onMessage.bind(this));
+            chrome.runtime.sendMessage({action: 'yomichanOnline'});
         }
 
         // Public
+
+        prepare() {
+            return this._isBackendPreparedPromise;
+        }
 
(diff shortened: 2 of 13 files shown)


siikamiika:
Also, https://github.com/FooSoft/yomichan/commit/e6e5f23cf8481db31b94c18244f404dd3374ad90 isn't added to `bg/js/api.js` because it's only used by `Backend` and its dependencies after it's prepared.

siikamiika:
I'm thinking that maybe instead of polling `Backend` from API the `Yomichan` class in `core.js` could have a promise that resolves when `Backend` is ready

ext/bg/js/settings/backup.js
```diff
@@ -120,7 +120,7 @@ async function _onSettingsExportClick() {
 // Importing
 
 async function _settingsImportSetOptionsFull(optionsFull) {
-    return utilIsolate(await utilBackend().setFullOptions(
```
toasted-nutbread: I don't think places which use `utilBackend()` have a guarantee that the backend is fully prepared. `setFullOptions` may still need to internally wait for `prepare`, or `utilBackend` may need to wait.

toasted-nutbread: Looking at the new update that was pushed as I reviewed, honestly a bigger problem is the design of the settings page, and how none of it waits for initialization currently. Similar to how frontend-initialize.js is the main entry point for content scripts, at some point there should be a main entry point for all settings handler scripts. That's a large change outside the scope of this one though IMO.

ext/mixed/js/core.js
```diff
@@ -269,17 +269,26 @@ const yomichan = (() => {
...
+
             this._messageHandlers = new Map([
+                ['backendPrepared', this._onBackendPrepared.bind(this)],
                 ['getUrl', this._onMessageGetUrl.bind(this)],
                 ['optionsUpdated', this._onMessageOptionsUpdated.bind(this)],
                 ['zoomChanged', this._onMessageZoomChanged.bind(this)]
             ]);
 
             chrome.runtime.onMessage.addListener(this._onMessage.bind(this));
+            chrome.runtime.sendMessage({action: 'yomichanOnline'});
```
toasted-nutbread: core.js is used in the background window also, and you won't want this called from there. Also, a more descriptive name might be better, like `'yomichanCoreReady'` or something.

ext/bg/js/settings/backup.js
siikamiika: Do you think it would help if `await yomichan.prepare()` was called from `onReady` in `bg/js/settings/main.js`?

toasted-nutbread: Actually looking back at the code, it looks like that's the primary place, I thought there were more places where the scripts initialized themselves. So you're right.

May also be worth it to have utilBackend() do a synchronous check to see if the backend is ready before it returns it, and throws an error if it's not, since that should be a programmer error if that happens.

toasted-nutbread:
FYI the CI fail is coming from test-database.js. The [`chrome.runtime`](https://github.com/FooSoft/yomichan/blob/46fee07d36b3966af9bacca8c6253b044bde07ee/test/test-database.js#L26-L35) object needs a dummy `sendMessage`:

```cs
        sendMessage() {
            // NOP
        }
```

ext/bg/js/settings/backup.js
siikamiika: Addressed in https://github.com/FooSoft/yomichan/pull/386/commits/e0edb30efd51ac18167880d77c2dea11c73a1bfc

ext/mixed/js/core.js
siikamiika: This should be fixed in https://github.com/FooSoft/yomichan/pull/386/commits/56b2f2c853494b228aa4467a64ed91486432bffd.

The reason why `'backendPrepared'` can be triggered from two places is that `Backend` might not be up when `'yomichanCoreReady'` is sent. I'm not sure if it's possible that both events are missed by `Yomichan`, but I haven't seen that happen yet. Logically no, but browsers aren't always perfect.

siikamiika:
The only remaining entry point I know of is `browser_action` (`context.html`). It doesn't seem to support sending messages to `Backend` and I'm not sure what would be the best way to solve the issue without without too much complexity.

siikamiika:
Took me way too long to figure out why the messages wouldn't work both ways between browser_action and Backend, but everything should be done in https://github.com/FooSoft/yomichan/pull/386/commits/e6347a94e7ff0b7a5c1db7e6f83f50d93c2d4545

PR #387, opened 2020-03-03
Abstract Yomichan extension script execution

toasted-nutbread:
Use node [`vm`](https://nodejs.org/api/vm.html) module instead of `Function` constructors. This makes it possible to have the declarative environment record variables (i.e. variables defined using `let`, `const`, and `class`) implicitly passed around using the `context` variable. This also makes any error messages have a more meaningful stack trace, contrasted to [the stack trace seen with the `Function` constructor](https://github.com/FooSoft/yomichan/runs/478639605#step:7:15).

The only downside to this is that `assert.deepStrictEqual` doesn't work for certain values since `assert.deepStrictEqual` compares object prototypes using `===`, and the object prototypes of cross-context objects will be different. Therefore, `vm.assert` was added with a relaxed version of `deepStrictEqual`.

### DIFF
--- a/test/dictionary-validate.js
+++ b/test/dictionary-validate.js
@@ -18,10 +18,12 @@
 
 const fs = require('fs');
 const path = require('path');
-const yomichanTest = require('./yomichan-test');
+const {JSZip} = require('./yomichan-test');
+const {VM} = require('./yomichan-vm');
 
-const JSZip = yomichanTest.JSZip;
-const {JsonSchema} = yomichanTest.requireScript('ext/bg/js/json-schema.js', ['JsonSchema']);
+const vm = new VM();
+vm.execute('bg/js/json-schema.js');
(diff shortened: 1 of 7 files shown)


siikamiika:
> The only downside to this is that `assert.deepStrictEqual` doesn't work for certain values since `assert.deepStrictEqual` compares object prototypes using `===`, and the object prototypes of cross-context objects will be different. Therefore, `vm.assert` was added with a relaxed version of `deepStrictEqual`.

I don't think tokens like the ones used in `Display` (`setContentToken`) are tested at the moment, but that kind of comparisons could have issues with this approach.

toasted-nutbread:
> I don't think tokens like the ones used in `Display` (`setContentToken`) are tested at the moment, but that kind of comparisons could have issues with this approach.

It's still possible to do tests like that if necessary, they just aren't implemented yet since they weren't necessary. Can of course be added when necessary.

Issue #320, opened 2019-12-31
[Feature Request] Option to create different note type


Pintliz:
Currently the only way to utilize multiple types of cards(e.g. vocab-front, sentence-front, cloze) is to generate them from the same note. But this would cause all notes to have multiple cards. 

I currently use vocab-front card. Sometimes I encounter a vocab that's already added in a different context with a different definition used. It'd be helpful if I can add it again as a sentence-front card.

It would be very helpful if there is an option in settings to enable the creation of different note types, which could appear as a plus button of different color in the popup.

toasted-nutbread:
So are you essentially asking for a way to have multiple Anki card templates and have some button/menu to add a term using a different (non-default) template if desired?

Pintliz:
yes

oudajosefu:
I would also like this feature added as my current method uses two different templates and I have to constantly manually change the Deck, Model, and fields myself to switch between them.

Pintliz:
one other possible solution that came to my mind is to have different modifier keys trigger different profiles.(a new type of profile usage condition)
I don't know if this is easier to implement than the original request though.

Pintliz:
#389 possible workaround

siikamiika:
@Pintliz I'll investigate the modifier key approach next, it seems like a good idea and simple to present to the user unlike yet another button or dropdown in https://github.com/FooSoft/yomichan/pull/389

siikamiika:
@Pintliz https://github.com/FooSoft/yomichan/pull/487

Issue #349, opened 2020-02-03
Kanji information page dividers


psychodelicwasabi:
I thought about adding this to my previous submission, but I guess they're not so close to be dealt with in the same place, so I made a new one. 
 Ever since the built-in themes updates, the lines separating the kanji information vanished. More than an aesthetic issue, I guess that makes sorting the information a slightly harder, which is quite annoying.
 Speaking about aesthetics, though, I'd just like to put in words how much I disliked the change to the little "open book" you click to go forward or backwards on the pop-ups. I know that was a request by an user, but I'd just like to have it as it used to be, if possible.

toasted-nutbread:
> Ever since the built-in themes updates, the lines separating the kanji information vanished. More than an aesthetic issue, I guess that makes sorting the information a slightly harder, which is quite annoying.

This has actually been fixed in #317, but it is not in a stable release yet. The testing release does include it.

> Speaking about aesthetics, though, I'd just like to put in words how much I disliked the change to the little "open book" you click to go forward or backwards on the pop-ups. I know that was a request by an user, but I'd just like to have it as it used to be, if possible.

This has also been changed such that the icons and the extra space are not shown if there is no history (also not in a stable release yet). Does that address your issue with it?

psychodelicwasabi:
> This has actually been fixed in #317, but it is not in a stable release yet. The testing release does include it.

Great!

>This has also been changed such that the icons and the extra space are not shown if there is no history (also not in a stable release yet). Does that address your issue with it?

I'm not sure, my problem with it is how it appears above the rest of the icons, while I liked how it used to appear by their side in the past.

siikamiika:
> I'm not sure, my problem with it is how it appears above the rest of the icons, while I liked how it used to appear by their side in the past.

I changed it to be a button that exists regardless of your scroll position to be consistent with how web browsers work. You can now use it even if you scroll the other icons related to single results out of view. So the changes were added from functional standpoint. I understand that it's hard to get used to something you've used differently for a long time, and it should be quite simple to add alternative display options for the history button(s) using the framework added in #317 if you don't like how it works in the current testing release.

PR #389, opened 2020-03-05
Switch matching profiles

### LINKED ISSUE #388: quick way to switch between profiles
Is there a quick way to switch between profiles? For example some sort of hotkey 
I have two profiles each for a different anki template, I go to settings and switch up the profile whenever I want to make a different type of card.
It would be helpful if there's a way to do this faster.

siikamiika:
The first few commits are refactoring and framework building, and https://github.com/FooSoft/yomichan/commit/c7d0df4da89453025785a6e5bbde741dc19f3d1e is the actual feature.

The pattern used in https://github.com/FooSoft/yomichan/commit/f59e760b8281682019a51ce7f992d1e2dbbd9ebb helps with https://github.com/FooSoft/yomichan/issues/368. It took me a while to completely (maybe) undestand how the popup proxy framework works, and did some misguided work on https://github.com/siikamiika/yomichan/commit/875b04bf34c1ebe5c8eebe906495007c3df32b66. While most of it is useless, I think the "selective flooding algorithm" can be reused for iframes https://github.com/FooSoft/yomichan/issues/29 https://github.com/FooSoft/yomichan/pull/364.

I'd appreciate some ideas about displaying the dropdown menu or profile selection. It can be quite ugly depending on the operating system and theme.

Fixes #388

Possible workaround for #320, #378, #379

### DIFF
--- a/ext/bg/js/backend.js
+++ b/ext/bg/js/backend.js
@@ -237,22 +237,29 @@ class Backend {
     }
 
     getProfile(optionsContext) {
-        const profiles = this.options.profiles;
-        if (typeof optionsContext.index === 'number') {
-            return profiles[optionsContext.index];
-        }
-        const profile = this.getProfileFromContext(optionsContext);
-        return profile !== null ? profile : this.options.profiles[this.options.profileCurrent];
+        return this._getProfilesFromContext(optionsContext).next().value;
     }
 
-    getProfileFromContext(optionsContext) {
-        for (const profile of this.options.profiles) {
+    *_getProfilesFromContext(optionsContext) {
+        const {profiles, profileCurrent} = this.options;
+
+        if (typeof optionsContext.index === 'number') {
+            yield profiles[optionsContext.index];
+            return;
+        }
+
+        let contextHasResults = false;
+        for (const profile of profiles) {
             const conditionGroups = profile.conditionGroups;
             if (conditionGroups.length > 0 && Backend.testConditionGroups(conditionGroups, optionsContext)) {
-                return profile;
+                contextHasResults = true;
+                yield profile;
             }
         }
-        return null;
+
+        if (!contextHasResults) {
+            yield profiles[profileCurrent];
+        }
     }
 
     static testConditionGroups(conditionGroups, data) {
--- a/ext/bg/js/backend.js
+++ b/ext/bg/js/backend.js
@@ -238,7 +238,12 @@ class Backend {
     }
 
     getProfile(optionsContext) {
-        return this._getProfilesFromContext(optionsContext).next().value.profile;
+        for (const profile of this._getProfilesFromContext(optionsContext)) {
+            if (profile.selected) {
+                return profile.profile;
+            }
+        }
+        return null;
     }
 
     getMatchingProfiles(optionsContext) {
--- a/ext/bg/js/backend.js
+++ b/ext/bg/js/backend.js
@@ -247,7 +247,19 @@ class Backend {
     }
 
     getMatchingProfiles(optionsContext) {
-        return Array.from(this._getProfilesFromContext(optionsContext));
+        const matchingProfiles = [];
+        let selectedProfileIndex = 0;
+        let profileIndex = 0;
+
+        for (const {profile, index, selected} of this._getProfilesFromContext(optionsContext)) {
+            if (selected) {
+                selectedProfileIndex = profileIndex;
+            }
+            matchingProfiles.push({profile, index});
+            ++profileIndex;
+        }
+
+        return {matchingProfiles, selectedProfileIndex};
     }
 
     *_getProfilesFromContext(optionsContext) {
--- a/ext/bg/js/backend.js
+++ b/ext/bg/js/backend.js
@@ -79,6 +79,7 @@ class Backend {
             ['optionsGetFull', this._onApiOptionsGetFull.bind(this)],
             ['optionsSet', this._onApiOptionsSet.bind(this)],
             ['optionsSave', this._onApiOptionsSave.bind(this)],
+            ['profilesGetMatching', this._onApiProfilesGetMatching.bind(this)],
             ['kanjiFind', this._onApiKanjiFind.bind(this)],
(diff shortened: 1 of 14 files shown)


ext/bg/js/backend.js
```diff
@@ -212,22 +213,35 @@ class Backend {
...
-                return profile;
+                contextHasResults = true;
+                yield {index: profileIndex, profile};
             }
+            ++profileIndex;
+        }
+
+        if (!contextHasResults) {
+            yield {index: profileCurrent, profile: profiles[profileCurrent]};
         }
```
siikamiika: Could be useful to yield all profiles but also indicate which is the currently selected profile in case the user has not specified any conditions or none of them match the current context

siikamiika: Fixed in https://github.com/FooSoft/yomichan/pull/389/commits/5d93d0d853618dc9537327a85577f10cc9ac5457

siikamiika:
There are potentially some issues preparing `PopupProxy` before accessing its ID from `Frontend`, but I consider that out of scope of this feature.

siikamiika:
I saw this error a few times while testing, but on Firefox only. The problem seems to be that `PopupProxy.containsPoint` is called so that `BackendApiForwarder` either disconnects or doesn't exist yet. The latter seems improbable because it only happens with nested popups when `Backend` has been used already. It's difficult to reproduce, and I'm not sure if it's even caused by these changes. Every time has been a bit after reloading Yomichan and opening a new tab for scanning, so maybe the messaging system is just overloaded when this happens.

```
Yomichan v20.2.24.0 has encountered an error.
Originating URL: moz-extension://567019fb-43ef-4b76-b9cb-9984dcf73609/fg/float.html
Error: Disconnected
invoke@moz-extension://567019fb-43ef-4b76-b9cb-9984dcf73609/fg/js/frontend-api-sender.js:34:35
_invokeHostApi@moz-extension://567019fb-43ef-4b76-b9cb-9984dcf73609/fg/js/popup-proxy.js:129:32
containsPoint@moz-extension://567019fb-43ef-4b76-b9cb-9984dcf73609/fg/js/popup-proxy.js:84:27
Frontend/<@moz-extension://567019fb-43ef-4b76-b9cb-9984dcf73609/fg/js/frontend.js:26:35
searchAt@moz-extension://567019fb-43ef-4b76-b9cb-9984dcf73609/mixed/js/text-scanner.js:285:27
search@moz-extension://567019fb-43ef-4b76-b9cb-9984dcf73609/mixed/js/text-scanner.js:75:24
onMouseMove@moz-extension://567019fb-43ef-4b76-b9cb-9984dcf73609/mixed/js/text-scanner.js:78:9
EventListener.handleEvent*addEventListener@moz-extension://567019fb-43ef-4b76-b9cb-9984dcf73609/mixed/js/core.js:249:14
hookEvents@moz-extension://567019fb-43ef-4b76-b9cb-9984dcf73609/mixed/js/text-scanner.js:246:33
setEnabled@moz-extension://567019fb-43ef-4b76-b9cb-9984dcf73609/mixed/js/text-scanner.js:227:22
setOptions@moz-extension://567019fb-43ef-4b76-b9cb-9984dcf73609/mixed/js/text-scanner.js:273:14
updateOptions@moz-extension://567019fb-43ef-4b76-b9cb-9984dcf73609/fg/js/frontend.js:129:14
async*prepare@moz-extension://567019fb-43ef-4b76-b9cb-9984dcf73609/fg/js/frontend.js:55:24
main@moz-extension://567019fb-43ef-4b76-b9cb-9984dcf73609/fg/js/frontend-initialize.js:38:20
async*@moz-extension://567019fb-43ef-4b76-b9cb-9984dcf73609/fg/js/frontend-initialize.js:41:1

Issues can be reported at https://github.com/FooSoft/yomichan/issues
```

siikamiika:
https://github.com/FooSoft/yomichan/pull/389/commits/7b7db897216b9885563fc562d621c835745e157d could have actually been done on master first and rebased on this, but it doesn't change anything in practice except one property access so let's include it here while we're at it.

toasted-nutbread:
So you're right, this is something that will need a UI/UX pass before fully implementing, since there have been issues before with users disliking some UI layout settings (#349).

I have created #402 to have a separate discussion on UI design changes. Changing the existing UI design should be considered outside the scope of this PR.

siikamiika:
Ok, I think I'll go with the current design and the profile name to the floating header, making it only visible when there are more profiles than one. After that I'll add some hotkeys to switch between them.

siikamiika:
It should now be usable with a reasonable amount of active profiles without changing the current layout too much.

siikamiika:
Sorry for going back and forth with the commits, I can try to squash related ones at some point

ext/mixed/js/display.js
```diff
@@ -354,15 +367,29 @@ class Display {
...
+        const options = this.options;
+        this.updateDocumentOptions(options);
+        this.updateTheme(options.general.popupTheme);
+        this.setCustomCss(options.general.customPopupCss);
+    }
+
     getOptionsContext() {
-        throw new Error('Override me');
+        return {
+            index: this.profileSwitcher.globalProfileIndex
```
siikamiika: This could be safer if instead of the index a unique ID was used. It could be either autoincrement from the DB or a randomly generated number. Not sure if same applies to backend/options because the index comes from there.

ext/bg/js/search.js
```diff
@@ -271,19 +268,24 @@ class DisplaySearch extends Display {
...
-    async updateOptions(options) {
-        await super.updateOptions(options);
+    onProfileChanged(profileIndex) {
+        super.onProfileChanged(profileIndex);
+        const query = this.query.value;
+        if (query) {
+            this.setQuery(query);
+            this.onSearchQueryUpdated(query, false);
+        }
+    }
```
siikamiika: This will be called from `prepare` before `QueryParser` has been prepared. It throws an error about missing template handler. Will have to look for a workaround, for example using a prepared flag.

PR #391, opened 2020-03-06
remove bg api.js

siikamiika:
Discussed in https://github.com/FooSoft/yomichan/pull/358


ext/bg/js/background-globals.js
```diff
@@ -0,0 +1,19 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+window.isBackground = true;
```
toasted-nutbread: Could probably save a file by putting this into core.js by doing something like:

```js
// In Yomichan constructor
this.isBackground = (
    typeof chrome.runtime.getBackgroundPage === 'function' &&
    chrome.runtime.getBackgroundPage() === window
);
```

I think the window references are the same, but I didn't test.

siikamiika: I'm not sure how that would work on browser_action, have to test

toasted-nutbread:
So the reason I didn't go this route is because the functionality will actually be different for certain functions, in particular the ones which use the `sender` information:

https://github.com/FooSoft/yomichan/blob/58d734b8f79a9d859063ef70afe1ff367afe7a29/ext/bg/js/backend.js#L517-L526

While these cases do throw an error, having the entire API exposed to the backend seems like overkill, particularly when only 3 functions are used, and the reason behind their usage could be attributed to other design issues. For example, we've mentioned a few times that the contents of dictionary.js could really be internals of translator.js. That file uses `apiTemplateRender`, and that's the only use in the entire project.

An alternative to this is dependency injection, which is already used for the `Frontend` class and the `popup` instance it ses. The `popup` can either be an instance of `Popup` or `PopupProxy`, both of which can be though of implementing an implicit common interface. Of course since this is Javascript, an explicit interface isn't necessary because of duck typing.

Applying a similar concept to `Translator` could look like the following:

```js
class Backend {
    constructor() {
        this._translator = new Translator(this);
    }
    renderTemplate(templates, data) {
        // ...
    }
}

class Translator {
    constructor(templateRenderer) {
        this._templateRenderer = templateRenderer;
    }
    dictFieldFormat(...args) {
        this._templateRenderer.renderTemplate(...something);
    }
}
```

Or:

```js
class Backend {
    constructor() {
        this._translator = new Translator();
        this._translator.renderTemplate = (...args) => this._renderTemplate(...args);
    }
    _renderTemplate(templates, data) {
        // ...
    }
}

class Translator {
    renderTemplate() {
        throw new Error('Template rendering not supported');
    }
    dictFieldFormat(...args) {
        this.renderTemplate(...something);
    }
}
```

Both are effectively doing the same thing, but it's Javascript so there's flexibility to do things different ways. Both patterns are used in the project currently.

Anyways, the reason I'm bringing this up is for a few reasons:
* Static/global variables can be indicative of design problems, and the less that are used, the better. In this case, the one I'm mentioning is `window.yomichanBackend`. I realize that there are some other places where globals are used, such as the `yomichan` object, but I would argue that that's more used to establish a common framework, similar to the `chrome` object.
* This isn't necessarily solving the design issues I had in mind in #358. I supposed I never fully explained what I had in mind, but the main point is that some of the functions which use the `api*` functions on the background page should really either be part of another class or use a reference to another class which has the methods.
* The changes made in this PR are valid, but could obscure the underlying design issues and further encourage use of `api*` functions where an alternative should be used.

(sorry for wall of text)

siikamiika:
Thanks for the thorough explanation, I didn't realize that this was the actual issue behind wanting to get rid of the bg api.js. To me the issue was that I wasn't sure when to use which although they looked like the same to the API user. There's a similar issue with `ClipboardMonitor` because it's used in both frontend and backend, and it seems like it could be changed to use dependency injection, too.

When you think about it, what's done in this PR is really a global scope dependency injection for every API function. Like you said, it shows that some of the API functions have been designed with the origin in mind.

I noticed the same issue when doing https://github.com/FooSoft/yomichan/pull/386 and trying to receive a message from `browser_action`. Now that I learned about `chrome.extension.getBackgroundPage()`, it seems that the current handling is unnecessary:
https://github.com/FooSoft/yomichan/blob/24d9001b1449a318113f5ecc5441113cabe7248f/ext/bg/js/backend.js#L281-L292

Instead, you could simply replace
https://github.com/FooSoft/yomichan/blob/24d9001b1449a318113f5ecc5441113cabe7248f/ext/mixed/js/core.js#L287-L290
with
```javascript
prepare() {
    if (typeof chrome.extension.getBackgroundPage === 'function') {
        const {yomichanBackend} = chrome.extension.getBackgroundPage();
        if (typeof yomichanBackend !== 'undefined' && yomichanBackend.isPrepared) {
            this._isBackendPreparedResolve();
        }
    } else {
        chrome.runtime.sendMessage({action: 'yomichanCoreReady'});
    }
    return this._isBackendPreparedPromise;
}
```

I'll look into fixing `ClipboardMonitor` and `Yomichan` API usage instead of doing this and exposing the entire API to the backend. This PR can be used for tracking and closed later.

siikamiika:
Actually I just realized that you said `window.yomichanBackend` could be bad. :smile: I'll hear any comments about my suggestion before starting to implement anything.

toasted-nutbread:
This does pose a stylistic question: which of the following should be preferred? Taking `ClipboardMonitor` as an example:

1. ```js
   class Something {
     constructor() {
       this.clipboardMonitor = new ClipboardMonitor(this);
     }
   }
   ```

2. ```js
   class Something {
     constructor() {
       this.clipboardMonitor = new ClipboardMonitor();
       this.clipboardMonitor.onClipboardText = this.onClipboardText.bind(this);
     }
   }
   ```

3. ```js
   class Something {
     constructor() {
       this.clipboardMonitor = new ClipboardMonitor({
         onClipboardText: this.onClipboardText.bind(this)
       });
     }
   }
   ```

Comparison:
1. `ClipboardMonitor` uses `Something`, but it's unclear how/what APIs it uses. Has potential for use creep in ways that weren't intended.
2. Dependencies are explicit and `ClipboardMonitor` doesn't have an explicit reference to `Something`. The assignment of overrides can be located somewhere away from the constructor call which could obscure setup of the object. This is currently how `ClipboardMonitor` is set up: the `onClipboardText` assignment is done in `prepare`.
3. Dependencies are explicit and they are all assigned in the constructor. This could permit `ClipboardMonitor` to store the `onClipboardText` reference "privately". This design is the similar to the "options object argument pattern" used frequently in Javascript. Examples [1](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver/observe) [2](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/query) [3](https://nodejs.org/api/http.html#http_new_agent_options)

toasted-nutbread:
> Actually I just realized that you said `window.yomichanBackend` could be bad. 😄 I'll hear any comments about my suggestion before starting to implement anything.

One reason I would suggest not using `yomichanBackend`, particularly from `getBackgroundPage()`, is that the window context will be different. When not carefully handled, the result can be dead objects in Firefox and prototype differences in general. This is why see many `isolate` calls in util.js, e.g.:

https://github.com/FooSoft/yomichan/blob/24d9001b1449a318113f5ecc5441113cabe7248f/ext/bg/js/util.js#L95-L100

Another reason is consistency. While extension pages do have access to the full set of chrome APIs that content scripts do not, it's probably better to have less code paths which function differently depending on whether or not something is a content script or not. The result should be less maintenance and potential for bugs in different scenarios, as the implementation would be the same for both cases.

siikamiika:
> This does pose a stylistic question: which of the following should be preferred? Taking `ClipboardMonitor` as an example:

I think option 3 is the most clean because it places the used dependencies near the place where the class is instantiated. If it would be overridden later, conditionally or otherwise, you would have to know to look for `this._classInstance` in the file and that's not obvious.

I think it's safer to override methods inside a class with something else, though, because then they will usually have a stub in place in the same file.

toasted-nutbread:
This is now resolved with #397, #399, and #400.

siikamiika:
Thanks!

Issue #394, opened 2020-03-07
Popup.containsPoint offset wrong when scanning a popup


siikamiika:
I tested with code similar to what's used in https://toasted-nutbread.github.io/firefox-touch-position-bug/, and it seems that the check is done with coordinates originating from the window that's scanned in `Frontend`. For popups that's an iframe.

My first idea was to fix the offset in `containsPoint` using the parent popup offset, but that's not necessarily correct in all cases.

edit: actually it might be correct, but I'm not completely sure yet

siikamiika:
For reference, here's the code I used to test https://github.com/siikamiika/yomichan/commit/baf7d9178f342ec9c603da8992d506c1c82cdb47

PR #395, opened 2020-03-07
adjust containsPoint offset with nested popups

siikamiika:
The point can originate from within a popup where the coordinates start from the top left corner of the popup iframe. containsPoint check is done on the root page, and thus the offset has to be adjusted to match the point inside the iframe rect.

Fixes https://github.com/FooSoft/yomichan/issues/394

### DIFF
--- a/ext/fg/js/popup-proxy-host.js
+++ b/ext/fg/js/popup-proxy-host.js
@@ -116,8 +116,8 @@ class PopupProxyHost {
 
     async _onApiContainsPoint({id, x, y}) {
         const popup = this._getPopup(id);
-        const rootPagePoint = PopupProxyHost._convertPopupPointToRootPagePoint(popup, x, y);
-        return await popup.containsPoint(...rootPagePoint);
+        [x, y] = PopupProxyHost._convertPopupPointToRootPagePoint(popup, x, y);
+        return await popup.containsPoint(x, y);
     }
 
     async _onApiShowContent({id, elementRect, writingMode, type, details}) {
--- a/ext/fg/js/popup-proxy-host.js
+++ b/ext/fg/js/popup-proxy-host.js
@@ -116,7 +116,8 @@ class PopupProxyHost {
 
     async _onApiContainsPoint({id, x, y}) {
         const popup = this._getPopup(id);
-        return await popup.containsPoint(x, y);
+        const rootPagePoint = PopupProxyHost._convertPopupPointToRootPagePoint(popup, x, y);
+        return await popup.containsPoint(...rootPagePoint);
     }
 
     async _onApiShowContent({id, elementRect, writingMode, type, details}) {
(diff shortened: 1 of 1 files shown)


toasted-nutbread:
Is this not handled with the following?

https://github.com/FooSoft/yomichan/blob/421b60db0f6f132cbc5993488041846055a76d11/ext/fg/js/popup-proxy.js#L128-L135

https://github.com/FooSoft/yomichan/blob/421b60db0f6f132cbc5993488041846055a76d11/ext/fg/js/popup-proxy-host.js#L154-L163

siikamiika:
That seems to do a similar offset fix, but I think it's actually not used in `Popup.containsPoint`. The popups in there are not proxies.

toasted-nutbread:
What is the scenario in which this is happening? Coordinates passed to `Popup.containsPoint` should already be in the correct coordinate space without `Popup` having to transform them.

siikamiika:
I'm scanning the second nested popup in a tight space on the search page. You can test with this branch https://github.com/siikamiika/yomichan/commit/baf7d9178f342ec9c603da8992d506c1c82cdb47

siikamiika:
Now that you brought up the `PopupProxyHost` private function, maybe the conversion should be moved there instead.

To me it looks like that `PopupProxyHost._convertJsonRectToDOMRect` is only used with `PopupProxyHost._onApiShowContent`, but incoming coordinates (in this case from the anonymous function in `Frontend`'s `TextScanner`) are not converted.

toasted-nutbread:
Yeah, the conversion should be in `_onApiContainsPoint`. The popup proxy host is in charge of converting points so that no special handling needs to be added to `Popup`.

ext/fg/js/popup-proxy-host.js
```diff
@@ -116,7 +116,8 @@ class PopupProxyHost {
 
     async _onApiContainsPoint({id, x, y}) {
         const popup = this._getPopup(id);
-        return await popup.containsPoint(x, y);
+        const rootPagePoint = PopupProxyHost._convertPopupPointToRootPagePoint(popup, x, y);
```
toasted-nutbread: For consistency with line 156, this should probably just destructure to `[x, y]`. The spread is going to do so anyways, so I don't think it saves anything.

siikamiika: I somehow thought I had to keep the variable names the same, but that's not required because `popup.containsPoint` doesn't take an object that has to be restructured from the existing names. I'll destructure it to `rootX` and `rootY` and pass them on to the function.

siikamiika: Oh, I didn't know that you can also destructure without assignment!

toasted-nutbread: You can do `[x, y] = getArray();` without needing `let` or `const`. Technically you can do `[x, y] = [y, x]` to swap. The syntax is a bit more strange when destructing an object, because you can't do inline objects due to being ambiguous with block scopes, so you have to do `({x, y}) = getObject();`

siikamiika: Makes sense, I think that's why I had the idea that you couldn't do what I did in https://github.com/FooSoft/yomichan/pull/395/commits/9fef0751f3dc7f213f158a50124371f0c6fb5a17.

PR #397, opened 2020-03-07
Clipboard monitor refactor

toasted-nutbread:
Based on some discussion from #391. A bit contrary to what I described in https://github.com/FooSoft/yomichan/pull/391#issuecomment-596096746, I converted `onClipboardText` to an event, but the comment could still apply to other scenarios where the function isn't as event-like, e.g. `ConditionsUI.Container.save`, though arguably this could be done using an event also:

https://github.com/FooSoft/yomichan/blob/24d9001b1449a318113f5ecc5441113cabe7248f/ext/bg/js/settings/profiles.js#L94-L104

### DIFF
--- a/ext/bg/js/clipboard-monitor.js
+++ b/ext/bg/js/clipboard-monitor.js
@@ -18,16 +18,13 @@
 
 /*global apiClipboardGet, jpIsStringPartiallyJapanese*/
 
-class ClipboardMonitor {
+class ClipboardMonitor extends EventDispatcher {
     constructor() {
+        super();
         this.timerId = null;
         this.timerToken = null;
         this.interval = 250;
         this.previousText = null;
-    }
-
-    onClipboardText(_text) {
-        throw new Error('Override me');
     }
 
     start() {
@@ -55,7 +52,7 @@ class ClipboardMonitor {
             ) {
                 this.previousText = text;
                 if (jpIsStringPartiallyJapanese(text)) {
-                    this.onClipboardText(text);
+                    this.trigger('change', {text});
                 }
             }
 
--- a/ext/bg/js/clipboard-monitor.js
+++ b/ext/bg/js/clipboard-monitor.js
@@ -16,22 +16,23 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
 
-/*global apiClipboardGet, jpIsStringPartiallyJapanese*/
+/*global jpIsStringPartiallyJapanese*/
 
 class ClipboardMonitor extends EventDispatcher {
-    constructor() {
+    constructor({getClipboard}) {
         super();
         this.timerId = null;
         this.timerToken = null;
         this.interval = 250;
         this.previousText = null;
+        this.getClipboard = getClipboard;
     }
 
     start() {
         this.stop();
 
         // The token below is used as a unique identifier to ensure that a new clipboard monitor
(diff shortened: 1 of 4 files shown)


ext/bg/js/clipboard-monitor.js
```diff
@@ -55,7 +52,7 @@ class ClipboardMonitor {
             ) {
                 this.previousText = text;
                 if (jpIsStringPartiallyJapanese(text)) {
-                    this.onClipboardText(text);
+                    this.trigger('change', {text});
```
siikamiika: I think this would be nicer to use if it was changed to
```javascript
this.trigger('change', [text]);
```

Then you could use
```javascript
this.clipboardMonitor.on('change', (args) => this.onExternalSearchUpdate(...args));
```

and keep
```javascript
onExternalSearchUpdate(text) {
```
and
```javascript
_onClipboardText(text) {
```
as is.

https://github.com/FooSoft/yomichan/issues/359#issuecomment-592934361

Another consistent option would be to go with object destructuring everywhere.

siikamiika: My motivation is that then you'll only have to know the arguments at the place where the function is called and where it ends up, not in the middle. It reduces the amount of work and probability of errors when doing changes. I got bitten by it again while doing changes here:

https://github.com/FooSoft/yomichan/pull/389/files#diff-7f1c7cabf21b3d982801b7f56c4f984e

toasted-nutbread: The reason I did `{text}` is to mimic an event object, similar to DOM events.
```js
document.body.onclick = (e) => {
  if (e.altKey) { /* do something */ }
};
```

In this case:
```js
something.on('change', (e) => {
  use(e.text);
}
```

Technically I could just pass `text` as the argument without structuring, but IMO using an object is more event-like. Another potential issue with passing directly doesn't apply in this case, but consider if there were more arguments:

```js
// for some event dispatcher "something" and variables text, value1, value2:
something.trigger('change', text, value1, value2);
// ...
something.on('change', (text/*, value1, value2*/) => {
  use(text); // text used; value1, value2 unused and can be omitted
});
something.on('change', (text, value1, value2) => {
  use(value2); // value2 used; text, value1 unused but cannot be omitted
});

// Alternatives which only use what they need without having to declare anything extra
something.trigger('change', {text, value1, value2});
// ...
something.on('change', ({text}) => {
  use(text);
});
something.on('change', ({value2}) => {
  use(value2);
});
```

siikamiika: Looks a lot more familiar when there's the `e` :smile: In that light, maybe the approach in this PR is the best way to go, because it addresses the problem with repeating arguments but is just a bit different from other functions around it that don't take in events.

There's a way to make the first argument of a function optional, but it's pretty confusing: https://developer.chrome.com/extensions/runtime#method-sendMessage

I guess that internally checks the type of the first two arguments. I wouldn't want that kind of style.

toasted-nutbread: Yeah, in at least one place, the structure of the arguments must be known. Using `.bind` puts this burden on the bound function to ensure it is using the expected argument types. The other places you are mentioning have 3 places which need to know the structure of the arguments: 2 for the handler arrow function and 1 for the target function.

And I agree on early optional function arguments; they are typically ugly and should be avoided since there's no automagic that does it for us.

PR #400, opened 2020-03-07
Template render refactor

toasted-nutbread:
Addresses #391, #358 concerns related to `apiTemplateRender`.


toasted-nutbread:
Related but not implemented yet: I think that the `_markers` check should be removed, since it prevents new templates defined in the settings from working, which could be useful for testing and some users.

https://github.com/FooSoft/yomichan/blob/5174ac6bbef3836b89aaf18f285a247cc1dc8941/ext/bg/js/anki-note-builder.js#L22-L41
https://github.com/FooSoft/yomichan/blob/5174ac6bbef3836b89aaf18f285a247cc1dc8941/ext/bg/js/anki-note-builder.js#L97-L99

siikamiika:
> Related but not implemented yet: I think that the `_markers` check should be removed, since it prevents new templates defined in the settings from working, which could be useful for testing and some users.

I thought the same earlier when trying to guide a user to achieve advanced stuff with the templates. Can be quite unintuitive when you create a new inline templates and they don't work.

toasted-nutbread:
Removed marker restrictions in 426c1534e7d740fa2c30488a64ad4fa6a382deed

PR #399, opened 2020-03-07
Audio system refactor

toasted-nutbread:
Addresses #391, #358 concerns related to `apiAudioGetUrl`.

PR #401, opened 2020-03-08
Audio refactor

toasted-nutbread:
As mentioned in #399, this refactors the other audio.js source.


ext/mixed/js/display.js
```diff
@@ -18,7 +18,7 @@
 
 /*global docRangeFromPoint, docSentenceExtract
 apiKanjiFind, apiTermsFind, apiNoteView, apiOptionsGet, apiDefinitionsAddable, apiDefinitionAdd
-apiScreenshotGet, apiForward, apiAudioGetUrl
+apiScreenshotGet, apiForward, apiAudioGetUri
```
siikamiika: Should these be made one per line, such as
```javascript
/* global
 * apiScreenshotGet
 * apiForward
 * apiAudioGetUri
 */ 
```

At the risk of taking up a bit more space, this would reduce the amount of conflicts to resolve.

toasted-nutbread: I would consider that outside of the scope of this change, but that sounds like a good idea. Particularly because, long-term, the number of global definitions should eventually be reduced. For example:

```js
/* global
 * apiScreenshotGet
 * apiForward
 * apiAudioGetUri
 */
```

Will likely eventually become:

```js
/* global
 * api
 */
```

siikamiika:
The URL --> URI change got me curious, what is `tts:?text=%E3%81%82%E3%81%84%E3%81%86%E3%81%88%E3%81%8A` in terms of URIs? At least the other default ones are all URLs. I guess you could add a custom sound for failing a lookup using a data URI, though.

ext/bg/js/backend.js
```diff
@@ -515,7 +515,7 @@ class Backend {
 
     async _onApiAudioGetUrl({definition, source, optionsContext}) {
         const options = this.getOptions(optionsContext);
-        return await audioGetUrl(definition, source, options);
+        return await this.audioUriBuilder.getUri(source, definition, options);
```
siikamiika: I think either `AudioUriBuilder.getUri` should be changed to the `definition, source, options` or the other way around, whichever is more desirable, to avoid confusion

toasted-nutbread: I'll update `source` to be the first argument consistently, since there's precedent for that in multiple places. There are a few places where this parameter is named either `source` or `mode`, and I'll try to settle on only one of those also.
https://github.com/FooSoft/yomichan/blob/0cd1f8f20a1b632847e14b280849e4af1b6dd9ad/ext/bg/js/translator.js#L144-L157
https://github.com/FooSoft/yomichan/blob/0cd1f8f20a1b632847e14b280849e4af1b6dd9ad/ext/mixed/js/display.js#L455-L477

toasted-nutbread:
> The URL --> URI change got me curious, what is `tts:?text=%E3%81%82%E3%81%84%E3%81%86%E3%81%88%E3%81%8A` in terms of URIs? At least the other default ones are all URLs. I guess you could add a custom sound for failing a lookup using a data URI, though.

It's made up, added in order to allow the same API to be used without having to make the return value more complicated than a string.

ext/bg/js/backend.js
siikamiika: Ok, I guess it can be done in another PR later too if the amount of changes needed would otherwise explode. The variable names are also becoming more relevant when using object destructuring as I noticed in https://github.com/FooSoft/yomichan/commit/0112dbab33ab214f9e1dc930558833956d4ad1c4.

toasted-nutbread: I'll take care of it in this PR, just haven't gotten around to it yet.

toasted-nutbread: Actually, it's considerably more simple to do the opposite of what I said. So in the interest of minimal changes, and hopefully minimal breakage, I went that route. 0cbf427ab50061b48c9027e63e9ee8a209946d37

siikamiika: Looks good, tested that it still works

PR #405, opened 2020-03-09
start popup depth from 1 on search page

siikamiika:
This was originally implemented in https://github.com/FooSoft/yomichan/commit/16521fec935845828b7e08baf1668657bb5a7cef, but broken somewhere in between.

### DIFF
--- a/ext/fg/js/frontend-initialize.js
+++ b/ext/fg/js/frontend-initialize.js
@@ -31,7 +31,7 @@ async function main() {
         const popupHost = new PopupProxyHost();
         await popupHost.prepare();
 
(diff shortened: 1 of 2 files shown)


siikamiika:
I guess this can be a bit confusing because the `else` in `frontend-initialize.js` is only run when `proxy` is false, and that comes from `search-frontend.js` only, but not sure what would be the best way to go about this.

toasted-nutbread:
The manual frontend init sequence is a bit hacky in general due to some timing issues related to when the injected scripts run. What you wrote looks fine to me unless it doesn't run for some reason.

siikamiika:
I added another check because this can be prone to programmer errors

Issue #335, opened 2020-01-26
Make settings element IDs match their internal name


siikamiika:
The amount of options is quite large, so I think we need a convention for naming their `id` attributes. It would also enable automatically reading options from HTML with the help of a data attribute or two, but that's not the scope of this issue.

I think a naming scheme such as the following would make sense:
```javascript
options.scanning.enableSearchTags = $('#scanning.enable-search-tags').prop('checked');
```

To make it even easier to parse, it could be in `camelCase` just like in the object.

#296

toasted-nutbread:
I already have something similar to this started, but the branch is out of date and will need some rebasing/refactoring at some point.

https://github.com/toasted-nutbread/yomichan/commits/settings-improvements

The branch has the original code for the JSON schemas which has already been merged, and then further builds upon it in a few ways:

* Commit "[Add data-option-target descriptors](https://github.com/toasted-nutbread/yomichan/commit/53859da069e37b9a9b4bab92be530e767f905469)" effectively does what you're talking about.
* Commit "[Update implementation of apiOptionsSet](https://github.com/toasted-nutbread/yomichan/commit/d871e020d30201f7aa1f08326f1ae8028c2d84e2)" extends the capabilities of `apiOptionsSet` so it can add/delete/etc more effectively. A longer term goal would be to make all options mutations performed via `apiOptionsSet`.
* Commit "[Create class to observe settings based on HTML](https://github.com/toasted-nutbread/yomichan/commit/d0afa69eca0c53d1c7da450735b3757bbd2c4722)" adds a class which observers DOM mutations on the settings page and automatically assigns settings observers to them using `apiOptionsSet`. This eliminates the need to write code.

siikamiika:
Nice! I'll leave this issue open to remind about the existing branch.

toasted-nutbread:
This issue has been addressed by #542 and subsequent settings refactoring PRs.

PR #406, opened 2020-03-09
Object property accessor

toasted-nutbread:
This class is intended to be used while implementing #335. This basically redoing part of the commit "[Update implementation of apiOptionsSet](https://github.com/toasted-nutbread/yomichan/commit/d871e020d30201f7aa1f08326f1ae8028c2d84e2#diff-dbd09b161e2e61a02b016ee5e11a5cb3)" that I mentioned in a comment on that issue. The end goal is to assign settings [like this](https://github.com/toasted-nutbread/yomichan/commit/d871e020d30201f7aa1f08326f1ae8028c2d84e2#diff-ae1c71fb13eb3a96c1cce3899144a9dcL78-R82) and enforce schema validation.

### DIFF
--- a/ext/mixed/js/object-property-accessor.js
+++ b/ext/mixed/js/object-property-accessor.js
@@ -89,32 +89,32 @@ class ObjectPropertyAccessor {
 
     static getPathArray(pathString) {
         const pathArray = [];
-        let state = 0;
+        let state = 'empty';
         let quote = 0;
         let value = '';
         let escaped = false;
         for (const c of pathString) {
             const v = c.codePointAt(0);
             switch (state) {
-                case 0: // Empty
-                case 1: // Expecting identifier start
+                case 'empty': // Empty
+                case 'id-start': // Expecting identifier start
                     if (v === 0x5b) { // '['
-                        if (state === 1) {
+                        if (state === 'id-start') {
                             throw new Error(`Unexpected character: ${c}`);
                         }
-                        state = 3;
+                        state = 'open-bracket';
                     } else if (
                         (v >= 0x41 && v <= 0x5a) || // ['A', 'Z']
                         (v >= 0x61 && v <= 0x7a) || // ['a', 'z']
                         v === 0x5f // '_'
                     ) {
-                        state = 2;
+                        state = 'id';
                         value += c;
                     } else {
                         throw new Error(`Unexpected character: ${c}`);
                     }
                     break;
-                case 2: // Identifier
+                case 'id': // Identifier
                     if (
                         (v >= 0x41 && v <= 0x5a) || // ['A', 'Z']
                         (v >= 0x61 && v <= 0x7a) || // ['a', 'z']
@@ -125,27 +125,27 @@ class ObjectPropertyAccessor {
                     } else if (v === 0x5b) { // '['
                         pathArray.push(value);
                         value = '';
-                        state = 3;
+                        state = 'open-bracket';
                     } else if (v === 0x2e) { // '.'
                         pathArray.push(value);
                         value = '';
-                        state = 1;
+                        state = 'id-start';
                     } else {
                         throw new Error(`Unexpected character: ${c}`);
                     }
                     break;
-                case 3: // Open bracket
+                case 'open-bracket': // Open bracket
                     if (v === 0x22 || v === 0x27) { // '"' or '\''
                         quote = v;
-                        state = 4;
+                        state = 'string';
                     } else if (v >= 0x30 && v <= 0x39) { // ['0', '9']
-                        state = 5;
+                        state = 'number';
                         value += c;
                     } else {
                         throw new Error(`Unexpected character: ${c}`);
                     }
                     break;
-                case 4: // Quoted string
+                case 'string': // Quoted string
                     if (escaped) {
                         value += c;
                         escaped = false;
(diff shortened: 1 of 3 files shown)


ext/mixed/js/object-property-accessor.js
```diff
@@ -0,0 +1,244 @@
...
+        this._setter = (typeof setter === 'function' ? setter : null);
+    }
+
+    getProperty(pathArray, pathLength) {
+        let target = this._target;
+        const ii = typeof pathLength === 'number' ? Math.min(pathArray.length, pathLength) : pathArray.length;
+        for (let i = 0; i < ii; ++i) {
+            const key = pathArray[i];
+            if (!ObjectPropertyAccessor.hasProperty(target, key)) {
+                throw new Error(`Invalid path: ${this.getPathString(pathArray.slice(0, i + 1))}`);
```
siikamiika: The error throws an error because `getPathString` is static. Same on line 49

toasted-nutbread: Fixed and updated tests to validate the error message

ext/mixed/js/object-property-accessor.js
```diff
@@ -0,0 +1,244 @@
...
+        const pathArray = [];
+        let state = 0;
+        let quote = 0;
+        let value = '';
+        let escaped = false;
+        for (const c of pathString) {
+            const v = c.codePointAt(0);
+            switch (state) {
+                case 0: // Empty
+                case 1: // Expecting identifier start
```
siikamiika: Do you think the states could be class variables instead? It would require putting them at bottom, but variable names are easier to read than plain integers especially when the state is changed from different places.

toasted-nutbread: So for class variables, the recommendation is to use strings instead of "enum" constants, as per https://github.com/FooSoft/yomichan/pull/157#discussion_r257526239. However, I chose to not do that in this case because there would be added overhead for string comparisons in the `switch`, which would occur for every character of the string.

IMO since the state integers are completely self-contained within the function, and the cases are commented, this should be sufficiently explanatory and maximally efficient. I could also comment the places where `state` is assigned with the same comment if you think that would be helpful.

siikamiika: Ok, after reading the earlier comments I understand the reason why class enums don't work so well in JS. Another option worth considering would be creating int-backed local constants for the states. Also, did you test how much it would affect performance to use strings for states?

siikamiika: > I could also comment the places where `state` is assigned with the same comment if you think that would be helpful.

I think that could be too difficult to maintain because the code would work just fine with incomplete or incorrect comments.

toasted-nutbread: > Another option worth considering would be creating int-backed local constants for the states. Also, did you test how much it would affect performance to use strings for states?

So I have just tested both of these using https://jsbench.me/ on Firefox, Chrome, and Edge, and these are the results:
* On Chrome, using direct integers, strings, or integer constants are all equally fast. v8 may have a good optimization for this which is causing the comparable speeds.
* On Firefox, direct integers are the fastest, strings are about 8% slower, and integer constants are about 4% slower.
* On Edge, direct integers are the fastest, strings are maybe 10% slower, and integer constants are roughly the same as direct integers if the constants are declared outside the function body.
* Chrome speed > Firefox speed > Edge speed.

So based on this, I could probably just switch to using strings. I was expecting the speed loss of string comparisons to be more significant.

siikamiika: Thanks a lot for testing! Did you test using this function or just comparisons? I got similar results testing a for loop with simple `===` checks, but realized that browsers might optimize those differently.

toasted-nutbread: https://gist.github.com/toasted-nutbread/366a15c1b4e93ddf656f44e184d439d2

Switch-case can be optimized better than `if (...) else if (...) else if (...) ...` when comparing the same var, and switch-case uses strict equality.

toasted-nutbread: Update to strings in 4b699a6b46869d5766f331d863aea38374ece50a

siikamiika:
We won't probably ever have a case when the settings object has non-ascii keys, but this would be valid in JS:
```javascript
const obj = {
  あ: {
    亜: 'a'
  }
};

obj.あ.亜;
```

It won't work here:
```javascript
ObjectPropertyAccessor.getPathArray('a.あ.亜');
// Uncaught Error: Unexpected character: あ
```

This will, though:
```javascript
ObjectPropertyAccessor.getPathArray('a["あ"]["亜"]');
// ["a", "あ", "亜"]
```

If there's no reason to limit the allowed characters, it could be changed, but I don't see this as a big issue because a workaround exists.

toasted-nutbread:
> We won't probably ever have a case when the settings object has non-ascii keys

There actually is: the keys of `options.dictionaries` are based on the imported dictionary names, which can be in Japanese.

However, the shorthand is not meant to be strictly compliant with how Javascript's syntax, since it's more simple to just support the standard `a-zA-Z0-9_`. And as you mentioned, the intended method is to use the accessor format for these cases.

siikamiika:
Oh, I imagined it would be an array with a `name` property or something.

toasted-nutbread:
Not as it is currently. That would probably be better, but it's been this way for quite a long time. I've thought about changing it a few times; however, I've not got around to it since it's a low priority.

```json
"dictionaries": {
  "JMdict (English)": {
    "priority": 0,
    "enabled": true,
    "allowSecondarySearches": false
  },
  "JMnedict": {
    "priority": 0,
    "enabled": false,
    "allowSecondarySearches": false
  },
  "KANJIDIC (English)": {
    "priority": 0,
    "enabled": true,
    "allowSecondarySearches": false
  },
  "KireiCake": {
    "priority": 0,
    "enabled": false,
    "allowSecondarySearches": false
  },
  "研究社　新和英大辞典　第５版": {
    "priority": 0,
    "enabled": false,
    "allowSecondarySearches": false
  },
  "Innocent Corpus": {
    "priority": 0,
    "enabled": true,
    "allowSecondarySearches": false
  },
  "明鏡国語辞典": {
    "priority": 0,
    "enabled": false,
    "allowSecondarySearches": false
  },
  "Anime & Jdrama Freq:": {
    "priority": 0,
    "enabled": true,
    "allowSecondarySearches": false
  },
  "三省堂　スーパー大辞林": {
    "priority": 0,
    "enabled": false,
    "allowSecondarySearches": false
  },
  "Kanjium Pitch Accents": {
    "priority": 0,
    "enabled": true,
    "allowSecondarySearches": false
  }
}
```

PR #408, opened 2020-03-10
Replace charCodeAt and fromCharCode with codePointAt and fromCodePoint

### LINKED ISSUE #331: Switch charCodeAt to codePointAt
Originally discussed in #314. Creating a new issue so it's harder to forget.

toasted-nutbread:
Resolves #331.

There are still some places in the codebase where UTF-16 character code iteration is still used. Some of these should be fine, while others may potentially have issues with the rare kanji.

For example:
https://github.com/FooSoft/yomichan/blob/7541517d8084b50e054393f3b4fa6d4630ad012e/ext/bg/js/search-query-parser.js#L158-L166

document.js could also have some issues, but I won't touch those for now due to #286.

### DIFF
--- a/ext/bg/js/handlebars.js
+++ b/ext/bg/js/handlebars.js
@@ -16,7 +16,7 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
 
-/*global jpIsCharCodeKanji, jpDistributeFurigana, Handlebars*/
+/*global jpIsCodePointKanji, jpDistributeFurigana, Handlebars*/
 
 function handlebarsEscape(text) {
     return Handlebars.Utils.escapeExpression(text);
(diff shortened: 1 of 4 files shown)


siikamiika:
These changes look good. We have to be extra careful when changing the functions that pass around length (`Translator.findTerms`, `TextSource.setEndOffset` etc.) because `Range.setStart/setEnd` might expect the string length and not the amount of code points.

toasted-nutbread:
> These changes look good. We have to be extra careful when changing the functions that pass around length (`Translator.findTerms`, `TextSource.setEndOffset` etc.) because `Range.setStart/setEnd` might expect the string length and not the amount of code points.

Yeah, I noticed that care needs to be taken in particular with the source mappings, since those are UTF-16 length based. The functions I changed which used a source map should be fine though.

PR #409, opened 2020-03-11
Update global declaration format

toasted-nutbread:
Resolves https://github.com/FooSoft/yomichan/pull/401#discussion_r389379783. The globals are now required to be listed with one entry per line in alphabetical order. The new lint script can also be run with the argument `--fix` to automatically fix formatting issues.


siikamiika:
Thank you! I didn't expect even a tool that does this automatically for you but I guess the amount of files was quite large :smile:

Issue #403, opened 2020-03-08
Technical documentation


siikamiika:
There has already been some comments and discussion about this such as https://github.com/FooSoft/yomichan/issues/322, but I'll open it as a new issue for better tracking.

I think the project is becoming so large that it would help new developers and our future selves if we documented the core components and their relationships so that it can be referenced while reading the source code and discussing the addition of new features.

The structure is already pretty good and consistent, but there are many hops between moving the mouse cursor and getting the results to display, not to mention in a second popup created by scanning the first.

This is somewhat related to https://github.com/FooSoft/yomichan/issues/356, as it would help to conclude when it's done. I think there could still be some work to do on the `Popup*` classes because they don't have a prepare method.

I don't have a strong opinion or even experience of choosing an appropriate tool for doing this, but I think it should be something that can indicate the entry points and flow of messages when some core feature events such as text scanning happen. In my opinion the code itself is good documentation for the internal functionality of the components, but some of them could maybe have a general comment about what they do.

toasted-nutbread:
My hopes are that a lot of documentation can be implicit by having the code be well structured and organized. For example, converting certain groups of functions to classes or module-like objects (#372) and marking internals as private (#258). In my opinion, this will go a long way, particularly for someone who is code savvy and knows how to scan through a codebase.

My intent with the JSON schemas has been that they are self-documenting for anyone who knows how to read them. This is why I have added the `description` fields which are otherwise not used in the code. I realize that it can't be assumed that everyone will already understand the schema format (i.e. less technical users who still understand basic JSON), but it can also help us to guide users correctly if it comes up (#382).

For high-level documentation on components, their relationships, design practices, etc., the wiki page is an option: https://github.com/FooSoft/yomichan/wiki. An alternative is creating a "docs" folder in the root project containing markdown files.

For more internal comments, [doc comments](https://jsdoc.app/about-getting-started.html) are certainly an option. I think there are tools to automatically compile doc comments to API HTML documentation pages, but personally I am usually satisfied to just read the comments inline in the code when I need them, since it's usually part of the development workflow anyway to look at a function's declaration.

For internal comments, usually private members aren't documented. Doc comments could be added to existing components on an on-demand basis, or whenever it's otherwise convenient to have additional info. For new classes, maybe minimally just adding a doc comment about what the class does could be helpful.

FooSoft:
[...]

PR #411, opened 2020-03-13
apply all options on profile change

siikamiika:
Extracted from https://github.com/FooSoft/yomichan/pull/389. I think these can be useful while deciding what's the best way to quickly change profiles and how to display it.

Was there a reason why `Frontend` was initialized with options from `Popup`? I removed it to make the code simpler, but it can make the popup appear a bit slower on the first scan.

### DIFF
--- a/ext/bg/js/search-query-parser.js
+++ b/ext/bg/js/search-query-parser.js
@@ -28,11 +28,10 @@
 
 class QueryParser extends TextScanner {
     constructor(search) {
-        super(document.querySelector('#query-parser-content'), [], [], []);
+        super(document.querySelector('#query-parser-content'), [], []);
         this.search = search;
 
         this.parseResults = [];
-        this.selectedParser = null;
 
         this.queryParser = document.querySelector('#query-parser-content');
         this.queryParserSelect = document.querySelector('#query-parser-select-container');
@@ -79,9 +78,7 @@ class QueryParser extends TextScanner {
 
     onParserChange(e) {
         const selectedParser = e.target.value;
-        this.selectedParser = selectedParser;
         apiOptionsSet({parsing: {selectedParser}}, this.search.getOptionsContext());
-        this.renderParseResult();
     }
 
     getMouseEventListeners() {
@@ -112,19 +109,16 @@ class QueryParser extends TextScanner {
 
     refreshSelectedParser() {
         if (this.parseResults.length > 0) {
-            if (this.selectedParser === null) {
-                this.selectedParser = this.search.options.parsing.selectedParser;
-            }
-            if (this.selectedParser === null || !this.getParseResult()) {
+            if (!this.getParseResult()) {
                 const selectedParser = this.parseResults[0].id;
-                this.selectedParser = selectedParser;
                 apiOptionsSet({parsing: {selectedParser}}, this.search.getOptionsContext());
             }
         }
     }
 
     getParseResult() {
-        return this.parseResults.find((r) => r.id === this.selectedParser);
+        const {selectedParser} = this.options.parsing;
+        return this.parseResults.find((r) => r.id === selectedParser);
     }
 
     async setText(text) {
@@ -176,7 +170,8 @@ class QueryParser extends TextScanner {
     renderParserSelect() {
         this.queryParserSelect.textContent = '';
         if (this.parseResults.length > 1) {
-            const select = this.queryParserGenerator.createParserSelect(this.parseResults, this.selectedParser);
+            const {selectedParser} = this.options.parsing;
+            const select = this.queryParserGenerator.createParserSelect(this.parseResults, selectedParser);
             select.addEventListener('change', this.onParserChange.bind(this));
             this.queryParserSelect.appendChild(select);
         }
--- a/ext/bg/js/search-query-parser.js
+++ b/ext/bg/js/search-query-parser.js
@@ -27,9 +27,12 @@
  */
 
 class QueryParser extends TextScanner {
-    constructor(search) {
+    constructor({getOptionsContext, setContent, setSpinnerVisible}) {
         super(document.querySelector('#query-parser-content'), [], []);
-        this.search = search;
+
+        this.getOptionsContext = getOptionsContext;
+        this.setContent = setContent;
+        this.setSpinnerVisible = setSpinnerVisible;
 
         this.parseResults = [];
 
@@ -55,18 +58,18 @@ class QueryParser extends TextScanner {
     async onSearchSource(textSource, cause) {
         if (textSource === null) { return null; }
 
-        this.setTextSourceScanLength(textSource, this.search.options.scanning.length);
+        this.setTextSourceScanLength(textSource, this.options.scanning.length);
         const searchText = textSource.text();
         if (searchText.length === 0) { return; }
 
-        const {definitions, length} = await apiTermsFind(searchText, {}, this.search.getOptionsContext());
+        const {definitions, length} = await apiTermsFind(searchText, {}, this.getOptionsContext());
         if (definitions.length === 0) { return null; }
 
-        const sentence = docSentenceExtract(textSource, this.search.options.anki.sentenceExt);
+        const sentence = docSentenceExtract(textSource, this.options.anki.sentenceExt);
 
         textSource.setEndOffset(length);
 
-        this.search.setContent('terms', {definitions, context: {
+        this.setContent('terms', {definitions, context: {
             focus: false,
             disableHistory: cause === 'mouse',
             sentence,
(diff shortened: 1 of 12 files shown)


ext/bg/js/search-query-parser.js
```diff
@@ -99,11 +102,23 @@ class QueryParser extends TextScanner {
         this.queryParser.dataset.termSpacing = `${options.parsing.termSpacing}`;
     }
 
+    getOptionsContext() {
```
toasted-nutbread: These declarations are redundant since the properties are assigned in the constructor. Adding a doc comment (#403) which describes the expected arguments for the functions could be helpful. If you want to make sure that the programmer has assigned proper values, better to verify that in the constructor rather than later on. I.e.
```js
if (typeof getOptionsContext !== 'function') { throw new Error('Invalid value for getOptionsContext'); }
```

siikamiika: You're right, I think I'll remove the stubs and don't add type checking for now because the class is only used in one place.

ext/fg/js/popup.js
```diff
@@ -207,20 +207,17 @@ class Popup {
             const parentFrameId = (typeof this._frameId === 'number' ? this._frameId : null);
             this._container.setAttribute('src', chrome.runtime.getURL('/fg/float.html'));
             this._container.addEventListener('load', () => {
-                const uniqueId = yomichan.generateId(32);
-                Popup._listenForDisplayPrepareCompleted(uniqueId, resolve);
+                this._listenForDisplayPrepareCompleted(resolve);
```
toasted-nutbread: There is a timing issue now that uniqueId has been removed and only the popup `id` is used. `id` is unique within a frame, but not unique within a tab, and the `'popupPrepareCompleted'` message is sent to all frames on a tab. `uniqueId` was used to provide uniqueness within a tab.

If two popups are opened at the same time in different frames on the same tab, the body of `_listenForDisplayPrepareCompleted.runtimeMessageCallback` can be triggered for the wrong frame's popup. In practice this is extremely rare to the point of being improbable, and the negative effects are even less likely to show up due to how quickly the messages are processed, but the issue is technically still there.

The following example demonstrates that IDs are not unique within a tab. Note that `prepared: 2` is printed, which indicates the timing issue _could_ occur.

```js
_listenForDisplayPrepareCompleted(resolve) {
    const runtimeMessageCallback = ({action, params}, sender, callback) => {
        setTimeout(() => { // added
        if (
            action === 'popupPrepareCompleted' &&
            isObject(params) &&
            params.targetPopupId === this._id
        ) {
            this._prepareCount = (this._prepareCount || 0) + 1; // added
            console.log('prepared: ' + this._prepareCount); // added
            chrome.runtime.onMessage.removeListener(runtimeMessageCallback);
            callback();
            resolve();
            return false;
        }
        }, 2000); // added
    };
    chrome.runtime.onMessage.addListener(runtimeMessageCallback);
}
/*
Prints:
prepared: 1
prepared: 2
prepared: 1
*/
```

Compared with the old method using `uniqueId`:

```js
static _listenForDisplayPrepareCompleted(uniqueId, resolve) {
    const runtimeMessageCallback = ({action, params}, sender, callback) => {
        setTimeout(() => { // added
        if (
            action === 'popupPrepareCompleted' &&
            typeof params === 'object' &&
            params !== null &&
            params.uniqueId === uniqueId
        ) {
            this._prepareCount = (this._prepareCount || 0) + 1; // added
            console.log('prepared: ' + this._prepareCount); // added
            chrome.runtime.onMessage.removeListener(runtimeMessageCallback);
            callback();
            resolve();
            return false;
        }
        }, 2000); // added
    };
    chrome.runtime.onMessage.addListener(runtimeMessageCallback);
}
/*
Prints:
prepared: 1
prepared: 1
*/
```

ext/fg/js/frontend.js
```diff
@@ -88,6 +87,9 @@ class Frontend extends TextScanner {
     }
 
     onRuntimeMessage({action, params}, sender, callback) {
+        const {targetPopupId} = params || {};
+        if (targetPopupId !== 'all' && targetPopupId !== this.popup.id) { return; }
```
toasted-nutbread: May be better to avoid the `'all'` special constant and check if `typeof targetPopupId !== undefined` instead. Reason: I've experimented with popup IDs being strings before, and this would prevent using 'all' as an ID, which although unlikely, adds a special handling case.

toasted-nutbread: This can also run into the same `id` uniqueness issue that I mentioned for `_listenForDisplayPrepareCompleted`.

toasted-nutbread:
> Was there a reason why Frontend was initialized with options from Popup? I removed it to make the code simpler, but it can make the popup appear a bit slower on the first scan.

Not sure what part exactly this is referring to; can you ref some lines of code? Nothing is standing out as a strong reason for something like that as far as I can currently tell.

toasted-nutbread:
This is kind of unrelated, but since this code touched it, I'm wondering if `QueryParser.parseText` should be moved into the backend. IMO checking whether or not to do parsing, particularly when checking mecab, seems like an implementation detail which isn't necessary for the "frontend" to know.

If this is the case, the data returned might have to be formatted differently in order to make it more generalized, but it wouldn't be a big difference. For example:

```js
// something like this
results.push({
    source: 'scanning-parser',
    content: results
});
results.push({
    source: 'mecab',
    dictionary: mecabDictName,
    content: mecabDictResults
});

// instead of the following
// the name/id are for presentation/options purposes
// parsedText may also be a misleading name since the data is an array of objects and not a string
results.push({
    name: 'Scanning parser',
    id: 'scan',
    parsedText: await apiTextParse(text, this.getOptionsContext())
});
results.push({
    name: `MeCab: ${mecabDictName}`,
    id: `mecab-${mecabDictName}`,
    parsedText: mecabDictResults
});
```

ext/fg/js/popup.js
siikamiika: Thanks for noticing this, I was under the impression that iframe popup IDs are also unique within the root page popup proxy host because the last time I consciously scanned iframes was when doing https://github.com/FooSoft/yomichan/pull/364. It actually doesn't even assign an ID for the iframe popups and guesses the position instead.

The simplest solution would be to generate an unique ID for each popup and use it instead of the sequence, but I'd like to centralize tab communication at some point (not this PR) so that the root page popup proxy is aware of popups within iframes too.

ext/fg/js/frontend.js
siikamiika: I can change the meaning of a missing `targetPopupId` to the same as `'all'` currently. There aren't any cases when `'all'` cannot also be sent to other frames, but if such functionality is needed, another frame specific ID can be generated.

siikamiika:
> Not sure what part exactly this is referring to; can you ref some lines of code? Nothing is standing out as a strong reason for something like that as far as I can currently tell.

https://github.com/FooSoft/yomichan/pull/411/files#diff-f156747a57332264c5853023f68d1b83L214
https://github.com/FooSoft/yomichan/pull/411/files#diff-d5a318d27e03a2960299ccc82727e376L154
https://github.com/FooSoft/yomichan/pull/411/files#diff-d5a318d27e03a2960299ccc82727e376L362

siikamiika:
> This is kind of unrelated, but since this code touched it, I'm wondering if `QueryParser.parseText` should be moved into the backend. IMO checking whether or not to do parsing, particularly when checking mecab, seems like an implementation detail which isn't necessary for the "frontend" to know.

Good idea, it would reduce the amount of API functions by 1. I'll do it separately because it might have some unexpected issues that need larger changes.

toasted-nutbread:
> > Not sure what part exactly this is referring to; can you ref some lines of code? Nothing is standing out as a strong reason for something like that as far as I can currently tell.
> 
> https://github.com/FooSoft/yomichan/pull/411/files#diff-f156747a57332264c5853023f68d1b83L214
> https://github.com/FooSoft/yomichan/pull/411/files#diff-d5a318d27e03a2960299ccc82727e376L154
> https://github.com/FooSoft/yomichan/pull/411/files#diff-d5a318d27e03a2960299ccc82727e376L362

Okay yeah, your changes look fine. I think it was just historically that way and I left it when I have done modifications.

ext/fg/js/popup.js
siikamiika: I read the code a bit, and it could be possible to drop `messageToken` as well. I remember I suggested it to come from `Backend`, but currently it doesn't seem to be used in a way that would require it to be the same among different tabs.

siikamiika: ...except in the case of `prepare`. I'll leave it there :smile:

Issue #372, opened 2020-02-20
Module-like design revisited


toasted-nutbread:
One thing that became more apparent as I added the unit tests in #371 is that it can be difficult to set up all of the global dependencies.

https://github.com/FooSoft/yomichan/blob/1e573f36c416b3955b2e679c8f3ce7ffa91bf504/test/test-database.js#L13-L22

While it was fairly manageable for database.js, something like translator.js would be more difficult:

https://github.com/FooSoft/yomichan/blob/1e573f36c416b3955b2e679c8f3ce7ffa91bf504/ext/bg/js/translator.js#L19-L24

This could be simplified if add all of the `jp*`, `dict*`, etc. definitions were members of an object (e.g. `jp`) rather than globally defined. That way, only the `jp` or `dict` objects would have to be manually assigned to the global scope, which is much more manageable. If we were able to use `require` and modules, this is how it would have to be done anyways (`module.exports`).

This is something I previously brought up in #258, but it was decided to not do it, partially because there not being a strong reason to do so. I think unit tests may provide a stronger motivation for using this pattern. The alternative would seem to be transpiling, which is not really a desirable change.

siikamiika:
I'm hopeful for getting module support to content scripts one day. Until then I think something like what's done with the `DOM` class could be an alternative: https://github.com/FooSoft/yomichan/blob/174d9e7429a0f7a45538e56a594ce627239d80c3/ext/mixed/js/dom.js#L20-L66
https://github.com/FooSoft/yomichan/blob/174d9e7429a0f7a45538e56a594ce627239d80c3/ext/mixed/js/text-scanner.js#L52

Every usage of `DOM.isMouseButtonDown` can be found from the codebase by name because the class isn't instantiated anywhere and only used statically.

This, provided that the functions are global in nature such the API functions, `jp*` functions or helpers like `isObject`. To me many of the `dict*` functions look like they could be a part of some class that deals with dictionary data such as `Translator` or another class that's used by it.

toasted-nutbread:
> Every usage of `DOM.isMouseButtonDown` can be found from the codebase by name because the class isn't instantiated anywhere and only used statically.

Agreed, and it's also possible to do a singleton approach, like how the Yomichan class is designed.

https://github.com/FooSoft/yomichan/blob/0f46e3a093e7f0c07ad310d8c17e2582bdfd2741/ext/mixed/js/core.js#L267-L322

[...]

Issue #281, opened 2019-11-14
furigana bug


Geniusssmit:
just seen that word
![image](https://user-images.githubusercontent.com/24941285/68890928-f6e13380-0730-11ea-8a45-b7b3e5216876.png)

siikamiika:
Duplicate of #94.

In short, it's not possible to get the furigana right without some level of guessing using shorter words in the dictionary (not perfect, see https://github.com/FooSoft/yomichan/issues/264#issuecomment-545143944), which would add a performance penalty that goes against the goal of making Yomichan faster.

It would be quite easy to check if there are many possible ways to distribute furigana for a word and fall back to showing the whole reading including kana already in the word, so that's something I might look into in the near future.

Issue #264, opened 2019-10-22
{furigana-plain} has improper spacing for Anki


Rapptz:
The Anki furigana support both in AnkiMobile and desktop Anki (I assume iOS as well) require spaces between each segment to properly render the furigana.

For example, a word like お祝い will spit out お祝[いわ]い for the plain furigana. However, this renders incorrectly: 

![Incorrect Anki Furigana Render](https://user-images.githubusercontent.com/1695103/67284272-4e2e1400-f4a3-11e9-844d-6a085136b915.png)

To fix this requires a space between the お and the kanji: お 祝[いわ]い

![Correct Anki Furigana Render](https://user-images.githubusercontent.com/1695103/67284304-60a84d80-f4a3-11e9-8457-0670e36c6dd8.png)

FooSoft:
How would you do furigana for 美味しい or other あてじ? The problem is that the term and reading are tied together, and it would be tricky (albeit not impossible), to try to match up parts of the reading to kanji in the term.

FooSoft:
Ah, actually nevermind, thinking about this more, there was a code contribution many moons ago to make this work, so that just needs to be propagated to the Anki layer.

Rapptz:
> How would you do furigana for 美味しい or other あてじ?

美味しい should be 美味[おい]しい. 

As far as I know this space requirement happens when there's kana between the kanji. So something like 食べ物 should be 食[た]べ 物[もの] and 試し切り would be 試[ため]し 切[ぎ]り

siikamiika:
In this case all the required information is contained in お祝い and its reading おいわい and it's up to the implementation to use that information, but cases like 飼い犬・かいいぬ (https://github.com/FooSoft/yomichan/issues/94) are ambiguous unless you have definitive information about the possible readings for each kanji. Even then exceptions with あてじ such as the one @FooSoft mentioned can come and ruin the party, and at least misplace kana on top of the wrong kanji (for example when the あてじ has partial matches) if there's no other ambiguity.

Rapptz:
Apologies. There might be some misunderstanding here.

I'm not really suggesting to correct the current segmentation (which is where the bug regarding かいいぬ being badly segmented for 飼い犬 comes from). I'm talking about the current segmentation results (i.e. the result of `jpDistributeFurigana`) being incorrect for the furigana-plain handlebars. I can open a PR to fix this rather than the segmentation which can be tackled later.

PR #413, opened 2020-03-14
Convert Japanese utilities to a module-like style

toasted-nutbread:
Creating this PR now in order to determine a preferred style for module-like style that can also be used for https://github.com/FooSoft/yomichan/pull/385#discussion_r386722493. I'll probably create some unit tests for this also, but they are not added yet.

The diff for japanese.js makes it look like everything changed, but the main differences are some variable renames (remove jp* prefix), added indentation, and rearranging the declarations.

Related: #372

### DIFF
--- a/ext/bg/js/japanese.js
+++ b/ext/bg/js/japanese.js
@@ -231,7 +231,7 @@ const jp = (() => {
         }
     }
 
-    function convertNumericTofullWidth(text) {
+    function convertNumericToFullWidth(text) {
         let result = '';
         for (const char of text) {
             let c = char.codePointAt(0);
@@ -485,7 +485,7 @@ const jp = (() => {
         convertHiraganaToKatakana,
         convertToRomaji,
         convertReading,
-        convertNumericTofullWidth,
+        convertNumericToFullWidth,
         convertHalfWidthKanaToFullWidth,
         convertAlphabeticToKana,
         distributeFurigana,
--- a/ext/bg/js/japanese.js
+++ b/ext/bg/js/japanese.js
@@ -20,439 +20,461 @@
  * wanakana
  */
 
-const JP_HALFWIDTH_KATAKANA_MAPPING = new Map([
-    ['ｦ', 'ヲヺ-'],
-    ['ｧ', 'ァ--'],
-    ['ｨ', 'ィ--'],
-    ['ｩ', 'ゥ--'],
-    ['ｪ', 'ェ--'],
-    ['ｫ', 'ォ--'],
-    ['ｬ', 'ャ--'],
-    ['ｭ', 'ュ--'],
-    ['ｮ', 'ョ--'],
-    ['ｯ', 'ッ--'],
-    ['ｰ', 'ー--'],
-    ['ｱ', 'ア--'],
-    ['ｲ', 'イ--'],
-    ['ｳ', 'ウヴ-'],
-    ['ｴ', 'エ--'],
-    ['ｵ', 'オ--'],
-    ['ｶ', 'カガ-'],
-    ['ｷ', 'キギ-'],
-    ['ｸ', 'クグ-'],
-    ['ｹ', 'ケゲ-'],
-    ['ｺ', 'コゴ-'],
-    ['ｻ', 'サザ-'],
-    ['ｼ', 'シジ-'],
-    ['ｽ', 'スズ-'],
-    ['ｾ', 'セゼ-'],
-    ['ｿ', 'ソゾ-'],
-    ['ﾀ', 'タダ-'],
-    ['ﾁ', 'チヂ-'],
-    ['ﾂ', 'ツヅ-'],
-    ['ﾃ', 'テデ-'],
-    ['ﾄ', 'トド-'],
-    ['ﾅ', 'ナ--'],
-    ['ﾆ', 'ニ--'],
-    ['ﾇ', 'ヌ--'],
-    ['ﾈ', 'ネ--'],
-    ['ﾉ', 'ノ--'],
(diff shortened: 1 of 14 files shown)


toasted-nutbread:
Unit tests are now added, which include tests for #264, #281. Additional CJK ranges have also been added in 8d374a34bbfcc1c84c357b14e6cc6b72797147b2, which adds support for `𠮟 (0x20b9f)`, which was initially mentioned in https://github.com/FooSoft/yomichan/pull/314#issuecomment-577935339.

One thing that I noticed which is strange is that `wanakana.toHiragana` converts all text to lower case before the conversion, which causes strange results for some cases:

https://github.com/FooSoft/yomichan/blob/ca285ed2b7cdf5bc0edcadc09b2ceaf6a66cd1da/test/test-japanese.js#L236-L244

toasted-nutbread:
I also had some questions about the use of `convertReading` and its arguments.
1. `expressionFragment` and `readingFragment` could probably be renamed to just `expression` and `reading` for consistency with some of the other `jp*` conversion functions. I think the semantics of the values being "fragments" is only relevant in the `_onApiTextParse*` functions they are used in.
2. What are the expected types of `readingFragment`? It looks like it can be falsy based on the `'hiragana'` and `'katakana'` cases.
3. What are the cases when `readingFragment` is falsy?
4. What are the cases when `readingFragment` is falsy and `isStringEntirelyKana(expressionFragment)` is true?

I didn't look too deeply into exactly how it's used, but it looks like this may be more suitable as a utility function defined in backend.js, since the cases are based on the option `options.parsing.readingMode`, and the functionality looks to be specialized for a single use case. IMO this is similar to how a lot of the `dict` functions are really utilities for the backend or translator rather than general purpose functions.

https://github.com/FooSoft/yomichan/blob/ca285ed2b7cdf5bc0edcadc09b2ceaf6a66cd1da/ext/bg/js/japanese.js#L212-L232

siikamiika:
> * `expressionFragment` and `readingFragment` could probably be renamed to just `expression` and `reading` for consistency with some of the other `jp*` conversion functions. I think the semantics of the values being "fragments" is only relevant in the `_onApiTextParse*` functions they are used in.

Sure, that makes sense.

> * What are the expected types of `readingFragment`? It looks like it can be falsy based on the `'hiragana'` and `'katakana'` cases.

It can be string, null or undefined. Maybe an empty string should be used consistently, I don't know.

> * What are the cases when `readingFragment` is falsy?
> * What are the cases when `readingFragment` is falsy and `isStringEntirelyKana(expressionFragment)` is true?

`null` is explicitly added in `Backend._onApiTextParse*`. `jpDistributeFuriganaInflected` and `jpDistributeFurigana` leave it out sometimes, making it `undefined`. It's also an empty string when a dictionary doesn't have a reading.

When there's a gap in furigana or a word is kana-only, this will happen. It's also valid in the schema for a dictionary to have an empty string (falsy) as a reading while the expression is not entirely kana, and I've seen that a few times with J-J dictionaries.

If a string isn't found in Yomichan dictionaries (scanning parser) or MeCab dictionaries, or if it matches `u'[\s\u30fb]'` (Python regex). The MeCab properties can also be `null` arbitrarily depending on the source dictionary.

toasted-nutbread:
Thanks for the explanations. The reason I asked was that I was unsure what unit tests to provide for it.

toasted-nutbread:
More unit tests added in 780d23b749325da0a95aa9cc7898df19f2ac1b31. I will let you (@siikamiika) verify that these are what you expect.

https://github.com/FooSoft/yomichan/blob/780d23b749325da0a95aa9cc7898df19f2ac1b31/test/test-japanese.js#L173-L237

I have also moved common Japanese functions which don't use WanaKana into a mixed japanese.js file. This is also in preparation for #385, and this also fixes the display generator's parsing of kanji. Kanji from the unicode CJK unified ideograph extensions B-F.

ext/bg/js/japanese.js
```diff
@@ -469,12 +382,7 @@ const jp = (() => {
 
     // Exports
 
-    return {
-        isCodePointKanji,
-        isCodePointKana,
-        isCodePointJapanese,
-        isStringEntirelyKana,
-        isStringPartiallyJapanese,
+    Object.assign(jp, {
```
siikamiika: I wonder if this and the reusage of `mixed` `jp` functions are too much complexity just to keep some of the functions out of frontend. I can see a problem with `wanakana` being undefined though, and adding it to frontend html files just to make the unused functions usable seems pointless.

test/test-japanese.js
```diff
@@ -0,0 +1,414 @@
...
+        [['有り難う', void 0, 'katakana'], ''],
+        [['有り難う', void 0, 'romaji'], void 0],
+        [['有り難う', void 0, 'none'], null],
+        [['有り難う', void 0, 'default'], void 0]
+    ];
+
+    for (const [[expressionFragment, readingFragment, readingMode], expected] of data) {
+        assert.strictEqual(jp.convertReading(expressionFragment, readingFragment, readingMode), expected);
+    }
+}
```
siikamiika: Yeah, these make sense as far as the current implementation goes. I guess it just highlights that the types are currently handled sloppily and should be unified. I can take care of it when doing https://github.com/FooSoft/yomichan/issues/418

ext/bg/js/japanese.js
toasted-nutbread: Yeah, it's a bit ugly, but unfortunately as you mentioned in a comment below, there's no good way to do modules. And having the object name be just `jp` is convenient; moreso than something like `jpFrontend` and `jpCommon` or something like that.

Issue #402, opened 2020-03-08
Popup menu buttons UI/UX


toasted-nutbread:
Here is a quick mock-up of a potential design, with some icons as placeholders. This is a fairly minimal design that doesn't feel like it would significantly change users' workflows.

![test](https://user-images.githubusercontent.com/11037431/76156979-627c5180-60d0-11ea-870d-27bddb15c1b3.gif)

For comparison, if the buttons were placed on the top or left sides, the cursor would need to be used in a different location than the current design, which could impede muscle memory patterns, reduce available layout space, or just generally an undesirable change.

High-level overview of tasks involved in implementing something like this:
* Create general layout and design
* Some focus issues would need to be handled differently
* Add button functionality
* Resize functionality may need to be reconsidered and reimplemented since the handle overlaps the bottom icon
  * This could also bring resizing to Firefox
* Ensure the design works with the search page without feeling out of place
* Display options such as auto-hide, location, buttons, etc.

This is not necessarily a final design; this issue is intended to discuss needs and evolve the design as necessary.

Related: #292

siikamiika:
This kind of design seems great, it would help to solve many of the issues that arise when trying to add new features. It would also help with https://github.com/FooSoft/yomichan/issues/355, because making the header stay above input and parse results on the search page feels wrong. I'm really happy that there's someone who can sketch these kinds of things, because my own experience is limited to poking CSS until it works.

siikamiika:
I'm not sure if this is a separate issue, but I think it could be a problem that scrolling doesn't change the current entry but alt+up/down/scroll does. Many users could be just scrolling through the results until they find the one they wish to make a note of, and click the button that's at the top of the entry. While changing the current entry to the one currently displayed could be made automatic, it's still ambiguous when that change actually happens: when you're half way to the second entry, which one are you viewing?

toasted-nutbread:
> I'm not sure if this is a separate issue, but I think it could be a problem that scrolling doesn't change the current entry but alt+up/down/scroll does.

Yeah, I've thought about this a few times too, part of the issue stemming from the indicator being offscreen. A common indicator of the "current selection" is a small, colored sidebar:

![image](https://user-images.githubusercontent.com/11037431/76166413-bde02b00-6134-11ea-9496-f692e9d01a5e.png)

![image](https://user-images.githubusercontent.com/11037431/76166415-c5073900-6134-11ea-9422-569cdc1763fe.png)

[...]

Issue #398, opened 2020-03-07
Profile condition to match querySelector on the page using the profile


siikamiika:
While doing https://github.com/FooSoft/yomichan/pull/389 I got some ideas about improving the ways profiles can be chosen. One possibly controversial idea was the ability to choose a profile based on the language used on the page. Controversial because Yomichan is meant for Japanese only, but my use case is choosing another profile when the page is in Chinese.

To make the feature compatible with Yomichan's task, I decided to try the following approach: https://github.com/siikamiika/yomichan/commit/b23621d1cacf86021b562601a15022c44d8cbffe

It might look simple now, but I'm not sure yet how big changes it would require adding to the current profile matching framework. At least it would involve passing an asynchronous function to it.

I'm also hoping that this could be useful for some other use cases, such as choosing a specific profile when the page contains a `<video>` element.

PR #415, opened 2020-03-15
Anki marker document title

### LINKED ISSUE #414: Add {page-title} marker for use in card creation
I would find it useful to be able to include the current page title when adding cards to Anki. The title can help provide more context for where something came from the just the URL.

toasted-nutbread:
Resolves #414 with a new tag `{document-title}`. These changes can be used to easily add additional context data in the future.

### DIFF
--- a/ext/fg/js/float.js
+++ b/ext/fg/js/float.js
@@ -162,6 +162,33 @@ class DisplayFloat extends Display {
     setContentScale(scale) {
         document.body.style.fontSize = `${scale}em`;
     }
+
+    async getDocumentTitle() {
+        try {
+            const uniqueId = yomichan.generateId(16);
+
+            const promise = yomichan.getTemporaryListenerResult(
+                chrome.runtime.onMessage,
+                ({action, params}, {resolve}) => {
+                    if (
+                        action === 'documentInformationBroadcast' &&
+                        isObject(params) &&
+                        params.uniqueId === uniqueId &&
+                        params.frameId === 0
+                    ) {
+                        resolve(params);
+                    }
+                },
+                2000
+            );
+            apiForward('requestDocumentInformationBroadcast', {uniqueId});
+
+            const {title} = await promise;
+            return title;
+        } catch (e) {
+            return '';
+        }
+    }
 }
 
 DisplayFloat.instance = new DisplayFloat();
--- a/ext/mixed/js/display.js
+++ b/ext/mixed/js/display.js
@@ -908,10 +908,15 @@ class Display {
 
     async getDefinitionsAddable(definitions, modes) {
         try {
-            return await apiDefinitionsAddable(definitions, modes, this._getNoteContext(), this.getOptionsContext());
+            const context = await this._getNoteContext();
+            return await apiDefinitionsAddable(definitions, modes, context, this.getOptionsContext());
         } catch (e) {
             return [];
         }
+    }
+
+    async getDocumentTitle() {
+        return document.title;
     }
 
     static indexOf(nodeList, node) {
@@ -934,10 +939,11 @@ class Display {
         return (typeof key === 'string' ? (key.length === 1 ? key.toUpperCase() : key) : '');
     }
 
-    _getNoteContext() {
+    async _getNoteContext() {
+        const documentTitle = await this.getDocumentTitle();
         return {
             document: {
-                title: document.title
+                title: documentTitle
             }
         };
     }
--- a/ext/mixed/js/display.js
+++ b/ext/mixed/js/display.js
@@ -752,15 +752,15 @@ class Display {
         try {
             this.setSpinnerVisible(true);
 
-            const context = {};
+            const details = {};
             if (this.noteUsesScreenshot(mode)) {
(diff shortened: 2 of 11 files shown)


ext/mixed/js/display.js
```diff
@@ -931,6 +931,14 @@ class Display {
         return (typeof key === 'string' ? (key.length === 1 ? key.toUpperCase() : key) : '');
     }
 
+    _getNoteContext() {
+        return {
+            document: {
+                title: document.title
```
siikamiika: This is the document of the popup itself. Needs to query the title from root frame `Frontend`.

siikamiika:
`document-title` is missing from the markers in `bg/js/settings/anki.js`

ext/mixed/js/display.js
siikamiika: Something like https://github.com/siikamiika/yomichan/commit/b23621d1cacf86021b562601a15022c44d8cbffe in https://github.com/FooSoft/yomichan/issues/398 using `apiForward` with the root popup ID from the broadcast done in https://github.com/FooSoft/yomichan/pull/417

siikamiika: Also, ID is not needed if popup depth is checked before responding.

siikamiika: This brings up an issue with the iframe popups though, the wrong `Frontend` might respond if it has a popup proxy with the same ID as the root page.

siikamiika: One more comment, the current way to get document context from the parent frame seems to be through `window.frontendInitializationData` where the URL is passed through the `Popup` -> `DisplayFloat.prepare` (-> ... -> `PopupProxy`) chain. Not sure which way would be the best for this use case, but at least the document title seems more dynamic than the URL although both can be modified after the page has loaded.

toasted-nutbread: > This is the document of the popup itself. Needs to query the title from root frame `Frontend`.

I know I had forgotten something and that this felt too simple.

> Not sure which way would be the best for this use case, but at least the document title seems more dynamic than the URL although both can be modified after the page has loaded.

Yeah, it's probably better to get the title dynamically since it's not a time-critical operation. And I was also aware that the URL is possible to change, but chose not to address the issue at that time due to added complexity.

siikamiika: > And I was also aware that the URL is possible to change, but chose not to address the issue at that time due to added complexity.

It seems that additional permissions would be needed to monitor the URL change properly anyway: https://developer.chrome.com/extensions/tabs#event-onUpdated

Polling would work, but it's good to avoid it as much as possible. For clipboard it was the only choice. Caching is also an option, but depending on cache miss handling it can have problems with outdated data or periodical high latency.

toasted-nutbread: Updated in 696897f6fa431c0c24332c5dd2a2c12a07923c0e.

At some point in the future, it may be better to just pass the document information as part of the call to `setContent`, but for now, I want to avoid cluttering the `context` value that is already passed to that function. A lot of that code will likely be touched when addressing #402 anyways.

ext/fg/js/float.js
```diff
@@ -162,6 +162,33 @@ class DisplayFloat extends Display {
...
+            const uniqueId = yomichan.generateId(16);
+
+            const promise = yomichan.getTemporaryListenerResult(
+                chrome.runtime.onMessage,
+                ({action, params}, {resolve}) => {
+                    if (
+                        action === 'documentInformationBroadcast' &&
+                        isObject(params) &&
+                        params.uniqueId === uniqueId &&
+                        params.frameId === 0
```
siikamiika, line 177: Not sure if we should rely on `frameId` being 0 for the root frame unless it's in the spec. There's already enough context from Yomichan like in
https://github.com/FooSoft/yomichan/blob/3df78904cf734da208c6fd1b6ae1cd6612323148/ext/fg/js/frontend.js#L261-L265
(that `frameId` is used for PopupProxy parent frame ID)

siikamiika: Checking popup depth 0 on search page popup might be a problem, though, because it's started from 1 on purpose to account for the page itself.

toasted-nutbread: > Not sure if we should rely on frameId being 0 for the root frame unless it's in the spec.

https://developer.chrome.com/extensions/webNavigation#frame_ids

>  Frames within a tab can be identified by a frame ID. The frame ID of the main frame is always 0, the ID of child frames is a positive number. Once a document is constructed in a frame, its frame ID remains constant during the lifetime of the document. As of Chrome 49, this ID is also constant for the lifetime of the frame (across multiple navigations).

toasted-nutbread: Additionally, `if (!this.popup.isProxy() && this.popup.depth === 0) {` can have multiple matches when iframes are involved.

siikamiika: Right, it seems that the check is only valid when iframe popups are shown in root frame.

I got the impression that frame id 0 for root frame was an implementation detail from an earlier comment somewhere, but it seems quite reliable based on the Chrome document.

PR #417, opened 2020-03-19
Show iframe popups on root page

siikamiika:
Replaces https://github.com/FooSoft/yomichan/pull/364.
Fixes https://github.com/FooSoft/yomichan/issues/29.

https://github.com/FooSoft/yomichan/commit/765c1feec9b954c0d0cf79fa2146b0954466c987 does basically the same thing as https://github.com/toasted-nutbread/yomichan/tree/iframe-popups. The difference is that it's compatible with the ID generation in https://github.com/FooSoft/yomichan/pull/411 and doesn't expect the ID to be `'root'`. It should also handle the timing issue.

https://github.com/FooSoft/yomichan/commit/5c216a9416475116ecba191fa6f0e648de94bc83 replaces `frameIdPromise` with `frameId` (https://github.com/FooSoft/yomichan/issues/356).

https://github.com/FooSoft/yomichan/pull/417/commits/ea1da87041fad7eb04f2b70d0bacb26eeb4428bf is the magic, it basically passes an offset starting from 0 to the parent frame checking the offset of the origin frame until it reaches the root frame, and after that it's bounced back using `apiForward` with the accumulated offset and an ID token.

~The iframe lookup logic is duplicated in two places and could be refactored to a new class.~ (done in https://github.com/FooSoft/yomichan/pull/417/commits/b5c3d07ac906dc6905175c6821c193fe6921a899) I also noticed that the offset check adds some latency, but is probably needed for each check because the frames can move between scans. It would be too complex to monitor each iframe for resizes, but a compromise could be to cache the position for some time before checking again.

### DIFF
--- a/ext/fg/js/frame-offset-forwarder.js
+++ b/ext/fg/js/frame-offset-forwarder.js
@@ -22,15 +22,23 @@
 
 class FrameOffsetForwarder {
     constructor() {
-        this._forwardFrameOffset = window !== window.parent ?
+        this._started = false;
+
+        this._forwardFrameOffset = (
+            window !== window.parent ?
             this._forwardFrameOffsetParent.bind(this) :
-            this._forwardFrameOffsetOrigin.bind(this);
+            this._forwardFrameOffsetOrigin.bind(this)
+        );
 
         this._windowMessageHandlers = new Map([
-            ['getFrameOffset', ({offset, uniqueId}, e) => { return this._onGetFrameOffset(offset, uniqueId, e); }]
+            ['getFrameOffset', ({offset, uniqueId}, e) => this._onGetFrameOffset(offset, uniqueId, e)]
         ]);
+    }
 
+    start() {
+        if (this._started) { return; }
         window.addEventListener('message', this.onMessage.bind(this), false);
+        this._started = true;
     }
 
     async applyOffset(x, y) {
@@ -44,7 +52,6 @@ class FrameOffsetForwarder {
                 chrome.runtime.onMessage.removeListener(runtimeMessageCallback);
                 callback();
                 frameOffsetResolve(params);
-                return false;
             }
         };
         chrome.runtime.onMessage.addListener(runtimeMessageCallback);
--- a/ext/fg/js/frame-offset-forwarder.js
+++ b/ext/fg/js/frame-offset-forwarder.js
@@ -41,7 +41,7 @@ class FrameOffsetForwarder {
         this._started = true;
     }
 
-    async applyOffset(x, y) {
+    async getOffset() {
         const uniqueId = yomichan.generateId(16);
 
         const frameOffsetPromise = yomichan.getTemporaryListenerResult(
@@ -58,7 +58,7 @@ class FrameOffsetForwarder {
             action: 'getFrameOffset',
             params: {
                 uniqueId,
-                offset: [x, y]
+                offset: [0, 0]
             }
         }, '*');
 
--- a/ext/fg/js/frame-offset-forwarder.js
+++ b/ext/fg/js/frame-offset-forwarder.js
@@ -44,17 +44,14 @@ class FrameOffsetForwarder {
     async applyOffset(x, y) {
         const uniqueId = yomichan.generateId(16);
 
-        let frameOffsetResolve = null;
-        const frameOffsetPromise = new Promise((resolve) => (frameOffsetResolve = resolve));
-
-        const runtimeMessageCallback = ({action, params}, sender, callback) => {
-            if (action === 'frameOffset' && isObject(params) && params.uniqueId === uniqueId) {
-                chrome.runtime.onMessage.removeListener(runtimeMessageCallback);
-                callback();
-                frameOffsetResolve(params);
+        const frameOffsetPromise = yomichan.getTemporaryListenerResult(
+            chrome.runtime.onMessage,
+            ({action, params}, {resolve}) => {
+                if (action === 'frameOffset' && isObject(params) && params.uniqueId === uniqueId) {
+                    resolve(params);
+                }
             }
-        };
-        chrome.runtime.onMessage.addListener(runtimeMessageCallback);
+        );
 
         window.parent.postMessage({
             action: 'getFrameOffset',
--- a/ext/fg/js/frame-offset-forwarder.js
+++ b/ext/fg/js/frame-offset-forwarder.js
@@ -80,7 +80,10 @@ class FrameOffsetForwarder {
             sourceFrame = frame;
             break;
         }
-        if (sourceFrame === null) { return; }
+        if (sourceFrame === null) {
+            this._forwardFrameOffsetOrigin(null, uniqueId);
+            return;
+        }
 
         const [forwardedX, forwardedY] = offset;
         const {x, y} = sourceFrame.getBoundingClientRect();
--- a/ext/fg/js/frame-offset-forwarder.js
+++ b/ext/fg/js/frame-offset-forwarder.js
@@ -50,7 +50,8 @@ class FrameOffsetForwarder {
                 if (action === 'frameOffset' && isObject(params) && params.uniqueId === uniqueId) {
                     resolve(params);
                 }
-            }
+            },
+            5000
         );
 
         window.parent.postMessage({
--- a/ext/fg/js/frame-offset-forwarder.js
+++ b/ext/fg/js/frame-offset-forwarder.js
@@ -1,0 +1,94 @@
+/*
+ * Copyright (C) 2020  Alex Yatskov <alex@foosoft.net>
+ * Author: Alex Yatskov <alex@foosoft.net>
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+/* global
+ * apiForward
+ */
+
+class FrameOffsetForwarder {
+    constructor() {
+        this._forwardFrameOffset = window !== window.parent ?
+            this._forwardFrameOffsetParent.bind(this) :
+            this._forwardFrameOffsetOrigin.bind(this);
+
+        this._windowMessageHandlers = new Map([
+            ['getFrameOffset', ({offset, uniqueId}, e) => { return this._onGetFrameOffset(offset, uniqueId, e); }]
+        ]);
+
+        window.addEventListener('message', this.onMessage.bind(this), false);
+    }
+
+    async applyOffset(x, y) {
+        const uniqueId = yomichan.generateId(16);
+
+        let frameOffsetResolve = null;
+        const frameOffsetPromise = new Promise((resolve) => (frameOffsetResolve = resolve));
+
+        const runtimeMessageCallback = ({action, params}, sender, callback) => {
+            if (action === 'frameOffset' && isObject(params) && params.uniqueId === uniqueId) {
+                chrome.runtime.onMessage.removeListener(runtimeMessageCallback);
+                callback();
+                frameOffsetResolve(params);
+                return false;
+            }
+        };
+        chrome.runtime.onMessage.addListener(runtimeMessageCallback);
+
+        window.parent.postMessage({
+            action: 'getFrameOffset',
+            params: {
+                uniqueId,
+                offset: [x, y]
+            }
+        }, '*');
+
+        const {offset} = await frameOffsetPromise;
+        return offset;
+    }
+
+    onMessage(e) {
+        const {action, params} = e.data;
+        const handler = this._windowMessageHandlers.get(action);
+        if (typeof handler !== 'function') { return; }
+        handler(params, e);
+    }
+
+    _onGetFrameOffset(offset, uniqueId, e) {
+        let sourceFrame = null;
+        for (const frame of document.querySelectorAll('frame, iframe:not(.yomichan-float)')) {
+            if (frame.contentWindow !== e.source) { continue; }
+            sourceFrame = frame;
+            break;
+        }
+        if (sourceFrame === null) { return; }
+
+        const [forwardedX, forwardedY] = offset;
+        const {x, y} = sourceFrame.getBoundingClientRect();
+        offset = [forwardedX + x, forwardedY + y];
+
+        this._forwardFrameOffset(offset, uniqueId);
+    }
+
+    _forwardFrameOffsetParent(offset, uniqueId) {
+        window.parent.postMessage({action: 'getFrameOffset', params: {offset, uniqueId}}, '*');
+    }
+
+    _forwardFrameOffsetOrigin(offset, uniqueId) {
+        apiForward('frameOffset', {offset, uniqueId});
+    }
+}
--- a/ext/fg/js/frontend-initialize.js
+++ b/ext/fg/js/frontend-initialize.js
@@ -32,19 +32,15 @@ async function main() {
 
     let popup;
     if (!proxy && (window !== window.parent)) {
-        let rootPopupInformationResolve;
-        const rootPopupInformationPromise = new Promise((resolve) => (rootPopupInformationResolve = resolve));
-
-        const runtimeMessageCallback = ({action, params}, sender, callback) => {
-            if (action === 'rootPopupInformation') {
-                chrome.runtime.onMessage.removeListener(runtimeMessageCallback);
-                callback();
-                rootPopupInformationResolve(params);
+        const rootPopupInformationPromise = yomichan.getTemporaryListenerResult(
+            chrome.runtime.onMessage,
+            ({action, params}, {resolve}) => {
+                if (action === 'rootPopupInformation') {
+                    resolve(params);
+                }
             }
-        };
-        chrome.runtime.onMessage.addListener(runtimeMessageCallback);
+        );
         apiForward('rootPopupRequestInformationBroadcast');
-
         const {popupId, frameId} = await rootPopupInformationPromise;
 
         const frameOffsetForwarder = new FrameOffsetForwarder();
--- a/ext/fg/js/frontend-initialize.js
+++ b/ext/fg/js/frontend-initialize.js
@@ -45,9 +45,9 @@ async function main() {
 
         const frameOffsetForwarder = new FrameOffsetForwarder();
         frameOffsetForwarder.start();
-        const applyFrameOffset = frameOffsetForwarder.applyOffset.bind(frameOffsetForwarder);
+        const getFrameOffset = frameOffsetForwarder.getOffset.bind(frameOffsetForwarder);
 
-        popup = new PopupProxy(popupId, 0, null, frameId, url, applyFrameOffset);
+        popup = new PopupProxy(popupId, 0, null, frameId, url, getFrameOffset);
         await popup.prepare();
     } else if (proxy) {
         popup = new PopupProxy(null, depth + 1, id, parentFrameId, url);
--- a/ext/fg/js/frontend-initialize.js
+++ b/ext/fg/js/frontend-initialize.js
@@ -20,6 +20,7 @@
  * Frontend
  * PopupProxy
  * PopupProxyHost
+ * apiForward
  */
 
 async function main() {
@@ -29,7 +30,25 @@ async function main() {
     const {id, depth=0, parentFrameId, url, proxy=false} = data;
 
     let popup;
-    if (proxy) {
+    if (!proxy && (window !== window.parent)) {
+        let rootPopupInformationResolve;
+        const rootPopupInformationPromise = new Promise((resolve) => (rootPopupInformationResolve = resolve));
+
+        const runtimeMessageCallback = ({action, params}, sender, callback) => {
+            if (action === 'rootPopupInformation') {
+                chrome.runtime.onMessage.removeListener(runtimeMessageCallback);
+                callback();
+                rootPopupInformationResolve(params);
+                return false;
+            }
+        };
+        chrome.runtime.onMessage.addListener(runtimeMessageCallback);
+        apiForward('rootPopupInformationGet');
+
+        const {popupId, frameId} = await rootPopupInformationPromise;
+
+        popup = new PopupProxy(popupId, 0, null, frameId, url);
+    } else if (proxy) {
         popup = new PopupProxy(null, depth + 1, id, parentFrameId, url);
     } else {
         const popupHost = new PopupProxyHost();
--- a/ext/fg/js/frontend-initialize.js
+++ b/ext/fg/js/frontend-initialize.js
@@ -22,6 +22,7 @@
  * PopupProxy
  * PopupProxyHost
  * apiForward
+ * apiOptionsGet
  */
 
 async function main() {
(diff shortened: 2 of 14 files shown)


siikamiika:
There are some issues positioning the popup when an iframe video is made full screen.

ext/fg/js/frame-offset-forwarder.js
```diff
@@ -0,0 +1,94 @@
...
+
+        let frameOffsetResolve = null;
+        const frameOffsetPromise = new Promise((resolve) => (frameOffsetResolve = resolve));
+
+        const runtimeMessageCallback = ({action, params}, sender, callback) => {
+            if (action === 'frameOffset' && isObject(params) && params.uniqueId === uniqueId) {
+                chrome.runtime.onMessage.removeListener(runtimeMessageCallback);
+                callback();
+                frameOffsetResolve(params);
+                return false;
```
toasted-nutbread: Move `return false` outside of the inner block (just always return false; this won't change the behaviour but will give a consistent return type)

ext/fg/js/frame-offset-forwarder.js
```diff
@@ -0,0 +1,94 @@
...
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+/* global
+ * apiForward
+ */
+
+class FrameOffsetForwarder {
+    constructor() {
+        this._forwardFrameOffset = window !== window.parent ?
```
toasted-nutbread: Prefer to wrap multi-line expressions in parentheses

```js
this._forwardFrameOffset = (
    window !== window.parent ?
    this._forwardFrameOffsetParent.bind(this) :
    this._forwardFrameOffsetOrigin.bind(this)
);

File: ext/fg/js/frame-offset-forwarder.js [hunk 3]
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    _onGetFrameOffset(offset, uniqueId, e) {
+        let sourceFrame = null;
+        for (const frame of document.querySelectorAll('frame, iframe:not(.yomichan-float)')) {
+            if (frame.contentWindow !== e.source) { continue; }
+            sourceFrame = frame;
+            break;
+        }
+        if (sourceFrame === null) { return; }
```
toasted-nutbread: May want this to always send a response, even if the frame couldn't be found.

Example case: if the iframe is in a shadow DOM, `document.querySelector` won't return anything, and thus `runtimeMessageCallback` will never be invoked. Repeated calls to `applyOffset` will then result in a memory leak since event callbacks are being added without ever being removed.

ext/fg/js/frontend-initialize.js
```diff
@@ -29,9 +31,32 @@ async function main() {
...
+        const runtimeMessageCallback = ({action, params}, sender, callback) => {
+            if (action === 'rootPopupInformation') {
+                chrome.runtime.onMessage.removeListener(runtimeMessageCallback);
+                callback();
+                rootPopupInformationResolve(params);
+                return false;
+            }
+        };
+        chrome.runtime.onMessage.addListener(runtimeMessageCallback);
+        apiForward('rootPopupInformationGet');
```
toasted-nutbread: `rootPopupInformationGet` => `rootPopupRequestInformationBroadcast`

Since it's not actually returning anything

ext/fg/js/frontend-initialize.js
```diff
@@ -29,9 +31,32 @@ async function main() {
     const {id, depth=0, parentFrameId, ignoreNodes, url, proxy=false} = data;
 
     let popup;
-    if (proxy) {
+    if (!proxy && (window !== window.parent)) {
+        let rootPopupInformationResolve;
+        const rootPopupInformationPromise = new Promise((resolve) => (rootPopupInformationResolve = resolve));
+
+        const runtimeMessageCallback = ({action, params}, sender, callback) => {
```
toasted-nutbread: We have this pattern in a few places now; we may want to abstract it into something that auto-removes the listener. Something like `yomichan.addListenerTemporary(chrome.runtime.onMessage, callback)`. Not necessary to do it now if you think it will make the PR more complex.

ext/fg/js/frontend-initialize.js
```diff
@@ -29,9 +31,32 @@ async function main() {
...
+                rootPopupInformationResolve(params);
+                return false;
+            }
+        };
+        chrome.runtime.onMessage.addListener(runtimeMessageCallback);
+        apiForward('rootPopupInformationGet');
+
+        const {popupId, frameId} = await rootPopupInformationPromise;
+
+        window._frameOffsetForwarder = new FrameOffsetForwarder();
```
toasted-nutbread: Any reason why this is assigned to `window._frameOffsetForwarder`?

ext/fg/js/frame-offset-forwarder.js
```diff
@@ -0,0 +1,94 @@
...
+ */
+
+class FrameOffsetForwarder {
+    constructor() {
+        this._forwardFrameOffset = window !== window.parent ?
+            this._forwardFrameOffsetParent.bind(this) :
+            this._forwardFrameOffsetOrigin.bind(this);
+
+        this._windowMessageHandlers = new Map([
+            ['getFrameOffset', ({offset, uniqueId}, e) => { return this._onGetFrameOffset(offset, uniqueId, e); }]
```
toasted-nutbread: Can be simplified

```js
['getFrameOffset', ({offset, uniqueId}, e) => this._onGetFrameOffset(offset, uniqueId, e)]
```

ext/fg/js/popup-proxy.js
```diff
@@ -21,14 +21,14 @@
  */
 
 class PopupProxy {
-    constructor(id, depth, parentId, parentFrameId, url) {
+    constructor(id, depth, parentId, parentFrameId, url, applyFrameOffset=async (x, y) => [x, y]) {
```
toasted-nutbread: Default `applyFrameOffset` to `null` instead, and then do a `null` check before invoking it later on. This simplifies the default arguments (looks complex) and can also elide an `async` call later on when null.

ext/fg/js/frame-offset-forwarder.js
siikamiika: This won't actually return anything so I don't know why I added the `return` there. Otherwise this style was copied from `_runtimeMessageHandlers` in `Frontend`. I can still make this one use the short syntax.

toasted-nutbread: The reason why that one is bracketed is to make it so that there is no return value, although technically `setVisibleOverride` already doesn't return anything.

ext/fg/js/frame-offset-forwarder.js
siikamiika: This is from `Popup._listenForDisplayPrepareCompleted`. Was there a reason why the return value was added? I could find in https://developer.chrome.com/apps/runtime that if you return `true` it will keep `sendResponse` (`callback` here) valid until it's called even after returning from the listener.

ext/fg/js/frame-offset-forwarder.js
siikamiika: I might add a timeout to the promise in `applyOffset`. I'm also considering creating an abstraction for the temporary listener as you mentioned, and it could take timeout as an argument.

ext/fg/js/frame-offset-forwarder.js
toasted-nutbread: It's probably not needed, particularly because true is never returned. It can be omitted from these functions. Not sure why I added it.

ext/fg/js/frontend-initialize.js
siikamiika: I'll see if it can be easily done to the places this PR uses, and changed to other places later.

ext/fg/js/frontend-initialize.js
siikamiika: I was thinking that it would get garbage collected if there wasn't a reference to it anywhere (had this issue with Python earlier), but not sure how JS works. Also the instance created in the `else` block looked funny with just `new FrameOffsetForwarder();`. I could add a method that actually starts the listener to work around the issue while keeping the instance in local scope.

toasted-nutbread: It won't get GC'd because there are still refs to it in the event listeners.

```js
window.addEventListener('message', this.onMessage.bind(this), false);
```

This implicitly gives `window` a reference to `this`, since `this.onMessage` is bound.

ext/fg/js/frontend-initialize.js
siikamiika: https://github.com/FooSoft/yomichan/pull/417/commits/9fdcedb4001a5e86cc7e648c61bd9961a202fa54 implements an abstraction for the pattern that was used here and also in `Popup._listenForDisplayPrepareCompleted`, but that wasn't converted to use the new abstraction yet.

ext/fg/js/frame-offset-forwarder.js
siikamiika: https://github.com/FooSoft/yomichan/commit/9fdcedb4001a5e86cc7e648c61bd9961a202fa54 adds a timeout that rejects everything that takes more than 30 seconds. The timeouts can be adjusted based on how critical the result is.

ext/mixed/js/core.js
```diff
@@ -288,6 +288,42 @@ const yomichan = (() => {
...
+            if (eventHandler === chrome.runtime.onMessage) {
+                const runtimeMessageCallback = ({action, params}, sender, sendResponse) => {
+                    const cleanupResolve = (value) => {
+                        resolved = true;
+                        eventHandler.removeListener(runtimeMessageCallback);
+                        sendResponse();
+                        resolve(value);
+                    };
+
+                    setTimeout(() => {
```
toasted-nutbread: Store the timer ID and clear the timer in cleanupResolve also (if it hasn't already completed). Clears the timer resource earlier, probably better for memory.

siikamiika: https://github.com/FooSoft/yomichan/pull/417/commits/894c13cc32d70de4a3ca5b78a7ea5fb04b5f791f

siikamiika:
https://github.com/FooSoft/yomichan/pull/417/commits/b64f974d9386ac0d5dd2767ab20caa799a611f14 makes the frame offset calls far less frequent and perceived latency less noticeable by giving the offset a lifetime of 1 second and only calling it synchronously on the first call.

siikamiika:
Now I'm thinking if the cache logic should be moved to `FrameOffsetForwarder`, it could be updated to `PopupProxy` with dependency injection

ext/mixed/js/core.js
siikamiika: Btw, I wonder if `clearTimeout` actually needs to check if the timer has been cleared. From what I can gather it's idempotent and you can even call it on ID's that never existed without warnings. Performance wise an `if (bool)` should be faster than a function call (when it's `false`), but it adds a few lines and some indentation.

ext/fg/js/frame-offset-forwarder.js
toasted-nutbread: This should probably still return some sort of "unknown" indicator and the timeout should be treated as a worst-case scenario. The `getFrameOffset` invocation is likely to happen a lot due to mouse movement, and if the timeout is 5 seconds, a very large message queue can build up.

If an "unknown" indicator is returned, this would also enable the child window to decide to skip the call in the future, since if it's unknown once, it's likely to be unknown forever. Not necessary to implement this now, but I think it's good to have that as possibility.

I think we should also start putting together a HTML page in the test folder with some of the odd edge cases we encounter / can think of, such as the iframe in a shadow DOM and the fullscreen issue.

ext/fg/js/popup-proxy.js
```diff
@@ -51,92 +55,109 @@ class PopupProxy {
...
+    async prepare() {
+        const {id} = await this._invokeHostApi('getOrCreatePopup', {id: this._id, parentId: this._parentId});
+        this._id = id;
+    }
+
     isProxy() {
         return true;
     }
 
+    broadcastRootPopupInformation() {
```
toasted-nutbread: IMO this implementation should be put into frontend, and frontend should be responsible for checking if it is the root popup. The sending of messages to the entire frame seems like something outside the scope of popups.

ext/fg/js/popup-proxy.js
```diff
@@ -51,92 +55,109 @@ class PopupProxy {
...
-    static _convertDOMRectToJson(domRect) {
-        return {
-            x: domRect.x,
-            y: domRect.y,
-            width: domRect.width,
-            height: domRect.height
-        };
+    async _updateFrameOffset() {
+        const firstRun = this._frameOffsetUpdatedAt === null;
+        const expired = firstRun || this._frameOffsetUpdatedAt < Date.now() - 1000;
```
toasted-nutbread: Define 1000 as a variable (private member?) so it's doesn't appear to be a magic number.

ext/mixed/js/core.js
```diff
@@ -288,6 +293,45 @@ const yomichan = (() => {
...
+                        timeoutId = window.setTimeout(() => {
+                            timeoutId = null;
+                            reject(new Error(`Listener timed out in ${timeout} ms`));
+                            eventHandler.removeListener(runtimeMessageCallback);
+                        }, timeout);
+                    }
+
+                    const cleanupResolve = (value) => {
+                        if (timeoutId !== null) {
+                            window.clearTimeout(timeoutId);
```
toasted-nutbread: Set timeoutId to null.

ext/mixed/js/core.js
```diff
@@ -288,6 +293,45 @@ const yomichan = (() => {
...
+                reject = _reject;
+            });
+
+            if (eventHandler === chrome.runtime.onMessage) {
+                const runtimeMessageCallback = ({action, params}, sender, sendResponse) => {
+                    let timeoutId = null;
+                    if (timeout !== null) {
+                        timeoutId = window.setTimeout(() => {
+                            timeoutId = null;
+                            reject(new Error(`Listener timed out in ${timeout} ms`));
```
toasted-nutbread: Move `reject` after `eventHandler.removeListener`; looks more return-like that way.

ext/mixed/js/core.js
```diff
@@ -288,6 +293,45 @@ const yomichan = (() => {
...
 
+        getTemporaryListenerResult(eventHandler, userCallback, timeout=null) {
+            let resolve;
+            let reject;
+            const listenerPromise = new Promise((_resolve, _reject) => {
+                resolve = _resolve;
+                reject = _reject;
+            });
+
+            if (eventHandler === chrome.runtime.onMessage) {
```
toasted-nutbread: I would omit this check and just rely on duck-typing. If we include this type, we may as well force eventHandler to be chrome.runtime.onMessage

ext/mixed/js/core.js
```diff
@@ -288,6 +293,45 @@ const yomichan = (() => {
             this.trigger('orphaned', {error});
         }
 
+        getTemporaryListenerResult(eventHandler, userCallback, timeout=null) {
+            let resolve;
+            let reject;
+            const listenerPromise = new Promise((_resolve, _reject) => {
+                resolve = _resolve;
```
toasted-nutbread: Just move everything inside of the promise executor. `listenerPromise` isn't used anywhere other than the return value, so that should be safer (in case of exceptions, etc.), and it avoids the need to capture resolve and reject.

ext/mixed/js/core.js
siikamiika: I had a comment about this earlier:

> Btw, I wonder if clearTimeout actually needs to check if the timer has been cleared. From what I can gather it's idempotent and you can even call it on ID's that never existed without warnings. Performance wise an if (bool) should be faster than a function call (when it's false), but it adds a few lines and some indentation.

In general I agree that it's good to avoid calling functions with unknown side effects or costly processing multiple times, and I can make it null the ID here to be consistent.

ext/mixed/js/core.js
siikamiika: Yeah, that would be cleaner. I was too excited about using the pattern I recently learned for passing resolve and reject to other methods in a class :smile:

siikamiika: https://github.com/FooSoft/yomichan/pull/417/commits/5ce02f6d29b663d7d80c30341127b058563a02a9

ext/mixed/js/core.js
siikamiika: https://github.com/FooSoft/yomichan/pull/417/commits/34d9ac6ad9bae72c6ce87dc49d30bf5b42908111

ext/mixed/js/core.js
siikamiika: https://github.com/FooSoft/yomichan/pull/417/commits/8c32ce7799fac5dc0694ab51ab65d31c3ea76fe1

ext/mixed/js/core.js
siikamiika: https://github.com/FooSoft/yomichan/commit/8c32ce7799fac5dc0694ab51ab65d31c3ea76fe1

siikamiika:
> This should probably still return some sort of "unknown" indicator and the timeout should be treated as a worst-case scenario. The getFrameOffset invocation is likely to happen a lot due to mouse movement, and if the timeout is 5 seconds, a very large message queue can build up.
> 
> If an "unknown" indicator is returned, this would also enable the child window to decide to skip the call in the future, since if it's unknown once, it's likely to be unknown forever. Not necessary to implement this now, but I think it's good to have that as possibility.

Addressed in https://github.com/FooSoft/yomichan/pull/417/commits/204ae40869f904a34d272204b9ba17d345686752. Updating the timestamp should also partially help in cases when it's always unknown by throttling the checks to one per second.

> I think we should also start putting together a HTML page in the test folder with some of the odd edge cases we encounter / can think of, such as the iframe in a shadow DOM and the fullscreen issue.

I think this is a good idea, I'm currently developing by repeatedly opening a Chrome profile with some of these test cases found in the wild open. There's some delay opening them and I got rate limited more than once :joy:

ext/fg/js/frame-offset-forwarder.js
siikamiika: https://github.com/FooSoft/yomichan/commit/204ae40869f904a34d272204b9ba17d345686752 also relevant here

siikamiika: > This should probably still return some sort of "unknown" indicator and the timeout should be treated as a worst-case scenario. The getFrameOffset invocation is likely to happen a lot due to mouse movement, and if the timeout is 5 seconds, a very large message queue can build up.
> 
> If an "unknown" indicator is returned, this would also enable the child window to decide to skip the call in the future, since if it's unknown once, it's likely to be unknown forever. Not necessary to implement this now, but I think it's good to have that as possibility.

Addressed in https://github.com/FooSoft/yomichan/pull/417/commits/204ae40869f904a34d272204b9ba17d345686752. Updating the timestamp should also partially help in cases when it's always unknown by throttling the checks to one per second.

> I think we should also start putting together a HTML page in the test folder with some of the odd edge cases we encounter / can think of, such as the iframe in a shadow DOM and the fullscreen issue.

I think this is a good idea, I'm currently developing by repeatedly opening a Chrome profile with some of these test cases found in the wild open. There's some delay opening them and I got rate limited more than once :joy:

ext/fg/js/popup-proxy.js
siikamiika: https://github.com/FooSoft/yomichan/pull/417/commits/67c5a8c2fb3fdb0050b7ec448ac8a030c0c9d1fb

ext/fg/js/popup-proxy.js
siikamiika: https://github.com/FooSoft/yomichan/pull/417/commits/320d63cdce7009f1a3733dcdf79f456bfe9131a8

ext/mixed/js/core.js
toasted-nutbread: Yeah, it technically doesn't need to be checked or null'd since the IDs are never reused. It was mainly a consistency remark.

siikamiika:
It's currently not possible to toggle https://github.com/FooSoft/yomichan/pull/417/commits/9b87a55d533b810028249e34e48551243490a143 without reloading the page. The issue is not limited to this option, and it also affects popup depth as noted in https://github.com/FooSoft/yomichan/issues/426. This PR already includes lots of unrelated changes and supporting it would require major refactoring or at least modifying multiple entry points, so I'll leave that for another pull request.

ext/mixed/js/core.js
```diff
@@ -288,6 +293,42 @@ const yomichan = (() => {
             this.trigger('orphaned', {error});
         }
 
+        getTemporaryListenerResult(eventHandler, userCallback, timeout=null) {
+            if (
```
toasted-nutbread, line 297:
```js
if (!(
  typeof eventHandler.addListener === 'function' &&
  typeof eventHandler.removeListener === 'function'
))
```
is probably more accurate

ext/fg/js/popup-proxy.js
```diff
@@ -51,92 +55,110 @@ class PopupProxy {
...
+        if (this._frameOffsetPromise === null && !expired) { return; }
+
+        if (this._frameOffsetPromise !== null) {
+            if (firstRun) {
+                await this._frameOffsetPromise;
+            }
+            return;
+        }
+
+        this._frameOffsetPromise = this._getFrameOffset();
```
toasted-nutbread, line 134: This could probably be simplified, from line 134-156:

```js
async _updateFrameOffset() {
  // ...

  const promise = this._updateFrameOffsetInner();
  if (firstRun) {
    await promise;
  }
}

async _updateFrameOffsetInner() {
  this._frameOffsetPromise = this._getFrameOffset();
  try {
    const offset = await promise;
    this._frameOffset = offset !== null ? offset : [0, 0];
    this._frameOffsetUpdatedAt = Date.now();
  } catch (e) {
    logError(e);
  } finally {
    this._frameOffsetPromise = null;
  }
}
```

ext/fg/js/popup-proxy.js
```diff
@@ -51,92 +55,110 @@ class PopupProxy {
...
 
-    static _convertDOMRectToJson(domRect) {
-        return {
-            x: domRect.x,
-            y: domRect.y,
-            width: domRect.width,
-            height: domRect.height
+    async _updateFrameOffset() {
+        const firstRun = this._frameOffsetUpdatedAt === null;
+        const expired = firstRun || this._frameOffsetUpdatedAt < Date.now() - PopupProxy._frameOffsetExpireTimeout;
```
toasted-nutbread, line 124: There are two `Date.now` calls which may have different values; should the value be stored in a variable at the start of the function? Or is it intended that it will be updated after the async calls?

siikamiika: It doesn't matter too much actually because both the time of initiation and getting the results are a bit off. I can make both use the time of initiation for consistency.

ext/fg/js/popup-proxy.js
siikamiika: You're right, that's a lot cleaner. `Promise`s all the way

ext/fg/js/popup-proxy.js
siikamiika: https://github.com/FooSoft/yomichan/pull/417/commits/883759b9b7f1d3b5ddb8d3f17a29841c83f725dd

ext/fg/js/popup-proxy.js
siikamiika: https://github.com/FooSoft/yomichan/pull/417/commits/883759b9b7f1d3b5ddb8d3f17a29841c83f725dd

ext/mixed/js/core.js
siikamiika: https://github.com/FooSoft/yomichan/pull/417/commits/ff6b79fdc7ef3fd7be4384dec570138e0ad27012

toasted-nutbread: Should be && not || now

toasted-nutbread:
I think this is the case, but just to confirm: without changing any settings, the default behaviour of the extension should currently be the same?

toasted-nutbread:
Looks like my PR merges caused a few merge conflicts, but they should be easy to fix since it's all options related.

siikamiika:
> I think this is the case, but just to confirm: without changing any settings, the default behaviour of the extension should currently be the same?

Apart from some changes to PopupProxy prepare, you're right. The feature itself will be disabled by default and behind advanced settings.

> Looks like my PR merges caused a few merge conflicts, but they should be easy to fix since it's all options related.

No worries, I'll fix it now.

ext/mixed/js/core.js
siikamiika: Sorry, that was careless. I tested it and assumed everything was fine because OR short circuited on the first check. https://github.com/FooSoft/yomichan/pull/417/commits/22a97d916fc6ecab1200b0ffea18cf2d5c9923d4

Issue #421, opened 2020-03-27
translate consecutive っ into a single one


Pintliz:
Consecutive っ indicates a stronger emphasis level than a single っ. words formed with these are not listed in dictionaries and is unable to be scanned.
By translating consecutive っ into a single one cases like (1) could be scanned, these are quite abundant on SNS and dialogues in fiction. imo this is relatively low risk and unlikely to cause unintended behaviors.

cases like (2) and (3) will require the deletion of っ under certain conditions. (for example when it occurs at the 2nd syllable of i-adjectives). these are out of scope of this issue.

(1) かっっっこい => かっこい
(2) かっわいい =>かわいい

(3) かたい
![image](https://user-images.githubusercontent.com/24824399/77765633-7e449900-7079-11ea-8ace-d326361991b7.png)
(https://static1.squarespace.com/static/57055a42ab48deae6cba15fd/t/5707ec58cf80a1640df4d8d4/1460137062627/emphatic-consonants-jipa.pdf)

siikamiika:
This would be doable by extending what was added in https://github.com/FooSoft/yomichan/pull/181, but there would be some extra performance penalty for checking the previous character in addition to the current. Anyway, I wouldn't touch it yet because it would conflict with https://github.com/FooSoft/yomichan/pull/286.

toasted-nutbread:
Proof of concept:
https://github.com/toasted-nutbread/yomichan/compare/master...collapse-small-tsu

siikamiika:
@toasted-nutbread Forgot about that feature. Seems like a better place for this anyway because some dictionaries might have entries with more than one consecutive っ or ッ. Do you think the ZWNJ handling should be moved out of of `source.js` eventually? Currently it would be problematic because maximum scanning length is handled in `source.js` and ZWNJ counts as one.

toasted-nutbread:
It's probably fine for now, though it could work in either location if text source maps are used.

If we want the ZWNJ to not count against the scanning length, that would have to be handled in source.js anyways in some form.

siikamiika:
When I did #181 I tested Chrome on Docs and the length limitation got really noticeable. I think it would be optimal if `TextSource` was like a file that had `seek` (relative) and `read`, and ZWNJ and maybe even surrogate pairs were internal. The API is currently a bit confusing.

toasted-nutbread:
[...]

PR #423, opened 2020-03-28
Text source map

toasted-nutbread:
Since the text source map functionality is likely to be used more in the future, it's probably best to abstract it so it's easier to reuse. Related to #421.

Note that the source map operates on UTF16 characters and not unicode code points. This makes it easier to do things like `string.substring` since the lengths are character-based.

### DIFF
--- a/ext/bg/js/text-source-map.js
+++ b/ext/bg/js/text-source-map.js
@@ -1,0 +1,115 @@
+/*
+ * Copyright (C) 2020  Alex Yatskov <alex@foosoft.net>
+ * Author: Alex Yatskov <alex@foosoft.net>
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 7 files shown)


ext/bg/js/text-source-map.js
```diff
@@ -0,0 +1,115 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+class TextSourceMap {
+    constructor(source, mapping=null) {
+        this._source = source;
+        this._mapping = (Array.isArray(mapping) ? TextSourceMap._normalizeMapping(mapping) : null);
```
siikamiika: Could be simpler to make this `TextSourceMap._createMapping(source)` instead of `null` and add a boolean flag indicating if the mapping has been touched. Now it's initialized in multiple places, even for another `TextSourceMap`. There's a minor performance hit, though.

toasted-nutbread: The reason why I left it as null is for that performance reason. In most of the general use cases, the source map won't be modified, so deferring creation of the array should save allocation time and memory at the cost of only some `null` checks later on. `equals` also has a few additional `null` checks, but this function is only used for unit tests currently.

siikamiika: I guess it's justified because dictionary lookup is such a frequent operation. It's better that the complexity is encapsulated in a class instead of having it in other places of code. I didn't realize that `equals` was for testing only.

PR #424, opened 2020-03-31
Dictionary importer

toasted-nutbread:
This change moves dictionary importing functionality into a standalone class. The motivation is to make a more clear distinction between the two types of functionality: reading vs writing. This should make it easier to understand what sort of changes are being made in the future.

This will be helpful as more internal functions are added to the import process, such as what will happen for #322.


ext/bg/js/database.js
```diff
@@ -448,11 +478,7 @@ class Database {
             prefixWildcardsSupported
         };
 
-        {
-            const transaction = db.transaction(['dictionaries'], 'readwrite');
-            const objectStore = transaction.objectStore('dictionaries');
-            await Database._bulkAdd(objectStore, [summary], 0, 1);
-        }
+        await this.bulkAdd('dictionaries', [summary], 0, 1);
```
siikamiika: Just a comment about the previous code, that's an interesting way to prevent temporary constants from polluting the entire function scope. As done in this change, another solution was to move the repeated code to a dedicated function.

siikamiika:
Were there many changes to the functions extracted from Database into DictionaryImporter in https://github.com/FooSoft/yomichan/pull/424/commits/9052ab8ebd5af505f1992bfc001b226202e2f393? Based on a few samples they look the same.

siikamiika:
I see some static functions were changed to non-static as a part of the extract process. Is it desirable to avoid using `static` in the future for class internal stuff? To my understanding the benefit is reducing memory usage in classes that have many instances and public utility functions. The downside is that sometimes you will need access to `this` and it can cause confusion when you try to access it from a static method.

ext/bg/js/database.js
toasted-nutbread: Yeah, I don't even think it was technically necessary in this case.

siikamiika: I tested and there was a linter error about redeclaring it below in another block. JS wouldn't need it, though.

toasted-nutbread:
> Were there many changes to the functions extracted from Database into DictionaryImporter in [9052ab8](https://github.com/FooSoft/yomichan/commit/9052ab8ebd5af505f1992bfc001b226202e2f393)? Based on a few samples they look the same.

It was intended to be as minimal as possible, so not many. The primary changes were things like making a few members of Database public and slightly tweaking how they work (bulkAdd). Everything else is basically copy-pasted (unfortunately the diffs don't have a good way of telling us that).



> I see some static functions were changed to non-static as a part of the extract process. Is it desirable to avoid using `static` in the future for class internal stuff? To my understanding the benefit is reducing memory usage in classes that have many instances and public utility functions. The downside is that sometimes you will need access to `this` and it can cause confusion when you try to access it from a static method.

So I've been thinking about this for a while, but I think a lot of the benefits of `static` that exist in a traditional strictly typed language are lost in Javascript, at least for "private" statics. The reason is because ultimately there is still an object reference to the static, the same as a member function. The only difference is that the object is the class or `this.constructor`. So for statics which are "private", I think the better solution is to use private member functions.

The main benefit to this is that eslint can do some static analysis based on underscore-prefixed "private" functions: see [no-underscore-dangle](https://eslint.org/docs/rules/no-underscore-dangle). However, it wants the functions to only follow `this` or `this.constructor`. So this would open up that avenue.

siikamiika:
> So I've been thinking about this for a while, but I think a lot of the benefits of `static` that exist in a traditional strictly typed language are lost in Javascript, at least for "private" statics. The reason is because ultimately there is still an object reference to the static, the same as a member function. The only difference is that the object is the class or `this.constructor`. So for statics which are "private", I think the better solution is to use private member functions.
> 
> So unlike in a compiled language where the "this" argument can be compiled out in static functions, there is still a "this" for static functions in Javascript. So nothing is truly saved by doing that for internally used methods.

Right, if I remember correctly `this` is always what's left of the dot when a function is called unless you do `bind`. I assumed classes were magic where it doesn't apply, but then again why not.

> The main benefit to this is that eslint can do some static analysis based on underscore-prefixed "private" functions: see [no-underscore-dangle](https://eslint.org/docs/rules/no-underscore-dangle). However, it wants the functions to only follow `this` or `this.constructor`. So this would open up that avenue.

That would be really useful for actually enforcing that private methods are private. There's a related place in code that should probably be fixed at some point:
https://github.com/FooSoft/yomichan/blob/a0c4ce779d35cab39c62ca42ad3fe58a82faa1bb/ext/fg/js/popup.js#L110-L118
Not sure if that's considered a violation because it's an alias of the instance itself or another instance but still of the same class.

toasted-nutbread:
> Right, if I remember correctly this is always what's left of the dot when a function is called unless you do bind. I assumed classes were magic where it doesn't apply, but then again why not.

Even `bind`'d functions still have a `this`. It's similar to how there's always a return value for functions in Javascript. Perhaps the JS compiler will optimize return values/`this` arguments out when it determines it's unused, but in terms of writing the function in Javascript, it's not really a relevant concern.

>  Not sure if that's considered a violation because it's an alias of the instance itself or another instance but still of the same class.

Yeah, there are a few places that would have to be handled differently if this switch is fully made. It's a longer term goal along with fully implementing private members for everything, which is only partially done.

PR #427, opened 2020-04-02
Anki card media injection refactor

toasted-nutbread:
Moving the media injection functions into AnkiNoteBuilder for some preliminary cleanup. Subsequent tasks will include:
* Removing `optionsContext` from `injectAudio` in favor of adding the relevant parameters directly to the object. The `optionsContext` is used to pass a copy of `options` to `AudioUriBuilder`'s `_getUriTextToSpeech` and `_getUriTextToSpeechReading` where only `options.audio.textToSpeechVoice` is used. This can be simplified to just passing `textToSpeechVoice` as a part of the `details` object that is used.
* Potentially refactoring how audio is injected into cards. This was originally brought up in https://github.com/FooSoft/yomichan/pull/277#issuecomment-552145454.

### DIFF
--- a/ext/bg/js/anki-note-builder.js
+++ b/ext/bg/js/anki-note-builder.js
@@ -17,7 +17,8 @@
  */
 
 class AnkiNoteBuilder {
-    constructor({renderTemplate}) {
+    constructor({audioSystem, renderTemplate}) {
+        this._audioSystem = audioSystem;
         this._renderTemplate = renderTemplate;
     }
 
@@ -84,14 +85,14 @@ class AnkiNoteBuilder {
         });
     }
 
-    async injectAudio(definition, fields, sources, audioSystem, optionsContext) {
+    async injectAudio(definition, fields, sources, optionsContext) {
         if (!this._containsMarker(fields, 'audio')) { return; }
 
         try {
             const expressions = definition.expressions;
             const audioSourceDefinition = Array.isArray(expressions) ? expressions[0] : definition;
 
-            const {uri} = await audioSystem.getDefinitionAudio(audioSourceDefinition, sources, {tts: false, optionsContext});
+            const {uri} = await this.audioSystem.getDefinitionAudio(audioSourceDefinition, sources, {tts: false, optionsContext});
             const filename = this._createInjectedAudioFileName(audioSourceDefinition);
             if (filename !== null) {
                 definition.audio = {url: uri, filename};
--- a/ext/bg/js/anki-note-builder.js
+++ b/ext/bg/js/anki-note-builder.js
@@ -84,6 +84,82 @@ class AnkiNoteBuilder {
         });
     }
 
+    async injectAudio(definition, fields, sources, audioSystem, optionsContext) {
+        let usesAudio = false;
+        for (const fieldValue of Object.values(fields)) {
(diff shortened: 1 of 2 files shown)


siikamiika:
Looks good. Just want to hear your opinion on passing entire instances to other functions vs. passing them or `this` to a constructor or `prepare`.

https://github.com/FooSoft/yomichan/blob/3a5d6b93834ca5b99fe6b12793516fb26ab40650/ext/bg/js/backend.js#L454-L470

We have used a style like this before:

https://github.com/FooSoft/yomichan/blob/3df78904cf734da208c6fd1b6ae1cd6612323148/ext/bg/js/backend.js#L53-L54

Also concerns https://github.com/FooSoft/yomichan/pull/424

toasted-nutbread:
For `injectScreenshot`, it's better to pass it in the function call, since `this.anki` changes. For `this.audioSystem`, it could honestly be done either way, but I used the same pattern I used for `injectScreenshot` due to proximity in the code.

siikamiika:
I didn't actually think of passing a function to the `AnkiNoteBuilder` constructor but to `injectAudio` and `injectScreenshot` instead. Looking at this again, passing the required dependencies to their constructors would be better. Maybe `AnkiNull` should be refactored away at some point to allow it.

toasted-nutbread:
In terms of passing the instance vs passing a bound function, it should be fine to just pass the instance in this case, since the object is dedicated to a specific functionality. There aren't interface declarations in JS, so it isn't explicitly typed what function we are expecting to use from the instance, but that should be fine.

`audioSystem` instance could probably be passed in the constructor, and you're right that the AnkiNull could be refactored at some point to have a start/stop functionality like some of the other codebase.

siikamiika:
Ok, thanks for sharing your opinion on this. Do you want to do the discussed changes in this PR or later together with AnkiNull? I think it's fine otherwise.

toasted-nutbread:
I'll update the audio part here and we can tackle the anki part later.

toasted-nutbread:
7c7a57a9e3982ba2869ec1e22509a33eb523212f

toasted-nutbread:
And just to further clarify why I think that passing the object instance in this situation is fine, contrasted to the places I was trying to avoid passing the `Backend` reference to its children: in that situation I am trying to avoid the bidirectional dependency, where `Backend` has a reference to `ClassX`, and `ClassX` has a reference to `Backend`.

If `Backend` has a reference to `ClassY`, and it shares the reference to `ClassY` with an instance of `ClassX`, that is fine because the dependencies are all single directional.

siikamiika:
> And just to further clarify why I think that passing the object instance in this situation is fine, contrasted to the places I was trying to avoid passing the `Backend` reference to its children: in that situation I am trying to avoid the bidirectional dependency, where `Backend` has a reference to `ClassX`, and `ClassX` has a reference to `Backend`.
> 
> If `Backend` has a reference to `ClassY`, and it shares the reference to `ClassY` with an instance of `ClassX`, that is fine because the dependencies are all single directional.

Makes sense, the code is much easier to understand when data flows in one direction. It could also cause a Promise "deadlock" or an accidental infinite loop. In this case it doesn't seem to matter much to Backend what AnkiNoteBuilder does with the AudioSystem instance.

PR #430, opened 2020-04-04
Create some tests for fullscreen elements, <iframe>s, and shadow DOMs

toasted-nutbread:
https://github.com/FooSoft/yomichan/pull/417#discussion_r400582235

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -106,6 +106,7 @@
         },
         {
             "files": ["test/**/*.js"],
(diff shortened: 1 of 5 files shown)


siikamiika:
Thanks!

To make the iframe tests work on Firefox, you have to add [`allowfullscreen="true"`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Feature-Policy/fullscreen) to the iframes.

Issue #429, opened 2020-04-04
Dictionaries on Yomichan's readme/homepage & licensing


toasted-nutbread:
Primarily addressed @FooSoft:

With the initial completion of #385, another task is determining how to make users aware of this dictionary. I am wondering how the dictionary file should be hosted (on foosoft.net like the others? or just a link the the repository that I made?), and if there are any special concerns about licensing or anything else. If it's decided that the files aren't to be hosted on foosoft's site, and links to github repository release pages are used instead, would it be good to have some sort of "Third party dictionaries" section that is separate from the others?

In terms of licensing, I have actually already embedded attribution information inside the generated database. In the index file, there is additional information for `author`, `url`, `description`, and `attribution`. Yomichan doesn't store or present this information yet, but that is one of my next tasks to implement.
https://github.com/toasted-nutbread/yomichan-pitch-accent-dictionary/blob/d8e7f181d04d0ed08f00de359b32dd5b9a0221bd/main.js#L142-L151
https://github.com/toasted-nutbread/yomichan-pitch-accent-dictionary/blob/d8e7f181d04d0ed08f00de359b32dd5b9a0221bd/attribution.txt

This also brings us back to #297 (and also #322). If this dictionary had attribution information included and Yomichan is updated to display this information on the settings page, would this be able to be included in the list of dictionaries?

FooSoft:
Yup, makes sense for it to go on FooSoft.net, especially considering that you will want to have it in order to use all of the Yomichan features available. I will add it there along with a link to your generator script. 

I think the main thing will be to have proper attribution on the homepage, as that will be the actual point of distribution. I believe the main reason there is dictionary attribution in Yomichan right now is that back in the olden days, all the dictionary data was distributed inside the extension itself. If the extension is merely consuming data coming from elsewhere, I don't think there is a requirement that you have to display the associated copyright/attribution info, but IANAL. That being said, it would still be a good feature to help users find out more about the source data that is being used from inside of Yomichan.

toasted-nutbread:
Sounds good. I'll work on making the few changes required to show the extra dictionary data fields and I'll comment on this thread again once completed.

toasted-nutbread:
#433 has now been merged, which can show the additional dictionary metadata added to the pitch accents dictionary. There is one more PR open which should come before the next release, and that is #437, which addresses some error messages in the backend. #376 is up to date so far with the changes.

FooSoft:
[...]

PR #433, opened 2020-04-05
Extra dictionary information

toasted-nutbread:
Add support for the extra dictionary information added in #385, mentioned in #429.

### DIFF
--- a/ext/bg/js/dictionary-importer.js
+++ b/ext/bg/js/dictionary-importer.js
@@ -150,13 +150,7 @@ class DictionaryImporter {
         }
 
         // Add dictionary
-        const summary = {
-            title: dictionaryTitle,
-            revision: index.revision,
-            sequenced: index.sequenced,
-            version,
-            prefixWildcardsSupported
-        };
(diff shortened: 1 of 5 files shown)


ext/bg/js/dictionary-importer.js
```diff
@@ -199,6 +193,25 @@ class DictionaryImporter {
...
+            revision: index.revision,
+            sequenced: index.sequenced,
+            version
+        };
+
+        const {author, url, description, attribution} = index;
+        if (typeof author === 'string') { summary.author = author; }
+        if (typeof url === 'string') { summary.url = url; }
+        if (typeof description === 'string') { summary.description = description; }
+        if (typeof attribution === 'string') { summary.attribution = attribution; }
```
siikamiika, line 208: The type check here might be redundant if it's possible to assign the entire `index` into `summary` as-is because it's already checked in the schema

toasted-nutbread: Yeah, and that would also allow unexpected data to be stored in the database, since the schema permits extra properties.

PR #434, opened 2020-04-05
Error fixes

toasted-nutbread:
There were some issues with dictionary import errors showing up as `[object Object]` due to being cross-window references.


ext/bg/js/settings/dictionaries.js
```diff
@@ -643,9 +643,9 @@ async function onDictionaryImport(e) {
             await settingsSaveOptions();
 
             if (errors.length > 0) {
-                errors.push(...errors);
```
toasted-nutbread, line 646: And I have no idea why this was even here.

PR #437, opened 2020-04-08
Backend api handler changes

toasted-nutbread:
This PR fixes a few issues causing the following error message to be shown:

```
Unchecked runtime.lastError: The message port closed before a response was received.
```

This was primarily caused by `apiForward`, which was was awaiting the response of its `sendMessage` call, rather than ignoring it. This behaviour was initially described in #185. There was a similar issue in the `yomichanCoreReady` handler.

This PR makes the following changes:
* API message handlers are now defined as an object of the form `{handler: function, async: boolean}`. The `handler` is the same function as before, and the `async` value indicates whether the function returns a promise or not. When `async` is `false`, the handler can be optimized to return early.
* A few handlers which didn't require `async` have been converter to synchronous handlers. This includes the handlers for the following:
  * `optionsSchemaGet`
  * `optionsGet`
  * `optionsGetFull`
  * `commandExec`
  * `getMessageToken`
  * `getDefaultAnkiFieldTemplates`
  * `yomichanCoreReady`
  * `forward`*
* The `apiForward` handler no longer waits for its internal `sendMessage` calls.
* `apiForward` has been renamed to `apiBroadcast` to better reflect its use.
* The `apiYomichanCoreReady` handler no longer waits for its internal `sendMessage` calls.

### DIFF
--- a/ext/bg/js/backend.js
+++ b/ext/bg/js/backend.js
@@ -94,7 +94,7 @@ class Backend {
             ['commandExec', {handler: this._onApiCommandExec.bind(this), async: false}],
             ['audioGetUri', {handler: this._onApiAudioGetUri.bind(this), async: true}],
             ['screenshotGet', {handler: this._onApiScreenshotGet.bind(this), async: true}],
-            ['forward', {handler: this._onApiForward.bind(this), async: true}],
+            ['forward', {handler: this._onApiForward.bind(this), async: false}],
             ['frameInformationGet', {handler: this._onApiFrameInformationGet.bind(this), async: true}],
             ['injectStylesheet', {handler: this._onApiInjectStylesheet.bind(this), async: true}],
             ['getEnvironmentInfo', {handler: this._onApiGetEnvironmentInfo.bind(this), async: true}],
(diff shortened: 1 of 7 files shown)


siikamiika:
Thanks! I also saw these errors sometimes, but thought they were just caused by heavy load.

> * The `apiForward` handler no longer waits for its internal `sendMessage` calls.

I've been wondering about this, what does even happen when multiple recipients return something to the callback?

> * `apiForward` has been renamed to `apiBroadcast` to better reflect its use.

I agree that the name could use a change, but couldn't `apiBroadcast` be misunderstood too? Maybe `apiBroadcastTab` since it only affects the tab where it's called from.

toasted-nutbread:
> I've been wondering about this, what does even happen when multiple recipients return something to the callback?

I believe the callback gets invoked multiple times.

> I agree that the name could use a change, but couldn't `apiBroadcast` be misunderstood too? Maybe `apiBroadcastTab` since it only affects the tab where it's called from.

I went with just `apiBroadcast` for simplicity, but I could change it to `apiBroadcastTab`. My motivation was that "forward" sounds more like something that should return a value, when the current implementation does/should not do that. "Broadcast" sounded more like something that doesn't expect a response.

(Also, whenever I hear "forward" I think of C++'s `forward`, which is quite different. I think it was a sub-optimal naming choice on my part when I initially added it :) )

siikamiika:
> I believe the callback gets invoked multiple times.

Ok, most places currently use promise resolve there so that would mean that only the first one is handled, but other functions could have weird side effects if called multiple times.

> I went with just `apiBroadcast` for simplicity, but I could change it to `apiBroadcastTab`.

I think that could add some clarity because the API functions can do lots of things not limited to the current tab.

> My motivation was that "forward" sounds more like something that should return a value, when the current implementation does/should not do that. "Broadcast" sounded more like something that doesn't expect a response.
> 
> (Also, whenever I hear "forward" I think of C++'s `forward`, which is quite different. I think it was a sub-optimal naming choice on my part when I initially added it :) )

The way I thought of it first was forwarding mail because you can't process it yourself. Then I was confused when I realized that there can be many recipients.

toasted-nutbread:
a296c758b95b428777548f5aada1be13429faf03

PR #438, opened 2020-04-10
Display generator refactoring

toasted-nutbread:
As I was working on #322, I ran into something that would require some changes to `DisplayGenerator`. Rather than trying to incorporate it into that change, I decided to to isolate it into a separate PR.

The following changes are made:
* `static` functions converted to normal functions. (https://github.com/FooSoft/yomichan/pull/424#issuecomment-609039565)
* Marked functions as private.
* Renamed identifiers which include "only" to use "disambiguation" instead. This is in order to improve clarity and better match `_createPitchAccentDisambiguations`.
* Removed `data-multi` attribute assignment and usage, since it is redundant.
* Refactored `_appendMultiple` to be more streamlined and support passing additional arguments. (This was the change I needed for #322.)
* Some cleanup related to the above changes.

### DIFF
--- a/ext/mixed/css/display.css
+++ b/ext/mixed/css/display.css
@@ -298,13 +298,13 @@ button.action-button {
     content: "\3001";
 }
 
-.term-expression-list[data-multi=true]>.term-expression:last-of-type:after {
+.entry[data-expression-multi=true] .term-expression-list>.term-expression:last-of-type:after {
     font-size: 2em;
     content: "\3000";
     visibility: hidden;
 }
 
-.term-expression-list[data-multi=true] .term-expression-details {
+.entry[data-expression-multi=true] .term-expression-list .term-expression-details {
     display: inline-block;
     position: relative;
     width: 0;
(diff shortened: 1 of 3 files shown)


siikamiika:
Looks like a good change and I tested that it works with different result grouping modes.

There's an issue about DOM element datasets I've been thinking, if you use `camelCase` it's not possible to easily look up the occurrences in CSS because it's automatically converted to `kebab-case`. Would it be worth it to break the convention and use `snake_case` for the dataset like this?

```javascript
element.dataset['some_variable'] = value;
```

Unfortunately `dataset['some-variable']` is not allowed. There could also be a helper that converts `some-variable` to `someVariable` so that you could write it as it appears in CSS (with some performance penalty).

toasted-nutbread:
> Would it be worth it to break the convention and use `snake_case` for the dataset like this?

I think that would have the same problem, this time with an additional convention + having to use `element.dataset.some_variable` (current lint rules error for `element.dataset['some_variable']`). The alternative is to use `element.setAttribute('data-some-variable', value)`, but that is much more verbose. IMO the brevity is worth the slight difficulty increase in searchability.

siikamiika:
I would actually prefer `element.setAttribute('data-some-variable', value)` although that would be the same as using completely custom data attributes, but I guess this issue could also be solved with proper tooling since dataset is a feature in browsers.

toasted-nutbread:
For what it's worth, `element.style.someProperty` properties have the same behaviour of converting case convention, and that is generally not a huge issue.

siikamiika:
It's often useful to scan the codebase for a name when changing it or finding a code path. You can't rename built-in stuff like styles, but datasets are different. I agree it's not a huge issue for datasets either, just adds another step to the process.

Issue #431, opened 2020-04-04
Apply options related to popup creation without reloading the page


siikamiika:
It's currently not possible to toggle https://github.com/FooSoft/yomichan/pull/417/commits/9b87a55d533b810028249e34e48551243490a143 without reloading the page. The issue is not limited to this option, and it also affects popup depth as noted in https://github.com/FooSoft/yomichan/issues/426. This PR already includes lots of unrelated changes and supporting it would require major refactoring or at least modifying multiple entry points, so I'll leave that for another pull request.

_Originally posted by @siikamiika in https://github.com/FooSoft/yomichan/pull/417#issuecomment-609105243_

siikamiika:
Related: https://github.com/FooSoft/yomichan/pull/411

PR #439, opened 2020-04-10
Popup live toggle

siikamiika:
Fixes https://github.com/FooSoft/yomichan/issues/431

### DIFF
--- a/ext/fg/js/frontend-initialize.js
+++ b/ext/fg/js/frontend-initialize.js
@@ -73,8 +73,6 @@ async function main() {
 
     const isIframe = !proxy && (window !== window.parent);
 
-    const initEventDispatcher = new EventDispatcher();
-
     const popups = {
         iframe: null,
         proxy: null,
@@ -99,18 +97,18 @@ async function main() {
             popups.normal = popup;
         }
 
-        if (isSearchPage) {
-            const disabled = !options.scanning.enableOnSearchPage;
-            initEventDispatcher.trigger('setDisabledOverride', {disabled});
-        }
-
-        if (isIframe) {
-            initEventDispatcher.trigger('popupChange', {popup});
-        }
-
         if (frontend === null) {
-            frontend = new Frontend(popup, initEventDispatcher);
+            frontend = new Frontend(popup);
             await frontend.prepare();
+        } else {
+            if (isSearchPage) {
+                const disabled = !options.scanning.enableOnSearchPage;
+                frontend.setDisabledOverride(disabled);
+            }
+
+            if (isIframe) {
+                await frontend.setPopup(popup);
+            }
         }
     };
 
--- a/ext/fg/js/frontend-initialize.js
+++ b/ext/fg/js/frontend-initialize.js
@@ -24,7 +24,7 @@
  * apiOptionsGet
  */
 
-async function createIframePopupProxy(url) {
+async function createIframePopupProxy(url, frameOffsetForwarder) {
     const rootPopupInformationPromise = yomichan.getTemporaryListenerResult(
         chrome.runtime.onMessage,
         ({action, params}, {resolve}) => {
@@ -36,8 +36,6 @@ async function createIframePopupProxy(url) {
     apiBroadcastTab('rootPopupRequestInformationBroadcast');
     const {popupId, frameId} = await rootPopupInformationPromise;
 
-    const frameOffsetForwarder = new FrameOffsetForwarder();
-    frameOffsetForwarder.start();
     const getFrameOffset = frameOffsetForwarder.getOffset.bind(frameOffsetForwarder);
 
     const popup = new PopupProxy(popupId, 0, null, frameId, url, getFrameOffset);
@@ -47,9 +45,6 @@ async function createIframePopupProxy(url) {
 }
 
 async function getOrCreatePopup(depth) {
-    const frameOffsetForwarder = new FrameOffsetForwarder();
-    frameOffsetForwarder.start();
-
     const popupHost = new PopupProxyHost();
     await popupHost.prepare();
 
@@ -81,14 +76,20 @@ async function main() {
 
     let frontend = null;
     let frontendPreparePromise = null;
+    let frameOffsetForwarder = null;
 
     const applyOptions = async () => {
         const optionsContext = {depth: isSearchPage ? 0 : depth, url};
         const options = await apiOptionsGet(optionsContext);
 
+        if (!proxy && frameOffsetForwarder === null) {
+            frameOffsetForwarder = new FrameOffsetForwarder();
+            frameOffsetForwarder.start();
+        }
+
         let popup;
         if (isIframe && options.general.showIframePopupsInRootFrame) {
-            popup = popups.iframe || await createIframePopupProxy(url);
+            popup = popups.iframe || await createIframePopupProxy(url, frameOffsetForwarder);
             popups.iframe = popup;
         } else if (proxy) {
             popup = popups.proxy || await createPopupProxy(depth, id, parentFrameId, url);
--- a/ext/fg/js/frontend-initialize.js
+++ b/ext/fg/js/frontend-initialize.js
@@ -80,6 +80,7 @@ async function main() {
     };
 
     let frontend = null;
+    let frontendPreparePromise = null;
 
     const applyOptions = async () => {
         const optionsContext = {depth: isSearchPage ? 0 : depth, url};
@@ -99,8 +100,10 @@ async function main() {
 
         if (frontend === null) {
             frontend = new Frontend(popup);
-            await frontend.prepare();
+            frontendPreparePromise = frontend.prepare();
+            await frontendPreparePromise;
         } else {
+            await frontendPreparePromise;
             if (isSearchPage) {
                 const disabled = !options.scanning.enableOnSearchPage;
                 frontend.setDisabledOverride(disabled);
--- a/ext/fg/js/frontend-initialize.js
+++ b/ext/fg/js/frontend-initialize.js
@@ -28,7 +28,18 @@ async function main() {
     await yomichan.prepare();
 
     const data = window.frontendInitializationData || {};
-    const {id, depth=0, parentFrameId, url, proxy=false} = data;
+    const {id, depth=0, parentFrameId, url, proxy=false, isSearchPage=false} = data;
+
+    const initEventDispatcher = new EventDispatcher();
+
+    yomichan.on('optionsUpdated', async () => {
+        const optionsContext = {depth: isSearchPage ? 0 : depth, url};
+        const options = await apiOptionsGet(optionsContext);
+        if (isSearchPage) {
+            const disabled = !options.scanning.enableOnSearchPage;
+            initEventDispatcher.trigger('setDisabledOverride', {disabled});
+        }
+    });
 
     const optionsContext = {depth, url};
     const options = await apiOptionsGet(optionsContext);
@@ -65,7 +76,7 @@ async function main() {
         popup = popupHost.getOrCreatePopup(null, null, depth);
     }
 
-    const frontend = new Frontend(popup);
+    const frontend = new Frontend(popup, initEventDispatcher);
     await frontend.prepare();
 }
 
--- a/ext/fg/js/frontend-initialize.js
+++ b/ext/fg/js/frontend-initialize.js
@@ -24,6 +24,47 @@
  * apiOptionsGet
  */
 
+async function createIframePopupProxy(url) {
+    const rootPopupInformationPromise = yomichan.getTemporaryListenerResult(
+        chrome.runtime.onMessage,
+        ({action, params}, {resolve}) => {
+            if (action === 'rootPopupInformation') {
+                resolve(params);
+            }
+        }
+    );
+    apiBroadcastTab('rootPopupRequestInformationBroadcast');
+    const {popupId, frameId} = await rootPopupInformationPromise;
+
+    const frameOffsetForwarder = new FrameOffsetForwarder();
+    frameOffsetForwarder.start();
+    const getFrameOffset = frameOffsetForwarder.getOffset.bind(frameOffsetForwarder);
+
+    const popup = new PopupProxy(popupId, 0, null, frameId, url, getFrameOffset);
+    await popup.prepare();
+
+    return popup;
+}
+
+async function getOrCreatePopup(depth) {
+    const frameOffsetForwarder = new FrameOffsetForwarder();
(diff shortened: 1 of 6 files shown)


siikamiika:
There should probably be a forced settings update when using the root frame popup after using it from an iframe, but I haven't investigated it yet

ext/fg/js/popup-nested.js
```diff
@@ -53,3 +36,40 @@ async function popupNestedInitialize(id, depth, parentFrameId, url) {
         document.body.appendChild(script);
     }
 }
+
+let popupNestedInitialized = false;
+
+async function popupNestedInitialize(id, depth, parentFrameId, url) {
+    if (popupNestedInitialized) {
```
toasted-nutbread, line 43: I don't think `popupNestedInitialized` is necessary anymore. Please remove.

ext/fg/js/frontend-initialize.js
```diff
@@ -25,49 +25,99 @@
...
-        popup = popupHost.getOrCreatePopup(null, null, depth);
-    }
-
-    const frontend = new Frontend(popup);
-    await frontend.prepare();
+    const {id, depth=0, parentFrameId, url, proxy=false, isSearchPage=false} = data;
+
+    const isIframe = !proxy && (window !== window.parent);
+
+    const initEventDispatcher = new EventDispatcher();
```
toasted-nutbread, line 77: The use of `initEventDispatcher ` can probably be removed by some simplification.

`initEventDispatcher` is only used by `frontend`, which is only instantiated once. Add a `setPopup` method on it for `popupChange`, and `setDisabledOverride` can probably be handled in a similar way, or by setting the popup to `null`.

ext/mixed/js/text-scanner.js
```diff
@@ -23,11 +23,13 @@
  */
 
 class TextScanner {
-    constructor(node, ignoreElements, ignorePoints) {
+    constructor(node, ignoreElements, ignorePoints, canEnable=null) {
```
toasted-nutbread, line 26: Rather than adding `canEnable`, might be better to add it as an argument to `setEnabled`. `setEnabled` is only called from `TextScanner.setOptions`, which is called from the subclasses of `TextScanner` (i.e. `Frontend`).

Meaning you can just pass `this.popup.depth <= this.options.scanning.popupNestingMaxDepth && !this._disabledOverride` here:
https://github.com/FooSoft/yomichan/blob/791c51f836c58ea7783acfc077d315c9897f224c/ext/fg/js/frontend.js#L147

And `true` here (or default value):
https://github.com/FooSoft/yomichan/blob/791c51f836c58ea7783acfc077d315c9897f224c/ext/bg/js/search-query-parser.js#L109

This should give better control over when the object is enabled/disabled.

ext/fg/js/popup-nested.js
siikamiika: It's called from `float.js`, but I can't think of cases when it would actually be called multiple times except maybe reloading the extension, and this check wouldn't help with that. Maybe there should be a check similar to the one in `search-frontend.js`

```javascript
        const node = document.querySelector(`script[src='${src}']`);
        if (node !== null) { continue; }
```

ext/fg/js/frontend-initialize.js
siikamiika: https://github.com/FooSoft/yomichan/pull/439/commits/44ff4d889942987adf365a701c01eb8c3f98c2dd

I'd rather not set popup to `null` because then I'd have to potentially handle nulls all the places where popup is used.

ext/mixed/js/text-scanner.js
siikamiika: https://github.com/FooSoft/yomichan/pull/439/commits/6b47f80f195954f158de44b2971f653c3e23b7b8

siikamiika:
Seems like something's up with the CI, none of the tests go through

ext/fg/js/popup-nested.js
toasted-nutbread: Yeah, that prepare should only be called once due to the `_prepareInvoked` guard, so that should handle it.

toasted-nutbread:
> Seems like something's up with the CI, none of the tests go through

Re-running it manually seems to have worked. ¯\\\_(ツ)\_/¯

ext/fg/js/popup-nested.js
siikamiika: https://github.com/FooSoft/yomichan/pull/439/commits/98d7e095a5c3533a9a2d787f3ee15b1640f77da0

siikamiika:
> There should probably be a forced settings update when using the root frame popup after using it from an iframe, but I haven't investigated it yet

Actually, the settings should be the same regardless of the iframe proxy because the context is the same there with the current profile conditions.

ext/fg/js/frontend-initialize.js
```diff
@@ -24,49 +24,100 @@
...
-
-    let popup;
-    if (!proxy && (window !== window.parent) && options.general.showIframePopupsInRootFrame) {
-        const rootPopupInformationPromise = yomichan.getTemporaryListenerResult(
-            chrome.runtime.onMessage,
-            ({action, params}, {resolve}) => {
-                if (action === 'rootPopupInformation') {
-                    resolve(params);
-                }
+    const {id, depth=0, parentFrameId, url=window.top.location.href, proxy=false, isSearchPage=false} = data;
```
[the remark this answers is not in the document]
siikamiika: I wonder what's going to happen with the line below:
```javascript
    const isIframe = !proxy && (window !== window.parent);
```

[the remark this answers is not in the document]
siikamiika: I tested and that doesn't throw an error

ext/fg/js/frontend-initialize.js
```diff
@@ -24,49 +24,100 @@
...
+    const getFrameOffset = frameOffsetForwarder.getOffset.bind(frameOffsetForwarder);
+
+    const popup = new PopupProxy(popupId, 0, null, frameId, url, getFrameOffset);
+    await popup.prepare();
+
+    return popup;
+}
+
+async function getOrCreatePopup(depth) {
+    const frameOffsetForwarder = new FrameOffsetForwarder();
```
toasted-nutbread, line 50: Two FrameOffsetForwarder instances can exist at the same time this way; one from `createIframePopupProxy` and one from `getOrCreatePopup`.

ext/fg/js/frontend-initialize.js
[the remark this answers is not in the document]
siikamiika: It seems that there's still an issue with `DisplayFloat`'s `optionsContext`. Fixing that would be quite complicated, and not necessarily something I want to do in this PR. Should I revert https://github.com/FooSoft/yomichan/pull/439/commits/b80fd00511c8d4fb02d0462c7f41e2c12c036367 and https://github.com/FooSoft/yomichan/pull/439/commits/b0d94e8b50c7a434415610175f63bd7d59e991d1 and create a new commit  making the default value `window.location.href`?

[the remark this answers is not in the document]
toasted-nutbread: Makes sense to me.

ext/fg/js/frontend-initialize.js
siikamiika: https://github.com/FooSoft/yomichan/pull/439/commits/362ac84e6c6a0a4c2353232a5c44e1b2157de44d

toasted-nutbread:
Noticed an issue:

0. Open Chrome.
1. Open [test-document2.html](https://github.com/FooSoft/yomichan/blob/master/test/data/html/test-document2.html) in a new tab.
2. Open a popup by scanning content in test 1.
3. Open the options tab.
4. Change the value of **Maximum number of additional popups** or **Show iframe popups in root frame**.
5. test-document2.html hijacks the tab focus from the options tab.

toasted-nutbread:
Another issue I noticed, which may be a timing issue. Not as critical, but not sure how complex this would be to fix.

1. Change **Maximum number of additional popups** to some large value like 10.
2. Open a document, scan text, open all popups.
3. Change value of **Maximum number of additional popups** to 1. All popups will remain open.
4. Change value of **Maximum number of additional popups**  to 2. Most of the popups will now close.

The options will use the old value of `popupNestingMaxDepth`.

For testing, on line 290 I added:
```js
console.log('_canEnable', this.options, this.popup.depth, '<=', this.options !== null ? this.options.scanning.popupNestingMaxDepth : null, this._disabledOverride);
```

Logged this:
```
// Initially opening all popups
frontend.js:290 _canEnable null 0 <= null false
frontend.js:290 _canEnable null 1 <= null false
frontend.js:290 _canEnable null 2 <= null false
frontend.js:290 _canEnable null 3 <= null false
frontend.js:290 _canEnable null 4 <= null false
frontend.js:290 _canEnable null 5 <= null false
frontend.js:290 _canEnable null 6 <= null false
frontend.js:290 _canEnable null 7 <= null false
frontend.js:290 _canEnable null 8 <= null false
frontend.js:290 _canEnable null 9 <= null false
frontend.js:290 _canEnable null 10 <= null false
// Changed value of popupNestingMaxDepth to 1
frontend.js:290 _canEnable {…} 0 <= 10 false
frontend.js:290 _canEnable {…} 1 <= 10 false
frontend.js:290 _canEnable {…} 7 <= 10 false
frontend.js:290 _canEnable {…} 6 <= 10 false
frontend.js:290 _canEnable {…} 4 <= 10 false
frontend.js:290 _canEnable {…} 2 <= 10 false
frontend.js:290 _canEnable {…} 5 <= 10 false
frontend.js:290 _canEnable {…} 3 <= 10 false
frontend.js:290 _canEnable {…} 10 <= 10 false
frontend.js:290 _canEnable {…} 8 <= 10 false
frontend.js:290 _canEnable {…} 9 <= 10 false
// Changed value of popupNestingMaxDepth to 2
frontend.js:290 _canEnable {…} 0 <= 1 false
frontend.js:290 _canEnable {…} 1 <= 1 false
frontend.js:290 _canEnable {…} 7 <= 1 false
frontend.js:290 _canEnable {…} 6 <= 1 false
frontend.js:290 _canEnable {…} 4 <= 1 false
frontend.js:290 _canEnable {…} 2 <= 1 false
frontend.js:290 _canEnable {…} 5 <= 1 false
frontend.js:290 _canEnable {…} 3 <= 1 false
frontend.js:290 _canEnable {…} 10 <= 1 false
frontend.js:290 _canEnable {…} 8 <= 1 false
frontend.js:290 _canEnable {…} 9 <= 1 false
```

siikamiika:
> Noticed an issue:
> 
>     1. Open Chrome.
> 
>     2. Open [test-document2.html](https://github.com/FooSoft/yomichan/blob/master/test/data/html/test-document2.html) in a new tab.
> 
>     3. Open a popup by scanning content in test 1.
> 
>     4. Open the options tab.
> 
>     5. Change the value of **Maximum number of additional popups** or **Show iframe popups in root frame**.
> 
>     6. test-document2.html hijacks the tab focus from the options tab.
> 
> 
> Appears to have something to do with the iframes, since it doesn't occur on documents without iframes.

This happens when https://github.com/FooSoft/yomichan/pull/439/files#diff-c0a35793e5728d9d3e4b1301d23a4c3bR113-R115 is run because `Frontend.setPopup` calls `       this.onSearchClear(true);`. I can change it to `false`, but it could lead to hiding a popup so that it's still focused. I guess forcefully focusing the tab is worse. I remember we had a similar issue in the past, but it was fixed somehow.

siikamiika:
> Another issue I noticed, which may be a timing issue. Not as critical, but not sure how complex this would be to fix.
> 
>     1. Change **Maximum number of additional popups** to some large value like 10.
> 
>     2. Open a document, scan text, open all popups.
> 
>     3. Change value of **Maximum number of additional popups** to 1. All popups will remain open.
> 
>     4. Change value of **Maximum number of additional popups**  to 2. Most of the popups will now close.
> 
> 
> The options will use the old value of `popupNestingMaxDepth`.
> 
> For testing, on line 290 I added:
> 
> ```js
> console.log('_canEnable', this.options, this.popup.depth, '<=', this.options !== null ? this.options.scanning.popupNestingMaxDepth : null, this._disabledOverride);
> ```
> 
> Logged this:
> 
> ```
> // Initially opening all popups
> frontend.js:290 _canEnable null 0 <= null false
> frontend.js:290 _canEnable null 1 <= null false
> frontend.js:290 _canEnable null 2 <= null false
> frontend.js:290 _canEnable null 3 <= null false
> frontend.js:290 _canEnable null 4 <= null false
> frontend.js:290 _canEnable null 5 <= null false
> frontend.js:290 _canEnable null 6 <= null false
> frontend.js:290 _canEnable null 7 <= null false
> frontend.js:290 _canEnable null 8 <= null false
> frontend.js:290 _canEnable null 9 <= null false
> frontend.js:290 _canEnable null 10 <= null false
> // Changed value of popupNestingMaxDepth to 1
> frontend.js:290 _canEnable {…} 0 <= 10 false
> frontend.js:290 _canEnable {…} 1 <= 10 false
> frontend.js:290 _canEnable {…} 7 <= 10 false
> frontend.js:290 _canEnable {…} 6 <= 10 false
> frontend.js:290 _canEnable {…} 4 <= 10 false
> frontend.js:290 _canEnable {…} 2 <= 10 false
> frontend.js:290 _canEnable {…} 5 <= 10 false
> frontend.js:290 _canEnable {…} 3 <= 10 false
> frontend.js:290 _canEnable {…} 10 <= 10 false
> frontend.js:290 _canEnable {…} 8 <= 10 false
> frontend.js:290 _canEnable {…} 9 <= 10 false
> // Changed value of popupNestingMaxDepth to 2
> frontend.js:290 _canEnable {…} 0 <= 1 false
> frontend.js:290 _canEnable {…} 1 <= 1 false
> frontend.js:290 _canEnable {…} 7 <= 1 false
> frontend.js:290 _canEnable {…} 6 <= 1 false
> frontend.js:290 _canEnable {…} 4 <= 1 false
> frontend.js:290 _canEnable {…} 2 <= 1 false
> frontend.js:290 _canEnable {…} 5 <= 1 false
> frontend.js:290 _canEnable {…} 3 <= 1 false
> frontend.js:290 _canEnable {…} 10 <= 1 false
> frontend.js:290 _canEnable {…} 8 <= 1 false
> frontend.js:290 _canEnable {…} 9 <= 1 false
> ```

Caused by calling `_canEnable` before it got the newest options. Fixed in https://github.com/FooSoft/yomichan/pull/439/commits/b6f7f8c02637cef9cadd4ab2b9f253c42903aee6

PR #440, opened 2020-04-10
Add support for collapsing emphatic character sequences

### LINKED ISSUE #421: translate consecutive っ into a single one

toasted-nutbread:
Resolves #421.

This change adds a new translation option: **Collapse emphatic character sequences**. It has three possible values:
* "Disabled"
  かっっっこい => かっっっこい
* "Collapse into single characters"
  かっっっこい => かっこい
* "Remove all characters"
  かっっっこい => かこい

One special case that I decided to not handle initially is what to do with っ sequences at the start of the string. Currently, they still collapse, so っっっこい converts to っこい. This case can be changed if necessary.

Related question: should `ー` characters also be able to be collapsed in some way?

### DIFF
--- a/ext/bg/data/options-schema.json
+++ b/ext/bg/data/options-schema.json
@@ -388,7 +388,8 @@
                                     "convertNumericCharacters",
                                     "convertAlphabeticCharacters",
                                     "convertHiraganaToKatakana",
-                                    "convertKatakanaToHiragana"
+                                    "convertKatakanaToHiragana",
+                                    "collapseEmphaticSequences"
                                 ],
                                 "properties": {
                                     "convertHalfWidthCharacters": {
@@ -415,6 +416,11 @@
                                         "type": "string",
                                         "enum": ["false", "true", "variant"],
                                         "default": "variant"
+                                    },
+                                    "collapseEmphaticSequences": {
+                                        "type": "string",
+                                        "enum": ["false", "true", "full"],
+                                        "default": "false"
                                     }
                                 }
                             },
--- a/ext/bg/js/japanese.js
+++ b/ext/bg/js/japanese.js
@@ -85,6 +85,7 @@
 
     const HIRAGANA_SMALL_TSU_CODE_POINT = 0x3063;
     const KATAKANA_SMALL_TSU_CODE_POINT = 0x30c3;
+    const KANA_PROLONGED_SOUND_MARK_CODE_POINT = 0x30fc;
 
     // Existing functions
 
@@ -383,7 +384,11 @@
         const hasSourceMap = (sourceMap !== null);
         for (const char of text) {
             const c = char.codePointAt(0);
-            if (c === HIRAGANA_SMALL_TSU_CODE_POINT || c === KATAKANA_SMALL_TSU_CODE_POINT) {
+            if (
+                c === HIRAGANA_SMALL_TSU_CODE_POINT ||
+                c === KATAKANA_SMALL_TSU_CODE_POINT ||
+                c === KANA_PROLONGED_SOUND_MARK_CODE_POINT
+            ) {
                 if (collapseCodePoint !== c) {
                     collapseCodePoint = c;
                     if (!fullCollapse) {
--- a/ext/bg/js/japanese.js
+++ b/ext/bg/js/japanese.js
@@ -83,6 +83,8 @@
 
     const ITERATION_MARK_CODE_POINT = 0x3005;
 
+    const HIRAGANA_SMALL_TSU_CODE_POINT = 0x3063;
+    const KATAKANA_SMALL_TSU_CODE_POINT = 0x30c3;
 
     // Existing functions
 
@@ -373,6 +375,39 @@
     }
 
 
+    // Miscellaneous
+
+    function collapseEmphaticSequences(sourceText, fullCollapse, sourceMap=null) {
+        let result = '';
+        let collapseCodePoint = -1;
+        const hasSourceMap = (sourceMap !== null);
+        for (const char of sourceText) {
+            const c = char.codePointAt(0);
+            if (c === HIRAGANA_SMALL_TSU_CODE_POINT || c === KATAKANA_SMALL_TSU_CODE_POINT) {
+                if (collapseCodePoint !== c) {
+                    collapseCodePoint = c;
+                    if (!fullCollapse) {
+                        result += char;
+                        continue;
+                    }
+                }
+            } else {
+                collapseCodePoint = -1;
+                result += char;
+                continue;
+            }
+
+            if (hasSourceMap) {
+                const index = result.length;
+                if (index > 0) {
+                    sourceMap.combine(index - 1, 1);
+                }
+            }
+        }
+        return result;
+    }
+
+
     // Exports
 
     Object.assign(jp, {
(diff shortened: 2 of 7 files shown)


siikamiika:
> Related question: should `ー` characters also be able to be collapsed in some way?

I think any character repeated more than 2 times almost always indicates emphasis, and when the preceding kana already contains the repeated vowel, 1 is enough:

- さあああっぱりわかんない
- そおおおだよ
- せえええの
- いいいんだぞ
- weeeeell (to illustrate this in English)

toasted-nutbread:
Of course, but I guess I'm wondering how often some of these are used in practice / would they be worth it to incorporate. I mainly mentioned `ー` since the algorithm would be the same as `っ`/`ッ`. Kana repetition, while doable, would be a slightly different algorithm (have to check the vowel of previous mora) and may be more likely to have false positives.

siikamiika:
From own experience you'll see both repeated characters and long vowel marks even mixed together, but I agree that long vowel marks would have less false positives, for example in cases where two separate words start and end with the same vowel (それええな).

A kanji followed by a long vowel mark will mean two or three repeated vowels, but without a text analyzer like MeCab or dictionary lookup it's not possible to know the reading of the kanji to do the same translation for the following kana. This might not be a big problem, because for example [長ーい](https://www.google.com/search?q=%22%E9%95%B7%E3%83%BC%E3%81%84%22) is a much more common spelling than [長あい](https://www.google.com/search?q=%22%E9%95%B7%E3%81%82%E3%81%84%22).

Looking at other Google results, when no kanji is involved, long vowel mark is also more comon with varying degrees:
- ["さああっぱり"](https://www.google.com/search?q=%22%E3%81%95%E3%81%82%E3%81%82%E3%81%A3%E3%81%B1%E3%82%8A%22) vs. ["さあーっぱり"](https://www.google.com/search?q=%22%E3%81%95%E3%81%82%E3%83%BC%E3%81%A3%E3%81%B1%E3%82%8A%22)
- [すごおい](https://www.google.com/search?q=%22%E3%81%99%E3%81%94%E3%81%8A%E3%81%84%22) vs. [すごーい](https://www.google.com/search?q=%22%E3%81%99%E3%81%94%E3%83%BC%E3%81%84%22)

toasted-nutbread:
I think for simplicity, I will add support for the `ー` character since it would be treated the identically to the current characters, and we can add support for repeated kana in the future if necessary.

Any thoughts on this?
> One special case that I decided to not handle initially is what to do with っ sequences at the start of the string. Currently, they still collapse, so っっっこい converts to っこい. This case can be changed if necessary.

siikamiika:
> Any thoughts on this?
> 
> > One special case that I decided to not handle initially is what to do with っ sequences at the start of the string. Currently, they still collapse, so っっっこい converts to っこい. This case can be changed if necessary.

I think it makes sense to collapse them as well, since there are some endings that are used with or without the sokuon such as こい/っこい and ぽい/っぽい. Some dictionaries might not include the emphatic version. It can result in more false positives, though, unless some kind of sorting based on original orthography, kana only results or "usually kana" tag is implemented.

toasted-nutbread:
> I think it makes sense to collapse them as well, since there are some endings that are used with or without the sokuon such as こい/っこい and ぽい/っぽい. Some dictionaries might not include the emphatic version. It can result in more false positives, though (濃い, 鯉, 故意, 恋), unless some kind of sorting based on original orthography, kana only results or "usually kana" tag is implemented.

I will leave it as in then.

`ー` support has been added in 90392ac9d6d3b54f811e3d056043a1ffe26fa963. Amusingly your search results link for [すごーい](https://www.google.com/search?q=%22%E3%81%99%E3%81%94%E3%83%BC%E3%81%84%22) had some results with `すっごーい`, which can now collapse both `っ` and `ー` to get `すごい`, so that's a good test case. I'll add some tests for the new character too.

siikamiika:
I think it would be useful if both collapsing into a single character and removing all characters were supported simultaneously. Could also update the example label to use すっっごーーい → すっごーい / すごい to illustrate collapsing both sokuon and chouon.

siikamiika:
I also noticed another pattern that's commonly used for emphasis, but that's probably too complex/rare/risky to include for the same translation: [さぁーっぱり](https://www.google.com/search?q=%22%E3%81%95%E3%81%81%E3%83%BC%E3%81%A3%E3%81%B1%E3%82%8A%22). In that example the small ぁ could be collapsed together with the long vowel marks.

Maybe there could be yet another translation for suppressing small ぁぃぅぇぉ and converting them to their regular counterparts. They have legit uses in many loanwords, but are also used for emphasis. Another example where it would be useful to convert it to uppercase: [さぁな](https://www.google.com/search?q=%22%E3%81%95%E3%81%81%E3%81%AA%22)

toasted-nutbread:
> I think it would be useful if both collapsing into a single character and removing all characters were supported simultaneously. Could also update the example label to use すっっごーーい → すっごーい / すごい to illustrate collapsing both sokuon and chouon.

They are supported simultaneously, it's just more wordy to make that very explicit. When removal is enabled, it uses variants for uncollapsed, collapsed (single char), and removed. I agree about changing the example, I'll do that.

> I also noticed another pattern that's commonly used for emphasis, but that's probably too complex/rare/risky to include for the same translation: [さぁーっぱり](https://www.google.com/search?q=%22%E3%81%95%E3%81%81%E3%83%BC%E3%81%A3%E3%81%B1%E3%82%8A%22). In that example the small ぁ could be collapsed together with the long vowel marks.
> 
> Maybe there could be yet another translation for suppressing small ぁぃぅぇぉ and converting them to their regular counterparts. They have legit uses in many loanwords, but are also used for emphasis. Another example where it would be useful to convert it to uppercase: [さぁな](https://www.google.com/search?q=%22%E3%81%95%E3%81%81%E3%81%AA%22)

Yeah I've seen that as well. Since the implementation of anything involving regular/small kana changes the current algorithm, I think it would be best to do those separately.

siikamiika:
> > I think it would be useful if both collapsing into a single character and removing all characters were supported simultaneously. Could also update the example label to use すっっごーーい → すっごーい / すごい to illustrate collapsing both sokuon and chouon.
> 
> They are supported simultaneously, it's just more wordy to make that very explicit. When removal is enabled, it uses variants for uncollapsed, collapsed (single char), and removed. I agree about changing the example, I'll do that.

Are you sure? I tested with the "full" option, and it doesn't find かっっこいい although explicit かっこいい works. With "true", both かっっこいい and かっこいい work. I think you'd have to add both `[true, false]` and `[true, true]` for "full" to make them work simultaneously. I was thinking of a separate option, though, but both work for me.

toasted-nutbread:
> Are you sure? I tested with the "full" option, and it doesn't find かっっこいい although explicit かっこいい works. With "true", both かっっこいい and かっこいい work. I think you'd have to add both `[true, false]` and `[true, true]` for "full" to make them work simultaneously. I was thinking of a separate option, though, but both work for me.

Oops, you're right. I _intended_ to make it so that they were both enabled when using `full`, but looks like I changed/forgot something. Will update.

PR #443, opened 2020-04-10
Fix getSelection not working in QueryParser

toasted-nutbread:
`getSelection` only works on window, not on DOM nodes.

### DIFF
--- a/ext/bg/js/search.js
+++ b/ext/bg/js/search.js
@@ -209,7 +209,7 @@ class DisplaySearch extends Display {
 
     onCopy() {
         // ignore copy from search page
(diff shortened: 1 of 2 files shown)


toasted-nutbread:
Seems to be the difference between the following:
https://developer.mozilla.org/en-US/docs/Web/API/Window/getSelection
https://developer.mozilla.org/en-US/docs/Web/API/DocumentOrShadowRoot/getSelection

It looks like the `document` one is actually not supported by the minimum Firefox version (v52); should I update that to use `window` while we're at it?

siikamiika:
Yeah, would be nice :+1:

PR #448, opened 2020-04-12
Badge improvements

toasted-nutbread:
Some improvements to the status reporting functionality on the browser badge. This satisfies the following task from #270:

> Indicate when the backend / database is upgrading. When the database version is upgrading, the extension won't run. This operation can take multiple seconds, so it can be noticeable. This was the case in the version upgrade in #287, despite the difference being very small.

When the backend is `prepare()`ing, the status will show as "..." if it takes longer than 1 second. Additionally, if any error occurs during `prepare()`, this will be logged and the status will change to a red "!!". Additionally, the hover tooltip text is changed to reflect this status.

Prepare errors don't/shouldn't currently happen in typical use cases, but it can in development workflows. For example, when switching to a branch which upgraded the database version and then switching branches to an older version, the `database.prepare()` will fail.

### DIFF
--- a/ext/bg/js/backend.js
+++ b/ext/bg/js/backend.js
@@ -77,6 +77,7 @@ class Backend {
 
         this._defaultBrowserActionTitle = null;
         this._isPrepared = false;
+        this._badgePrepareDelayTimer = null;
 
         this._messageHandlers = new Map([
             ['yomichanCoreReady', {handler: this._onApiYomichanCoreReady.bind(this), async: false}],
@@ -123,6 +124,10 @@ class Backend {
 
     async prepare() {
         this._defaultBrowserActionTitle = await this._getBrowserIconTitle();
+        this._badgePrepareDelayTimer = setTimeout(() => {
+            this._badgePrepareDelayTimer = null;
+            this._updateBadge();
+        }, 1000);
         this._updateBadge();
         await this.database.prepare();
         await this.translator.prepare();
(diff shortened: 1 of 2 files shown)


ext/bg/js/backend.js
```diff
@@ -804,6 +788,76 @@ class Backend {
...
     // Utilities
 
+    _getBrowserIconTitle() {
+        return (
+            chrome.browserAction !== null &&
+            typeof chrome.browserAction === 'object' &&
+            typeof chrome.browserAction.getTitle === 'function' ?
+            new Promise((resolve) => chrome.browserAction.getTitle({}, resolve)) :
+            Promise.resolve('')
+        );
```
siikamiika, line 812: The ternary should probably be indented for the parts after `?` or the condition could be stored in a constant.

ext/bg/js/backend.js
```diff
@@ -800,6 +802,82 @@ class Backend {
...
+    }
+
+    _updateBadge() {
+        let title = this._defaultBrowserActionTitle;
+        if (
+            title === null ||
+            chrome.browserAction === null ||
+            typeof chrome.browserAction !== 'object'
```
siikamiika, line 820: This could use `!isObject(chrome.browserAction)`. Also applies to the earlier comment about `_getBrowserIconTitle`

toasted-nutbread: 0628c333d9b0fc15aeb2ec8cdead9a8dff4fb087

ext/bg/js/backend.js
toasted-nutbread: 0628c333d9b0fc15aeb2ec8cdead9a8dff4fb087

PR #449, opened 2020-04-12
Fix default value of _orphaned

toasted-nutbread:
Oops


siikamiika:
I actually got orphaned popup instead of a proper error in the past when debugging, but didn't realize that there was a bug with the `_orphaned` value :smile:

PR #450, opened 2020-04-12
Add support for filtering frequency metadata based on readings

### LINKED ISSUE #382: Support for frequency lists with readings
Is it possible to import frequency list similar to:
```
[["方","ホウ",1],
["方","カタ",2],
["明日","アス",3]
["明日","アシタ",4]]
```
So different readings of the compound words would show different frequency information.
If it's impossible for now, that's my "feature request"

toasted-nutbread:
Resolves #382. Data format is as described in https://github.com/FooSoft/yomichan/issues/382#issuecomment-593178227.

### DIFF
--- a/ext/bg/js/translator.js
+++ b/ext/bg/js/translator.js
@@ -469,7 +469,9 @@ class Translator {
             switch (mode) {
                 case 'freq':
                     for (const term of termsUnique[index]) {
-                        term.frequencies.push({expression, frequency: data, dictionary});
+                        const frequencyData = this.getFrequencyData(expression, data, dictionary, term);
+                        if (frequencyData === null) { continue; }
+                        term.frequencies.push(frequencyData);
                     }
                     break;
                 case 'pitch':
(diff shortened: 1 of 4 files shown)


siikamiika:
Is the term metadata going to be imported into the database even if you switch `data` into an object? That part seems to be implicit to this code.

toasted-nutbread:
Yeah, that is the intent. There is no conversion since the data is already validated by the schema and is in the expected format.

siikamiika:
Ok, I was wondering if the DB schema had to be updated but seems that it supports storing objects as well.

ext/bg/js/translator.js
```diff
@@ -562,6 +564,18 @@ class Translator {
         return tagMetaList;
     }
 
+    getFrequencyData(expression, data, dictionary, term) {
+        if (data !== null && typeof data === 'object') {
```
siikamiika, line 568: Could use isObject as well

toasted-nutbread:
> Ok, I was wondering if the DB schema had to be updated but seems that it supports storing objects as well.

Yeah, it's not an indexed property currently, but even [indexes should be able to  store any type of data](https://developer.mozilla.org/en-US/docs/Web/API/IDBObjectStore/createIndex):

> Bear in mind that IndexedDB indexes can contain any JavaScript data type; IndexedDB uses the structured clone algorithm to serialize stored objects, which allows for storage of simple and complex objects.

ext/bg/js/translator.js
toasted-nutbread: This one I think is fine to leave as-is; the `Array.isArray` check is not necessary for this case since it should have been validated during import. Additionally, even if somehow the value was an array, it won't be any more wrong than having an incorrectly formatted object.

PR #453, opened 2020-04-12
Reuse AnkiConnect instance

toasted-nutbread:
Resolves https://github.com/FooSoft/yomichan/pull/427#issuecomment-609493372

### DIFF
--- a/ext/bg/js/anki.js
+++ b/ext/bg/js/anki.js
@@ -25,82 +25,84 @@
 
 class AnkiConnect {
     constructor(server) {
-        this.server = server;
-        this.localVersion = 2;
-        this.remoteVersion = 0;
+        this._server = server;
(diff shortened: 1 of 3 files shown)


ext/bg/js/anki.js
```diff
@@ -19,122 +19,123 @@
...
+        }
+    }
+
+    async _invoke(action, params) {
+        const result = await requestJson(this._server, 'POST', {action, params, version: this._localVersion});
+        if (
+            result !== null &&
+            typeof result === 'object' &&
+            !Array.isArray(result)
+        ) {
```
siikamiika, line 119: Can be replaced with `isObject(result)`

toasted-nutbread: 018913d03fff627fb7d34f594340c47a607e8839

Issue #418, opened 2020-03-21
Move QueryParser.parseText to backend


siikamiika:
This is kind of unrelated, but since this code touched it, I'm wondering if `QueryParser.parseText` should be moved into the backend. IMO checking whether or not to do parsing, particularly when checking mecab, seems like an implementation detail which isn't necessary for the "frontend" to know.

If this is the case, the data returned might have to be formatted differently in order to make it more generalized, but it wouldn't be a big difference. For example:

```js
// something like this
results.push({
    source: 'scanning-parser',
    content: results
});
results.push({
    source: 'mecab',
    dictionary: mecabDictName,
    content: mecabDictResults
});

// instead of the following
// the name/id are for presentation/options purposes
// parsedText may also be a misleading name since the data is an array of objects and not a string
results.push({
    name: 'Scanning parser',
    id: 'scan',
    parsedText: await apiTextParse(text, this.getOptionsContext())
});
results.push({
    name: `MeCab: ${mecabDictName}`,
    id: `mecab-${mecabDictName}`,
    parsedText: mecabDictResults
});
```

_Originally posted by @toasted-nutbread in https://github.com/FooSoft/yomichan/pull/411#issuecomment-599161575_

PR #456, opened 2020-04-13
Parse text refactor

siikamiika:
Fixes https://github.com/FooSoft/yomichan/issues/418

### DIFF
--- a/ext/bg/js/backend.js
+++ b/ext/bg/js/backend.js
@@ -85,7 +85,6 @@ class Backend {
             ['kanjiFind', {handler: this._onApiKanjiFind.bind(this), async: true}],
             ['termsFind', {handler: this._onApiTermsFind.bind(this), async: true}],
             ['textParse', {handler: this._onApiTextParse.bind(this), async: true}],
-            ['textParseMecab', {handler: this._onApiTextParseMecab.bind(this), async: true}],
             ['definitionAdd', {handler: this._onApiDefinitionAdd.bind(this), async: true}],
             ['definitionsAddable', {handler: this._onApiDefinitionsAddable.bind(this), async: true}],
             ['noteView', {handler: this._onApiNoteView.bind(this), async: true}],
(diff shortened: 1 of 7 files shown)


ext/bg/js/search-query-parser-generator.js
```diff
@@ -71,7 +71,16 @@ class QueryParserGenerator {
...
-            optionContainer.textContent = parseResult.name;
+            switch (parseResult.source) {
+                case 'scanning-parser':
+                    optionContainer.textContent = 'Scanning parser';
+                    break;
+                case 'mecab':
+                    optionContainer.textContent = `MeCab: ${parseResult.dictionary}`;
+                    break;
+                default:
+                    optionContainer.textContent = 'Unrecognized dictionary';
```
toasted-nutbread, line 82: break

ext/bg/js/backend.js
```diff
@@ -315,6 +314,61 @@ class Backend {
...
+        const results = [];
+        while (text.length > 0) {
+            const term = [];
+            const [definitions, sourceLength] = await this.translator.findTerms(
+                'simple',
+                text.substring(0, options.scanning.length),
+                {},
+                options
+            );
+            if (definitions.length > 0) {
```
toasted-nutbread, line 327: `if (definitions.length > 0 && sourceLength > 0) {`

Explicitly prevent infinite loops. I _hope_ this case should never happen, but better safe than sorry.

ext/bg/js/backend.js
```diff
@@ -315,6 +314,61 @@ class Backend {
...
+        return results;
+    }
+
+    async _textParseMecab(text, options) {
+        const results = [];
+        const rawResults = await this.mecab.parseText(text);
+        for (const [mecabName, parsedLines] of Object.entries(rawResults)) {
+            const result = [];
+            for (const parsedLine of parsedLines) {
+                for (let {expression, reading, source} of parsedLine) {
```
toasted-nutbread, line 352: const
`for (const {expression, reading, source} of parsedLine) {`

Then just do

```js
jp.distributeFuriganaInflected(
    expression.length > 0 ? expression : source,
    // ...
```

(Since `expression` only appears to be used once)

ext/bg/js/search-query-parser-generator.js
toasted-nutbread: (I'm used to some other languages requiring it; unfortunately [not enforceable](https://github.com/eslint/eslint/issues/7697).)

siikamiika: Yeah, I think it's good practice as well although not required by the language. I actually like trailing commas as well for the same reason, but we've opted to not use them.

siikamiika: https://github.com/FooSoft/yomichan/pull/456/commits/a02cf28c5775bdd72f209b46e9974768e5c5db71

ext/bg/js/backend.js
siikamiika: https://github.com/FooSoft/yomichan/pull/456/commits/3fa8b8e1866933fdfccd0598a11e443fc1693840

ext/bg/js/backend.js
siikamiika: https://github.com/FooSoft/yomichan/pull/456/commits/619df42aedaa8da97d0a16d539b7211349143a0a

PR #458, opened 2020-04-17
DOMTextScanner

toasted-nutbread:
I am planning on redoing #286 in a more modular manner, and this is the first part. The ability to scan text in a document can be decoupled from the concept of `TextSourceRange` itself. `TextSourceRange` would eventually use this class, but due to the algorithm complexity of scanning the text, I feel it is best to isolate it as much as possible.

This PR does not solve the issue I originally mentioned in https://github.com/FooSoft/yomichan/pull/286#issuecomment-562902190, but it opens the door to supporting both functionalities with an option to toggle between them.

### Public API

The new `DOMTextScanner` class has the following public API (excluding static helpers):
```js
DOMTextScanner {
    constructor(node, offset, forcePreserveWhitespace=false, generateLayoutContent=true);
    seek(length);
    get node;
    get offset;
    get content;
}
```

* `seek()` is intended to be able to be used repeatedly on the same instance and should yield consistent results, so long as the `length` value passed has a consistent sign. This means that `seek(1).seek(1).seek(1) ... .seek(1)` should always yield the same state as `seek(N)`.
* `forcePreserveWhitespace` and `generateLayoutContent` are used to control the behaviour of whitespace handling. The default values of `[false, true]` are intended to match the new functionality initially added in #286, and using the values of `[true, false]` should give the current functionality.
  * `forcePreserveWhitespace` changes how whitespace is handled. When `true`, whitespace is not treated in a special manner and is naively appended to the content string. When `false`, the CSS `white-space` property is evaluated to determine how to handle the characters. Whitespace and newline character collapsing is handled according the the table specified in [MDN's white-space docs](https://developer.mozilla.org/en-US/docs/Web/CSS/white-space).
  * `generateLayoutContent` controls whether or not to add newlines to the content string based on DOM layout. For example, `<div>line1</div><div>line2</div>` will have a line separation, but just reading the [`nodeValue`](https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeValue) of a text node will not indicate this.

### Implementation

Internal to the class, there are two functions which handle text scanning. These are `_seekTextNodeForward` and `_seekTextNodeBackward`. `_seekTextNodeBackward` is largely a copy-paste of `_seekTextNodeForward`, but works in the opposite direction.

Both of these functions use a technique of copying the object's state into locally scoped variables at the start of the function, and then copying it back at the end. This is intended to help the compiler optimize the variables which are used frequently in the loop, potentially allowing them to be stored in registers. I haven't done profiling comparisons as of yet.

Not sure what exactly this is called off the top of my head, but I've seen this described a few times. The general idea is that `for (let i = 0; i < N; ++i)` is easier to optimize by the compiler than `for (this.i = 0; this.i < N; ++this.i)`. [Here is a reference I've seen, see **Locals instead of fields**](https://roman.st/Article/Faster-Marsaglia-Xorshift-pseudo-random-generator-in-unsafe-C). Not the same language, but the same concept.

### Tests and Documentation

Several tests have been added to test the new functionality, and the API includes documentation comments.

### DIFF
--- a/ext/fg/js/dom-text-scanner.js
+++ b/ext/fg/js/dom-text-scanner.js
@@ -1,0 +1,538 @@
+/*
+ * Copyright (C) 2020  Yomichan Authors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+/**
+ * A class used to scan text in a document.
+ */
+class DOMTextScanner {
+    /**
+     * Creates a new instance of a DOMTextScanner.
+     * @param node The DOM Node to start at.
+     * @param offset The character offset in to start at when node is a text node.
+     *   Use 0 for non-text nodes.
+     */
+    constructor(node, offset, forcePreserveWhitespace=false, generateLayoutContent=true) {
+        const ruby = DOMTextScanner.getParentRubyElement(node);
+        const resetOffset = (ruby !== null);
+        if (resetOffset) { node = ruby; }
+
+        this._node = node;
+        this._offset = offset;
+        this._content = '';
+        this._remainder = 0;
+        this._resetOffset = resetOffset;
+        this._newlines = 0;
+        this._lineHasWhitespace = false;
+        this._lineHasContent = false;
+        this._forcePreserveWhitespace = forcePreserveWhitespace;
+        this._generateLayoutContent = generateLayoutContent;
+    }
+
+    /**
+     * Gets the current node being scanned.
+     * @returns A DOM Node.
+     */
+    get node() {
+        return this._node;
+    }
+
+    /**
+     * Gets the current offset corresponding to the node being scanned.
+     * This value is only applicable for text nodes.
+     * @returns An integer.
+     */
+    get offset() {
+        return this._offset;
+    }
+
+    /**
+     * Gets the accumulated content string resulting from calls to seek().
+     * @returns A string.
+     */
+    get content() {
+        return this._content;
+    }
+
+    /**
+     * Seeks a given length in the document and accumulates the text content.
+     * @param length A positive or negative integer corresponding to how many characters
+     *   should be added to content. Content is only added to the accumulation string,
+     *   never removed, so mixing seek calls with differently signed length values
+     *   may give unexpected results.
+     * @returns this
+     */
+    seek(length) {
+        const forward = (length >= 0);
+        this._remainder = (forward ? length : -length);
+        if (length === 0) { return this; }
+
+        const TEXT_NODE = Node.TEXT_NODE;
+        const ELEMENT_NODE = Node.ELEMENT_NODE;
+
+        const generateLayoutContent = this._generateLayoutContent;
+        let node = this._node;
+        let resetOffset = this._resetOffset;
+        let newlines = 0;
+        while (node !== null) {
+            let enterable = false;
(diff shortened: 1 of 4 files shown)


siikamiika:
I'll come back to this with some more time later (during the week or next weekend). Seems very useful at a glance.

toasted-nutbread:
No problem, I'm not in a big hurry on this. Just want to eventually capture the goals targeted in #286, and as I revisited that branch, it seemed like it could be done better.

siikamiika:
>  * `seek()` is intended to be able to be used repeatedly on the same instance and should yield consistent results, so long as the `length` value passed has a consistent sign. This means that `seek(1).seek(1).seek(1) ... .seek(1)` should always yield the same state as `seek(N)`.

If I understand the scope of this PR correctly, this is about to get more changes after this stage, but after reading the code I think it would be possible to have a head and tail version of the node and the accumulated text, and also an internal text position pointer that can be positive or negative. This way it would be possible to seek back and forward and either choose the appropriate element or do nothing if that part is already cached. Example:

```javascript
// Page content:
// コンピュータにおけるファイル（英: file）は、コンピュータでデータを扱いやすくするための仕組み。
// scan position: after "は、"

const scanner = new DOMTextScanner(...);

// scans from after 、to the beginning using the tail node
scanner.seek(-25); // = -25

// outputs the whole sentence above by concatenating the
// tail content of 25 characters and
// scanning 25 more using the head node,
// also storing the content for future seeks or reads
scanner.read(50);

// attempt to read past the end of the page content
scanner.read(1); // = null or '' or exception

// attempt to seek past the end of the page content
scanner.seek(1); // = 0
```

Alternatively you could also have two `DOMTextScanner`s inside another abstraction.

There's a problem not limited to the new `DOMTextScanner` class that the current node can become out of date. I used `Node.isConnected` in https://github.com/FooSoft/yomichan/pull/460, but that doesn't help if the node was just moved or something was removed in between. I wonder if it would be possible to cache all the nodes in the order they're found and check that the neighboring nodes are still the same. This could be used each time a `DOMTextScanner` is reused after leaving it unused for a while.

> Both of these functions use a technique of copying the object's state into locally scoped variables at the start of the function, and then copying it back at the end. This is intended to help the compiler optimize the variables which are used frequently in the loop, potentially allowing them to be stored in registers. I haven't done profiling comparisons as of yet.

Would be interesting to see how browsers perform with each case, this could be simple enough to optimize to the same code but at least Firefox uses different execution tiers depending on the calling frequency of functions: https://hacks.mozilla.org/2019/08/the-baseline-interpreter-a-faster-js-interpreter-in-firefox-70/

toasted-nutbread:
The reason why it's expected to only use them in one direction and not do something like `scanner.seek(-25); scanner.read(50);` is that the "50" value in that case is misleading. It wouldn't be reading 50 character from the seek point after moving -25; it would be reading 25 in either direction from the start point. There are some cases where this is not the same, e.g. when starting scanning inside a node that can be skipped.

```html
<span id="test1">abc<span id="test2" style="display: none;">def123</span>456XXXXXX</span>
```
```js
const test1 = document.querySelector('#test1');
const test2 = document.querySelector('#test2');

const x = (
    new DOMTextScanner(test2.firstChild, 3).seek(-6) +
    new DOMTextScanner(test2.firstChild, 3).seek(6)
); // 'abcdef123456'
const y = new DOMTextScanner(test1.firstChild, 0).seek(12); // 'abc456XXXXXX'
const z = new DOMTextScanner(test1.lastChild, 9).seek(-12); // 'abc456XXXXXX'
```

siikamiika:
Going to get back to this later too, sorry for the delays. I didn't realize that starting from skippable nodes would have issues. My idea was that the API would be so simple that you wouldn't have to think about DOM scanning at all when using it, but it seems that it might be too much work for the small benefit.

toasted-nutbread:
> Going to get back to this later too, sorry for the delays. I didn't realize that starting from skippable nodes would have issues. My idea was that the API would be so simple that you wouldn't have to think about DOM scanning at all when using it, but it seems that it might be too much work for the small benefit.

I've largely tried to make this class such that it can easily replace some of the existing code which does scanning without having to write new code, so that's why the API is somewhat familiar and "scanning-aware". It is also intended to be flexible enough to do some potentially useful changes with it in the future. For example, changing the sentence scanning algorithm a bit, since it currently always scans 200 characters and _then_ filters that text down to the actual sentence.

Issue #432, opened 2020-04-05
"is fullscreen" profile condition (or special case)


siikamiika:
Due to the issue noted in https://github.com/FooSoft/yomichan/pull/417#issuecomment-600945163, showing iframe popups in the root frame has to be made conditional in order to make it work when the iframe is in fullscreen. It can be a special case for this option, but it could also be a separate profile condition if there's a use case for that.

Depends on https://github.com/FooSoft/yomichan/issues/431.

toasted-nutbread:
May also be required for iframes inside of shadow DOMs. Test 4 in [test-document2.html](https://github.com/FooSoft/yomichan/blob/a4c7d243f4cddfc4e2b29132fe65fe07ff86bc36/test/data/html/test-document2.html) isn't positioned correctly. The fallback for both of these cases could be to just use the old method of having the popup inside the iframe, but this will add some complexity since the popup/proxy will have to handle both cases.

PR #460, opened 2020-04-17
Iframe popup edge cases

siikamiika:
Support the following edge cases:

- Iframe is fullscreen: create a popup inside the iframe regardless of `options.general.showIframePopupsInRootFrame` value, but return it back when fullscreen is closed
- Iframe in open shadow dom: find the content window recursively and support positioning normally
- Iframe in closed shadow dom: fall back to popups inside iframe

Fixes https://github.com/FooSoft/yomichan/issues/432

### DIFF
--- a/ext/fg/js/frame-offset-forwarder.js
+++ b/ext/fg/js/frame-offset-forwarder.js
@@ -81,23 +81,9 @@ class FrameOffsetForwarder {
         }
 
         if (sourceFrame === null) {
-            const getShadowRootElements = (documentOrElement) => {
-                const elements = Array.from(documentOrElement.querySelectorAll('*'))
-                    .filter((el) => !!el.shadowRoot);
-                const childElements = elements
-                    .map((el) => el.shadowRoot)
-                    .map(getShadowRootElements);
-                elements.push(childElements.flat());
-
-                return elements.flat();
-            };
-
-            sourceFrame = getShadowRootElements(document)
-                .map((el) => Array.from(el.shadowRoot.querySelectorAll('frame, iframe:not(.yomichan-float)')))
-                .flat()
-                .find((el) => el.contentWindow === e.source);
-
+            sourceFrame = this._getOpenShadowRootSourceFrame(e.source);
             if (!sourceFrame) {
+                // closed shadow root etc.
                 this._forwardFrameOffsetOrigin(null, uniqueId);
                 return;
             }
@@ -108,6 +94,24 @@ class FrameOffsetForwarder {
         offset = [forwardedX + x, forwardedY + y];
 
         this._forwardFrameOffset(offset, uniqueId);
+    }
+
+    _getOpenShadowRootSourceFrame(sourceWindow) {
+        const getShadowRootElements = (documentOrElement) => {
+            const elements = Array.from(documentOrElement.querySelectorAll('*'))
+                .filter((el) => !!el.shadowRoot);
+            const childElements = elements
+                .map((el) => el.shadowRoot)
+                .map(getShadowRootElements);
+            elements.push(childElements.flat());
+
+            return elements.flat();
+        };
+
+        return getShadowRootElements(document)
+            .map((el) => Array.from(el.shadowRoot.querySelectorAll('frame, iframe:not(.yomichan-float)')))
+            .flat()
+            .find((el) => el.contentWindow === sourceWindow);
     }
 
     _forwardFrameOffsetParent(offset, uniqueId) {
--- a/ext/fg/js/frame-offset-forwarder.js
+++ b/ext/fg/js/frame-offset-forwarder.js
@@ -79,9 +79,28 @@ class FrameOffsetForwarder {
             sourceFrame = frame;
             break;
         }
+
         if (sourceFrame === null) {
-            this._forwardFrameOffsetOrigin(null, uniqueId);
-            return;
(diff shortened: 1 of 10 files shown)


ext/fg/js/frame-offset-forwarder.js
```diff
@@ -91,6 +96,24 @@ class FrameOffsetForwarder {
         this._forwardFrameOffset(offset, uniqueId);
     }
 
+    _getOpenShadowRootSourceFrame(sourceWindow) {
```
toasted-nutbread, line 99: This method is likely to be very expensive, particularly on complex pages. Consider something like this which can skip many operations once the result is found:

```js
// use document.documentElement as startElement
function findFrameWithContentWindow(startElement, contentWindow) {
  const ELEMENT_NODE = Node.ELEMENT_NODE;
  const elements = [startElement];
  while (elements.length > 0) {
    const element = elements.shift();
    switch (element.nodeName.toUpperCase()) { // toUpperCase potentially costly
      case 'IFRAME':
      case 'FRAME':
        if (element.contentWindow === contentWindow) {
          return element;
        }
        break;
    }

    const shadowRoot = element.shadowRoot;
    if (shadowRoot) {
      for (const child of shadowRoot.children) {
        if (child.nodeType === ELEMENT_NODE) {
          elements.push(child);
        }
      }
    }
    
    for (const child of element.children) {
      if (child.nodeType === ELEMENT_NODE) {
        elements.push(child);
      }
    }
  }
}
```

ext/fg/js/frontend-initialize.js
```diff
@@ -88,8 +95,8 @@ async function main() {
         }
 
         let popup;
-        if (isIframe && options.general.showIframePopupsInRootFrame) {
-            popup = popups.iframe || await createIframePopupProxy(url, frameOffsetForwarder);
+        if (isIframe && options.general.showIframePopupsInRootFrame && !document.fullscreen && iframePopupsInRootFrameAvailable) {
```
toasted-nutbread, line 98: `document.fullscreen` "is no longer recommended](https://developer.mozilla.org/en-US/docs/Web/API/Document/fullscreen). Would `Popup._getFullscreenElement`'s implementation work?

siikamiika: https://github.com/FooSoft/yomichan/pull/460/commits/1f071df50e0363a7e52295512a1019d99c4c8434

ext/fg/js/frame-offset-forwarder.js
siikamiika: Yeah, it's a naive approach and basically what I ended up with after testing in the browser console. However, I think this is very rare in the wild, and you only end up there if a scan actually happens in an iframe that's in shadow dom (or potentially some other unknown case), plus it's throttled to max 1 per second. I can try to optimize it a bit to help with the worst case.

siikamiika: https://github.com/FooSoft/yomichan/pull/460/commits/859c8480bd330502ed1ba91067c023c28139c5a1

I can also add some better data structure for the cache to avoid memory leaks if needed. I wonder if it would also need some error handling when the frames are removed from DOM.

toasted-nutbread: I doubt the cache will really be a significant cause for memory leaks unless some website is deliberately doing something stupid, but maybe a simple cache size limit could be used when adding a new item.

```js
function addSet(set, value, limit) {
  limit -= set.size;
  if (limit >= 0) {
    for (const value2 of set) {
      set.delete(value2); // This appears to be valid to do during iteration
      if (limit <= 0) { break; }
      --limit;
    }
  }
  set.add(value);
}
```

> I wonder if it would also need some error handling when the frames are removed from DOM.

Just check `document.documentElement.contains(element)` before returning, and probably remove from caches.

ext/fg/js/frame-offset-forwarder.js
```diff
@@ -91,6 +93,53 @@ class FrameOffsetForwarder {
         this._forwardFrameOffset(offset, uniqueId);
     }
 
+    _findFrameWithContentWindow(contentWindow) {
+        const elementSources = [
```
toasted-nutbread, line 97: 97-116 can be simplified by using a generator.

```js
function* getElementSources() {
  yield [...this._frameCache];
  yield [...document.querySelectorAll('frame, iframe:not(.yomichan-float)')];
  yield [document.documentElement];
}

const ELEMENT_NODE = Node.ELEMENT_NODE;
for (const elements of getElementSources()) {
  while (elements.length > 0 || elements.push(...getMoreElements())) {
    // ...
```

siikamiika: https://github.com/FooSoft/yomichan/pull/460/commits/08ded3154f45d23687e310800b6b7b4eb2133aef

ext/fg/js/frame-offset-forwarder.js
siikamiika: https://github.com/FooSoft/yomichan/pull/460/commits/8b4d5473d33256f4e050c0f85af3fcd4479b8764

It's a nice property of `Set` that the elements preserve insertion order. I was thinking of creating a parallel `Array` to remove older items first.

I used `Node.isConnected` instead of `contains` because I believe that `contains` doesn't work for shadow DOM unless you know the root element. It required bumping the minimum Firefox version to 53, though.

PR #463, opened 2020-04-19
Explicit entry points

toasted-nutbread:
This change makes it so that there is a more explicit entry point for all HTML pages. Since most of the script files are only declaring a class (e.g. [`ObjectPropertyAccessor`](https://github.com/FooSoft/yomichan/blob/a6344f635de2464465317b2ceeb6e19aef056848/ext/mixed/js/object-property-accessor.js)), utility object (e.g. [`jp`](https://github.com/FooSoft/yomichan/blob/a6344f635de2464465317b2ceeb6e19aef056848/ext/mixed/js/japanese.js)), or framework object (e.g. [`yomichan`](https://github.com/FooSoft/yomichan/blob/a6344f635de2464465317b2ceeb6e19aef056848/ext/mixed/js/core.js#L275)), I think it makes sense to separate out the actual instantiation logic and to make it follow a consistent pattern.

This change makes all entry points follow a consistent pattern, i.e. a `main()` function which is defined and invoked in a standalone file. `content-script-main` is a bit different because it's main function is called from a few different files. So in order to avoid a naming conflict, it's named `contentScriptMain` instead.

The primary motivation is to make components more reusable, testable, and standard. For example, if we want to test `Backend` in a special manner, it's probably best to have a distinction between importing the class and instantiating the class.


toasted-nutbread:
One point of consideration that is relevant to this: For the most part, the source code design uses composition rather than inheritance, with the exception of a few places. `EventDispatcher`, `TextScanner`, and `Display` are all subclassed. Composition in JS doesn't require any strict ordering of the dependencies, as long as they are all available at the time they need to be used. Inheritance, however, requires the base classes to be available at the time of subclass declaration.

The reason why this is related is due to the ordering of source files in the HTML `<script>` tags and manifest.json's `content_scripts.js` array. For the most part, the order is done alphabetically, with the exception of framework-like scripts coming first, and the `main` script coming last. What isn't exactly clear is that for the scripts with a class which `extends` something defined in another script, there is a specific ordering required. Not a big deal for `EventDispatcher`, since it's defined in core.js (first), but the other ones are a bit less obvious.

This isn't necessarily something that I think we need to go and focus a big change on, but it's just something to keep in mind. I was about to organize a few of the `<script>` imports in some of the HTML files until I remembered this.

siikamiika:
Some thoughts:

- `japanese.js` is split into two parts. The way it's structured now makes it work when it's placed so that `mixed` comes before `bg`. I think this isn't obvious without looking inside the files. Rather than keeping the files separate, would it be better if `jp` was a monolith that required calling something like `jp.setWanaKana(wanakana)` in the `main` function that used it? There could be a placeholder object for `wanakana` that throws for every function call. I think there was a discussion about this before, but I'm asking again with this new direction.
- In general, utility and framework object creation can be thought of as an entry point especially if they do things other than instantiating a class or creating completely new original things that only depend on DOM or `chrome` (but strictly always). Not sure if class definitions can have strange side effects.
- I searched the code for `const [A-Z]` and `^\s*window.*=` to find constants that are used across files. The only one I found quickly was `EXTENSION_IS_BROWSER_EDGE` and that seemed safe, but there could be others.
- To avoid the name conflict with `content-script-main.js`, would it be simpler to just run an anonymous function? I get the convention, but I'm afraid it adds some mental overhead to naming when you can't be sure if `main` is reserved. At least the file names already contain `main` so they're easy to find.

I went through all the JS files in the project and it seemed that everything was in place as long as the `main` functions work as expected.

siikamiika:
> One point of consideration that is relevant to this: For the most part, the source code design uses composition rather than inheritance, with the exception of a few places. `EventDispatcher`, `TextScanner`, and `Display` are all subclassed. Composition in JS doesn't require any strict ordering of the dependencies, as long as they are all available at the time they need to be used. Inheritance, however, requires the base classes to be available at the time of subclass declaration.

I agree that this is an issue with the way JS works without modules. At least now we should have some place to start from. We already had some discussion about making `TextScanner` use composition, but I went with inheritance because it was easier with the existing structure. Some more decoupling is needed to make composition viable. Not sure what should be done with `Display`. At least `DisplayFloat` is starting to get more disconnected from `DisplaySearch` in https://github.com/FooSoft/yomichan/pull/464 because they don't update their settings the same way anymore. `EventDispatcher` should be ok like you mentioned.

siikamiika:
> I went through all the JS files in the project and it seemed that everything was in place as long as the `main` functions work as expected.

The new `main` functions look like they do what they should, just not sure about some of the things I mentioned in the comment.

-------

It also seems that you can just redeclare the `main` function so overriding it shouldn't break things if it's run only once. However I don't think it's optimal to override potential earlier `main` as a convention either, it would just make writing the name less scary.

toasted-nutbread:
> `japanese.js` is split into two parts. The way it's structured now makes it work when it's placed so that `mixed` comes before `bg`. I think this isn't obvious without looking inside the files. Rather than keeping the files separate, would it be better if `jp` was a monolith that required calling something like `jp.setWanaKana(wanakana)` in the `main` function that used it? There could be a placeholder object for `wanakana` that throws for every function call. I think there was a discussion about this before, but I'm asking again with this new direction.

I think at some level we should get rid of the `Object.assign`, but I'm still conflicted as to what the best method would be. I've been hesitant to merge them because it would theoretically add initialization overhead + memory footprint increase for float.html, but in reality maybe it's not something worth being concerned with.


> In general, utility and framework object creation can be thought of as an entry point especially if they do things other than instantiating a class or creating completely new original things that only depend on DOM or `chrome` (but strictly always).

This is true. To kind of clarify my terms, I was thinking of "utility" as something without side effects, and "framework" as something with side effects (e.g. hooking an event listener). For both of them, my intent is that the side effects should, for the most part, be isolated to only affect things which use that object. So indeed, something [like this](https://github.com/FooSoft/yomichan/blob/master/ext/fg/js/document.js#L260-L294) would be considered a side effect, but the scope is intended to be limited. Similar for the `yomichan` object; it doesn't cause any side effects on the client page. (In this case I guess I mean side effects which affect the page visually or do something to user input.)

> Not sure if class definitions can have strange side effects.

I don't think so, at least not without doing something bizarre.

> I searched the code for `const [A-Z]` and `^\s*window.*=` to find constants that are used across files. The only one I found quickly was `EXTENSION_IS_BROWSER_EDGE` and that seemed safe, but there could be others.

`frontendInitializationData`. `window.wanakana` should probably be changed to just `wanakana` in a few locations so it can be `global`'d.

> To avoid the name conflict with `content-script-main.js`, would it be simpler to just run an anonymous function? I get the convention, but I'm afraid it adds some mental overhead to naming when you can't be sure if `main` is reserved. At least the file names already contain `main` so they're easy to find.

That would be fine by me. I didn't notice until the last file that there was a naming collision, which is why I didn't change it at that point.

> I agree that this is an issue with the way JS works without modules. At least now we should have some place to start from. We already had some discussion about making `TextScanner` use composition, but I went with inheritance because it was easier with the existing structure. Some more decoupling is needed to make composition viable. Not sure what should be done with `Display`. At least `DisplayFloat` is starting to get more disconnected from `DisplaySearch` in #464 because they don't update their settings the same way anymore. `EventDispatcher` should be ok like you mentioned.

Yeah it's not a huge deal, just something to be aware of.

> It also seems that you can just redeclare the `main` function so overriding it shouldn't break things if it's run only once. However I don't think it's optimal to override potential earlier `main` as a convention either, it would just make writing the name less scary.

Yeah, since `function` declarations are just on `window`, there isn't a `const` issue, so everything would technically still work. Just trying to avoid a case where we want to use it and then see something unexpected, like in a test or something.

siikamiika:
> I think at some level we should get rid of the `Object.assign`, but I'm still conflicted as to what the best method would be. I've been hesitant to merge them because it would theoretically add initialization overhead + memory footprint increase for float.html, but in reality maybe it's not something worth being concerned with.

I think it might be more difficult for browsers to optimize if they do stuff like storing multiple instances of the same object in the same place, but I don't know how they actually do that. I agree there's some overhead if you go by the raw size of the file times popups.

> This is true. To kind of clarify my terms, I was thinking of "utility" as something without side effects, and "framework" as something with side effects (e.g. hooking an event listener). For both of them, my intent is that the side effects should, for the most part, be isolated to only affect things which use that object. So indeed, something [like this](https://github.com/FooSoft/yomichan/blob/master/ext/fg/js/document.js#L260-L294) would be considered a side effect, but the scope is intended to be limited. Similar for the `yomichan` object; it doesn't cause any side effects on the client page. (In this case I guess I mean side effects which affect the page visually or do something to user input.)

Thanks for the clarification, agreed.

> `frontendInitializationData`. `window.wanakana` should probably be changed to just `wanakana` in a few locations so it can be `global`'d.

Could be worth it to change `frontendInitializationData` to something else at some point, but it's quite simple this way too. I approved your PR about `window.wanakana`. Do you think `wanakana.toKana` is fine to be used directly without `jp`? It basically does the IME conversion manually.

> That would be fine by me. I didn't notice until the last file that there was a naming collision, which is why I didn't change it at that point.

:+1: 

> Yeah, since `function` declarations are just on `window`, there isn't a `const` issue, so everything would technically still work. Just trying to avoid a case where we want to use it and then see something unexpected, like in a test or something.

I noticed that Firefox uses `let` internally for classes, but Chrome will override a class if defined multiple times.

toasted-nutbread:
> Do you think `wanakana.toKana` is fine to be used directly without `jp`?

I don't see any issue with it. The only consideration would be if it ever needs to be phased out due to deprecation or some bug or something. I only mention that because [it doesn't appear to be very active](https://github.com/WaniKani/WanaKana/commits/master) (2 years ago for most commits), but maybe that's because it's considered feature-complete.

> Could be worth it to change frontendInitializationData to something else at some point, but it's quite simple this way too.

I've been thinking about that too, I'll see if I come up with anything good.

> I noticed that Firefox uses `let` internally for classes, but Chrome will override a class if defined multiple times.

Interesting behaviour from the dev console for both browsers indeed:
Firefox:
```js
// New tab
class X{};class X{}; // enter
// SyntaxError: redeclaration of class X
```
```js
// New tab
class X{}; // enter
class X{}; // enter
// SyntaxError: redeclaration of let X
```

Chrome:
```js
// New tab
class X{};class X{}; // enter
// Uncaught SyntaxError: Identifier 'X' has already been declared
```
```js
// New tab
class X{}; // enter
class X{}; // enter
// no error
```

---

Another thing that came up in this PR is a very obvious `new` with side effects.
https://github.com/FooSoft/yomichan/blob/9ca906ef900dacfd10a548757d8a2b842a549088/ext/fg/js/float-main.js#L22-L24

I think it would be good to address this (later) at some level and add [`"no-new": "error"`](https://eslint.org/docs/rules/no-new) to .eslintrc.json, as well as addressing any other places where side effects are set up in constructors (i.e. move to `prepare`).

toasted-nutbread:
FYI, d106c638ed69b2c72895c1040b0e7bea2e31cdb7 moves popup-nested.js content into float-main.js. Both files are used only once and in that order, so it makes more sense to merge them. Also gives better parity with search-main.js, which also handles some script injection.

siikamiika:
> Another thing that came up in this PR is a very obvious `new` with side effects.

Its `prepare` is called from `Popup` with window messaging. I wonder how that should be handled.

Some other places:
- `backend-api-forwarder.js`
- `template-handler.js`
- `audio-system.js`
- `frontend-api-receiver.js`
- `popup.js`
- `audio-ui.js`
- `dictionaries.js`

toasted-nutbread:
Entry points are now IIFE in d8276a9d5d119edf1747593608d3e135947019f0. I didn't include any other indicator that it's the main function (comment or `(function main() { ... })()`) since the `*-main.js` file names should cover that.

Issue #452, opened 2020-04-12
Use Frontend's options for Popup and DisplayFloat


siikamiika:
The issue was first noted in https://github.com/FooSoft/yomichan/pull/439#issuecomment-612620944, but it turns out that the issue is not limited to the `Popup` class and `DisplayFloat` should also be updated when the `Frontend` using it changes.

https://github.com/FooSoft/yomichan/pull/417

siikamiika:
In order to make https://github.com/FooSoft/yomichan/issues/412 possible, there has to be a centralized place that determines which profile is used. I think this should be `Frontend`. This means that the pressed modifier is stored in `Frontend` until the popup closes.

In the case of https://github.com/FooSoft/yomichan/pull/389, `Display*` would send a message to its `Frontend` that actually changes the profile when the button or hotkey is pressed, or dropdown item is selected.

Search page is an exception since it has no `Frontend` (`search-frontend.js` creates a new popup).

siikamiika:
@toasted-nutbread Which one do you think is the best:

1. Broadcasting `optionsContext` from `Frontend` to everyone who shares the same `Popup.id`
2. Broadcasting `options` as-is without context, making `Frontend` the only place within popups that queries options from the backend
3. Broadcasting profile index or ID, similar to https://github.com/FooSoft/yomichan/pull/389

Performance wise this could go either way. If a tab has lots of frames there will be lots of noise when using `apiBroadcastTab`. However, it's faster and more consistent than letting everyone query options (takes away one hop).

There's an issue that API functions like `apiTermsFind` use `optionsContext` to decide different things. If they should be that way and it's not viable to move their invocation to `Frontend`, option 2 is a no-go.

`optionsContext` is sometimes ambiguous. If some profile is explicitly chosen (https://github.com/FooSoft/yomichan/pull/389), option 3 could be the way to go. It's not necessarily something to do in the scope of this issue. It's compatible with option 1 because `optionsContext` has `index`.

siikamiika:
It also seems possible to use window messaging to propagate options from `Frontend` to `DisplayFloat` through `Popup`, improving the performance.

toasted-nutbread:
Maybe the optimal choice might be a combination of 1 and 2. `Frontend` (or similar for search.html) could observe when changes to `options` and `optionsContext` occurs, and instruct the `Display` to update accordingly.

[...]

PR #464, opened 2020-04-19
Update Popup and DisplayFloat optionsContext from Frontend

siikamiika:
Resolves https://github.com/FooSoft/yomichan/issues/452.

Now setting `optionsContext` to `Popup` and `DisplayFloat` means also updating options. Because both use `optionsContext` separately from the `options` object, I simplified the code so that new options are fetched by the classes themselves. A bit different from the discussion in #452, but I thought it makes more sense this way.

Each time `optionsContext` is set to a `Popup`, it comes along with the unique ID of the `Frontend` that set it. This is needed when the `Frontend` using the `Popup` changes. For performance reasons, it's sent along with `Popup.showContent` arguments when scanning, but otherwise using `Popup.setOptionsContext`. Scanning checks if the `Frontend` ID has changed, but explicit calls don't because they're meant to be done only when the options can be expected to have changed.

### DIFF
--- a/ext/fg/js/frontend.js
+++ b/ext/fg/js/frontend.js
@@ -26,7 +26,7 @@
  */
 
 class Frontend extends TextScanner {
-    constructor(popup, url) {
+    constructor(popup, getUrl) {
         super(
             window,
             () => this.popup.isProxy() ? [] : [this.popup.getContainer()],
@@ -37,7 +37,8 @@ class Frontend extends TextScanner {
 
         this.popup = popup;
 
-        this._url = url;
+        this._url = null;
+        this._getUrl = getUrl;
 
         this._disabledOverride = false;
 
@@ -142,11 +143,12 @@ class Frontend extends TextScanner {
     async setPopup(popup) {
         this.onSearchClear(false);
         this.popup = popup;
-        await popup.setOptionsContext(this.getOptionsContext(), this._id);
+        await popup.setOptionsContext(await this.getOptionsContext(), this._id);
     }
 
     async updateOptions() {
-        this.options = await apiOptionsGet(this.getOptionsContext());
+        const optionsContext = await this.getOptionsContext();
+        this.options = await apiOptionsGet(optionsContext);
         this.setOptions(this.options, this._canEnable());
 
         const ignoreNodes = ['.scan-disable', '.scan-disable *'];
@@ -155,7 +157,7 @@ class Frontend extends TextScanner {
         }
         this.ignoreNodes = ignoreNodes.join(',');
 
-        await this.popup.setOptionsContext(this.getOptionsContext(), this._id);
+        await this.popup.setOptionsContext(optionsContext, this._id);
 
         this._updateContentScale();
 
@@ -214,7 +216,7 @@ class Frontend extends TextScanner {
         const searchText = textSource.text();
         if (searchText.length === 0) { return null; }
 
-        const {definitions, length} = await apiTermsFind(searchText, {}, this.getOptionsContext());
+        const {definitions, length} = await apiTermsFind(searchText, {}, await this.getOptionsContext());
         if (definitions.length === 0) { return null; }
 
         textSource.setEndOffset(length);
@@ -228,7 +230,7 @@ class Frontend extends TextScanner {
         const searchText = textSource.text();
         if (searchText.length === 0) { return null; }
 
-        const definitions = await apiKanjiFind(searchText, this.getOptionsContext());
+        const definitions = await apiKanjiFind(searchText, await this.getOptionsContext());
         if (definitions.length === 0) { return null; }
 
         return {definitions, type: 'kanji'};
@@ -240,7 +242,12 @@ class Frontend extends TextScanner {
         super.onSearchClear(changeFocus);
     }
 
-    getOptionsContext() {
+    async getOptionsContext() {
+        this._url = await this._getUrl();
+        return this.getOptionsContextCached();
+    }
+
+    getOptionsContextCached() {
         return {
             depth: this.popup.depth,
             url: this._url
@@ -248,7 +255,7 @@ class Frontend extends TextScanner {
     }
 
     _showPopupContent(textSource, type=null, details=null) {
-        const context = {optionsContext: this.getOptionsContext(), source: this._id};
+        const context = {optionsContext: this.getOptionsContextCached(), source: this._id};
         this._lastShowPromise = this.popup.showContent(
             textSource.getRect(),
             textSource.getWritingMode(),
--- a/ext/fg/js/frontend.js
+++ b/ext/fg/js/frontend.js
@@ -33,6 +33,8 @@ class Frontend extends TextScanner {
             [(x, y) => this.popup.containsPoint(x, y)]
         );
 
+        this._id = yomichan.generateId(16);
+
         this.popup = popup;
 
         this._url = url;
@@ -141,7 +143,7 @@ class Frontend extends TextScanner {
         this.onSearchClear(false);
         this.popup = popup;
         await popup.setOptions(this.options);
-        await popup.setOptionsContext(this.getOptionsContext());
+        await popup.setOptionsContext(this.getOptionsContext(), this._id);
     }
 
     async updateOptions() {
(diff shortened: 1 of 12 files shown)


siikamiika:
I did https://github.com/FooSoft/yomichan/pull/464/commits/92d4449f2149a7e39c5cf8c602353fbf0bc1629f, but I think the preview frame has been broken by some earlier update. Its theme doesn't update when changing options, but it does flash briefly.

ext/fg/js/popup.js
```diff
@@ -72,19 +75,20 @@ class Popup {
...
-        this._options = options;
+    async setOptionsContext(optionsContext, source) {
+        this._optionsContext = optionsContext;
+        this._previousOptionsContextSource = source;
+
+        this._options = await apiOptionsGet(optionsContext);
         this.updateTheme();
+
+        this._invokeApi('setOptionsContext', {optionsContext});
     }
```
siikamiika, line 92: I hope this is correct usage of "source" :joy:  Could also be `frontendId` or something similar

toasted-nutbread:
> I did [92d4449](https://github.com/FooSoft/yomichan/commit/92d4449f2149a7e39c5cf8c602353fbf0bc1629f), but I think the preview frame has been broken by some earlier update. Its theme doesn't update when changing options, but it does flash briefly.

Works for me on latest master in Chrome dev. Chrome has a text color bug, but changing the theme works for me.

ext/fg/js/frontend.js
```diff
@@ -26,24 +26,23 @@
  */
 
 class Frontend extends TextScanner {
-    constructor(popup) {
+    constructor(popup, url) {
```
toasted-nutbread, line 29: This, or some other new class, will need to handle URL changes for the root frame.

https://developer.mozilla.org/en-US/docs/Web/API/Window/popstate_event

siikamiika:
> > I did [92d4449](https://github.com/FooSoft/yomichan/commit/92d4449f2149a7e39c5cf8c602353fbf0bc1629f), but I think the preview frame has been broken by some earlier update. Its theme doesn't update when changing options, but it does flash briefly.
> 
> Works for me on latest master in Chrome dev. Chrome has a text color bug, but changing the theme works for me.

The issue is actually that the preview frame chooses the wrong profile. It works when I edit the profile that matches the settings page optionsContext. Will have to override `Frontend.getOptionsContext` with something that returns the current edited profile index.

ext/fg/js/frontend.js
siikamiika: It seems that it's not possible to detect a `history.pushState` call without monkey patching the object. It could work similar to `FrameOffsetForwarder` asking the parent `Frontend` for the root frame or iframe URL since the context URL is for the current frame.

siikamiika: Actually that's not possible here, but the popup proxy framework could be used instead.

siikamiika:
I'll review https://github.com/FooSoft/yomichan/pull/463 before doing more changes to this PR since there are some potential conflicts and this one seems to have some issues.

ext/fg/js/frontend.js
toasted-nutbread: Oh right, I forgot about that. I'm not sure if you can even monkey patch it due to the script context isolation. Not sure what the best workaround would be in that case.

siikamiika:
Noticed that `content-script-main.js` is becoming quite large and messy, should probably refactor it later

ext/fg/js/frontend.js
siikamiika: https://github.com/FooSoft/yomichan/pull/464/commits/b050333078b12f2ae14f49a46c04085b8e11d0b8 and https://github.com/FooSoft/yomichan/pull/464/commits/2b2445ec31f0a1d05f1cabf30caeefb558195aa7

siikamiika:
https://github.com/FooSoft/yomichan/pull/464/commits/52ae06ff92ea36ed5ace6a4c08ef10eebe1e0a2b fixes the issues in the preview frame. It changes the prepare style similar to that of `Popup` and `DisplayFloat`: https://github.com/FooSoft/yomichan/pull/463#issuecomment-617458268

toasted-nutbread:
> Noticed that `content-script-main.js` is becoming quite large and messy, should probably refactor it later

Agreed; probably best to do after the changes to `frontendInitializationData` are made (#481).

ext/fg/js/popup.js
```diff
@@ -120,8 +124,14 @@ class Popup {
         return false;
     }
 
-    async showContent(elementRect, writingMode, type=null, details=null) {
+    async showContent(elementRect, writingMode, type, details, context) {
```
toasted-nutbread, line 127: Maybe not something worth addressing here, but there are a bunch of terms used for `{...}` arguments in functions throughout the project: `details`, `context`, `params`, and `options`. We may want to address this at some point in the future since the terms are so general and can feel interchangeable.

The reason why this one stands out is that there are two. I understand why a new one had to be added, because one is sent to the frame and one is used directly. However, looking at it from the perspective of someone using the `Popup` class's public API, it would be more straightforward to use a single `details` object. The fact that some of the options are passed to the frame is an implementation detail. However x2, this would require `Popup` to know which `details` are relevant for the frame, and which aren't.

It's potentially a design flaw with the APIs and how they use eachother, so I would consider it out of scope for this change.

ext/fg/js/frontend.js
```diff
@@ -230,7 +230,7 @@ class Frontend extends TextScanner {
         const searchText = textSource.text();
         if (searchText.length === 0) { return null; }
 
-        const definitions = await apiKanjiFind(searchText, this.getOptionsContext());
+        const definitions = await apiKanjiFind(searchText, await this.getOptionsContext());
```
toasted-nutbread, line 233: There is now a redunant `await this.getOptionsContext()` when the following happens:

```js
results = (
    await this.findTerms(textSource) ||
    await this.findKanji(textSource)
);
```

Could instead `await this.getOptionsContext()` once inside of `onSearchSource` and pass it to both functions. It could also be passed to the `showContent` and `_showPopupContent` calls, partially eliminating the need for `getOptionsContextCached`. `_updatePopupPosition` could them be updated to `await this.getOptionsContext()` also, completely eliminating the need for `getOptionsContextCached`.

I don't think `Frontend.showContent` is used externally, so I don't think there would be a concern with adding an `optionsContext` parameter to it. `showContent` could then also use the `url` from the context instead of `this._url`, which may remove the need for that field (I think it's only used for the cached version?).

ext/fg/js/frontend.js
```diff
@@ -26,24 +26,24 @@
  */
 
 class Frontend extends TextScanner {
-    constructor(popup) {
+    constructor(popup, getUrl) {
```
toasted-nutbread, line 29: `getUrl` could be made to accept `null` also, and just return `window.location.href` in that case. That would remove an `await` call for the typical use case (non-nested popup on a content page).

siikamiika:
https://github.com/FooSoft/yomichan/pull/464/commits/649bff9a2df397cd688da3bf3e29f316d2dfa1ba addresses some of the issues that came up in the review. I'm not sure how much it helps that the `window.location.href` fallback is in `Frontend` instead of `content-script-main` since in both cases there wasn't any actual async code, but at least the preview frame override makes more sense now.

There's one issue though. `Frontend.showContent` needs `url` for `Display.setContentTerms` and `setContentKanji`. This is now making `Frontend.showContent` dependent on `optionsContext` format. It actually already happens in the preview frame that you can't get the URL anywhere, but I doubt anyone will try to add cards from there and wonder why the settings page URL isn't there. Maybe the override should be changed to merge `{url, depth}` with `{index}`.
`backend.js`:
```javascript
    getProfile(optionsContext) {
        const profiles = this.options.profiles;
        if (typeof optionsContext.index === 'number') {
            return profiles[optionsContext.index];
        }
        const profile = this.getProfileFromContext(optionsContext);
        return profile !== null ? profile : this.options.profiles[this.options.profileCurrent];
    }
```

ext/fg/js/popup.js
siikamiika: I agree that this is is somewhat confusing, especially when one of them contains another but there's still another argument with the same name. I hope this PR will help to simplify it a bit though, since options aren't passed as an object from `Frontend` to `Popup` now and every class gets their own options using the context.

siikamiika:
> Maybe the override should be changed to merge `{url, depth}` with `{index}`.

I'll leave it as it is for now since this shouldn't matter much.

PR #466, opened 2020-04-21
Use wanakana object directly

toasted-nutbread:
As mentioned in https://github.com/FooSoft/yomichan/pull/463#issuecomment-616831927


siikamiika:
Thanks!

PR #467, opened 2020-04-22
Create dynamic script/CSS loader utility

toasted-nutbread:
First part of addressing https://github.com/FooSoft/yomichan/pull/463#issuecomment-617450215. The original issue was that, despite being `async=false`, the scripts are not loaded immediately after injection into the DOM. So there wasn't a way to know when all the scripts were completed loading. Injecting an additional script which triggers an event solves this problem.

This should open the path to not using `frontendInitializationData`, but instead just calling a function that is now available in the global scope.

### DIFF
--- a/ext/mixed/js/dynamic-loader-sentinel.js
+++ b/ext/mixed/js/dynamic-loader-sentinel.js
@@ -1,0 +1,18 @@
+/*
+ * Copyright (C) 2020  Yomichan Authors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


siikamiika:
Looks cool, and nice use of `document.currentScript` :+1:  Should there be a cleanup for the injected sentinel script?

siikamiika:
That must be the fastest reaction to a review ever :joy:

toasted-nutbread:
What the hell, I didn't even read this before I pushed that change. I remembered yesterday that I forgot to do this, but didn't change it until just now. What are the odds.

siikamiika:
We've been working on this for a while so I guess sometimes the improbable happens too. That, or I'm reading your mind

PR #472, opened 2020-04-23
Fix high pitch calculation

### LINKED ISSUE #469: Inconsistent downstep notation for pitch accent on first mora
First of all, thanks so much for the pitch accent update. This update is truly amazing.

I just started trying it out for some words, and I'm seeing some minor inconsistencies with the line over the first mora for the downstep notation.

---
The line is not shown over the first mora for heiban words.
image

I assume this is intentional to show that the word starts low. However, words with the nakadaka and odaka pitch accents, where it starts low, goes high, and then drops again, still include the line over the first mora. If the intent is to only show high pitch mora and the downstep, probably only atamadaka words should have the line over the first mora. Here's a direct comparison.

image

toasted-nutbread:
Fixes #469.

I created [this](https://toasted-nutbread.github.io/japanese-pitch-accent-table/) for testing, which contains data which comes up in Google searches for "[Japanese pitch accent table](https://www.google.com/search?q=Japanese+pitch+accent+table&tbm=isch)". https://imgur.com/oEjxhWU

### DIFF
--- a/ext/mixed/js/japanese.js
+++ b/ext/mixed/js/japanese.js
@@ -116,7 +116,11 @@ const jp = (() => {
     // Mora functions
 
     function isMoraPitchHigh(moraIndex, pitchAccentPosition) {
-        return pitchAccentPosition === 0 ? (moraIndex > 0) : (moraIndex < pitchAccentPosition);
+        switch (pitchAccentPosition) {
+            case 0: return (moraIndex > 0);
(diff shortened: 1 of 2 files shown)


siikamiika:
Thanks for the fix. That table is really neat! I felt that there was something off about the actual pronunciation and what was shown in the pitch accent graph, but I thought it was just regional difference. The examples in that table look more consistent with the pronunciation now.

PR #476, opened 2020-04-23
Add duplicateScope: 'deck' option

toasted-nutbread:
See: https://github.com/FooSoft/anki-connect/pull/157

### DIFF
--- a/ext/bg/js/anki-note-builder.js
+++ b/ext/bg/js/anki-note-builder.js
@@ -32,7 +32,10 @@ class AnkiNoteBuilder {
             fields: {},
             tags,
             deckName: modeOptions.deck,
-            modelName: modeOptions.model
+            modelName: modeOptions.model,
+            options: {
+                duplicateScope: 'deck'
+            }
         };
 
         for (const [fieldName, fieldValue] of modeOptionsFieldEntries) {
--- a/ext/bg/js/anki-note-builder.js
+++ b/ext/bg/js/anki-note-builder.js
@@ -34,7 +34,7 @@ class AnkiNoteBuilder {
             deckName: modeOptions.deck,
             modelName: modeOptions.model,
             options: {
-                duplicateScope: 'deck'
+                duplicateScope: options.anki.duplicateScope
             }
         };
 
--- a/ext/bg/data/options-schema.json
+++ b/ext/bg/data/options-schema.json
@@ -492,6 +492,7 @@
                                     "screenshot",
                                     "terms",
                                     "kanji",
+                                    "duplicateScope",
                                     "fieldTemplates"
                                 ],
                                 "properties": {
@@ -586,6 +587,11 @@
                                                 }
                                             }
                                         }
+                                    },
+                                    "duplicateScope": {
+                                        "type": "string",
+                                        "default": "collection",
+                                        "enum": ["collection", "deck"]
                                     },
                                     "fieldTemplates": {
                                         "type": ["string", "null"],
--- a/ext/bg/js/anki.js
+++ b/ext/bg/js/anki.js
@@ -91,7 +91,7 @@ class AnkiConnect {
         if (!this._enabled) { return []; }
         await this._checkVersion();
         const actions = notes.map((note) => {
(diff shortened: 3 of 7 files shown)


siikamiika:
Is this intended to be configurable from the settings, or will that require some changes to AnkiConnect first?

toasted-nutbread:
I decided I would add a setting to control that, so I'll update this PR with that soon.

toasted-nutbread:
Setting added, also updated `findNoteIds` to obey the setting.

ext/bg/js/anki-note-builder.js
```diff
@@ -34,7 +34,7 @@ class AnkiNoteBuilder {
             deckName: modeOptions.deck,
             modelName: modeOptions.model,
             options: {
-                duplicateScope: 'deck'
+                duplicateScope: options.anki.duplicateScope
```
siikamiika, line 37: Just checking, is the default changed to `collection` on purpose?

siikamiika:
Another issue, the button for viewing the added note is not available when scanning the word for the second time, but only right after adding it. It's been like this as long as I remember, and I'm just wondering if this is something that can be fixed with the current AnkiConnect API.

ext/bg/js/anki-note-builder.js
toasted-nutbread: This is intentional; I decided to make 'deck' an opt-in behaviour since it is different from the previous default.

toasted-nutbread:
> Another issue, the button for viewing the added note is not available when scanning the word for the second time, but only right after adding it. It's been like this as long as I remember, and I'm just wondering if this is something that can be fixed with the current AnkiConnect API.

Assuming you mean showing the book icon for terms which are already added, this should already work. I think there was an issue that asked for this too, but I can't find it.
c5d6b9452d1bfa51a5fef9a19082d115ef180a9b...7ce54864f3fbaaeade2ebc20d80bcb084d9fa426

siikamiika:
> Assuming you mean showing the book icon for terms which are already added, this should already work. I think there was an issue that asked for this too, but I can't find it.
> [c5d6b94...7ce5486](https://github.com/FooSoft/yomichan/compare/c5d6b9452d1bfa51a5fef9a19082d115ef180a9b...7ce54864f3fbaaeade2ebc20d80bcb084d9fa426)

This doesn't seem to work for me for some reason. It doesn't seem to be an issue with this branch.
simplescreenrecorder-2020-04-27_00 05 16

--------

(Unrelated, I think there's some issue with the pitch accent disambiguation because it shows the katakana for ウイ although it's already available)

toasted-nutbread:
Works for me. You can try checking the search query generated and inputting it directly into Anki's browser, see if anything comes up. I think `"` characters need to be stripped from deck/expression names due to there being no apparent way to escape them, not sure if that's the issue or not.

test

siikamiika:
> Works for me. You can try checking the search [query](https://github.com/FooSoft/yomichan/pull/476/commits/1f946d798d388c056de785b1744dcc952fc5345e#diff-ecd1d3e8d221b35d06f9b0e50f9e833aR95) generated and inputting it directly into Anki's browser, see if anything comes up. I think `"` characters need to be stripped from deck/expression names due to there being no apparent way to escape them, not sure if that's the issue or not.

With this branch, the value of `query` will be `front:"<span class=expression-normal>ウィキペディア</span>"`. AnkiConnect returns `[ [], [] ]`.

toasted-nutbread:
I'm guessing it's stripping spaces from `<span class="expression-normal">ウィキペディア</span>`, changing it to `<span class=expression-normal>ウィキペディア</span>`? Also, it looks like spaces are supported in field names, so I've updated the location of the quotes to support that.

siikamiika:
This seems to be an issue with `resultOutputMode: "merge"` only. The previous commit didn't fix the issue for me, not sure how it would affect other setups.

toasted-nutbread:
What's odd is that it seems possible to quote strings using a single quote also. But as I looked into the Anki source code, I couldn't find anywhere that shows that case being handled. I also saw some code which looked like it should handle escape sequences, but they didn't work from the browser. So I'll refrain from adding that as a backup since it's unclear how that is working, but I have tested the recent change that I just made and it shouldn't break anything.

Issue #455, opened 2020-04-12
WebExtension manifest v3


toasted-nutbread:
The Google Chrome team is working on a version 3 of the WebExtension manifest. While the timelines are still indeterminate at this point, it's better to start tracking and thinking about the changes this will entail sooner rather than later. We should use this issue to document any potential problems we may encounter.

Relevant links:
* [[Google] Migrating to Manifest V3](https://developer.chrome.com/extensions/migrating_to_manifest_v3)
* [[Mozilla] Mozilla’s Manifest v3 FAQ](https://blog.mozilla.org/addons/2019/09/03/mozillas-manifest-v3-faq/)

Noteworthy points:

> Are you currently using chrome.tabs.executeScript({code: '...'}), eval(), or new Function() in background contexts or content scripts?
> * Move all external code (JS, Wasm, CSS) into your extension bundle.
> * Update script and style references to load resources from the extension bundle.
> * Use chrome.runtime.getURL() to build resource URLs at runtime.

This will present an issue with Handlebars.js, since it uses `Function` constructor to dynamically eval code.

> Are you currently using chrome.runtime.getBackgroundPage(), chrome.extension.getBackgroundPage(), chrome.extension.getExtensionTabs(), or chrome.extension.getViews()?
> * Migrate to passing messages between other contexts and the background service worker.

This is planned and in progress. Already started here: #447.

> Are you currently using chrome.runtime.getBackgroundPage(), chrome.extension.getBackgroundPage(), chrome.extension.getExtensionTabs(), or chrome.extension.getViews()?
> * Migrate to passing messages between other contexts and the background service worker.

This is potentially the biggest change, and will affect at least a few things that I am aware of.
* Service workers have no DOM, so the `clipboardPasteTarget` shenanigans won't work. Not sure what the workaround for that would be.
* Service workers lifetime is a bit odd, as they can be terminated to save memory. It seems there are ways to [keep a service worker alive](https://developer.mozilla.org/en-US/docs/Web/API/ExtendableEvent/waitUntil), but the browser may still kill it for other reasons, "such as infinite loops and tasks exceeding imposed time limits (if any) while handling the events." Not sure exactly how this will function in practice, or if extensions are granted more leniency.
* IndexedDB access is still present.

Related:
* [[Google] Migrating from Background Pages to Service Workers](https://developer.chrome.com/extensions/migrating_to_service_workers)
[...]

PR #477, opened 2020-04-25
Anki audio download

### LINKED ISSUE #419: anki cloze note /{audio} problem
I'd like fill my fied like `{{c1::{expression}}} {{c2::{audio}}}`, but the result is `{{c2::教師}}{{c1::}}[sound:yomichan_きょうし_教師.mp3]`. 

It semms the {audio} will always be appended at end. Is there any way let it work as I need?

toasted-nutbread:
Download audio binary instead of downloading an `Audio` object and testing its `duration` parameter for validity. This eliminates a redundant download in Anki for a file that has already been downloaded by the web browser.

Resolves #419, and would be necessary for #455. Originally brought up in https://github.com/FooSoft/yomichan/pull/277#issuecomment-552145454.

### DIFF
--- a/ext/bg/js/anki-note-builder.js
+++ b/ext/bg/js/anki-note-builder.js
@@ -39,25 +39,6 @@ class AnkiNoteBuilder {
             note.fields[fieldName] = await this.formatField(fieldValue, definition, mode, context, options, templates, null);
         }
 
-        if (!isKanji && definition.audio) {
-            const audioFields = [];
-
-            for (const [fieldName, fieldValue] of modeOptionsFieldEntries) {
-                if (fieldValue.includes('{audio}')) {
-                    audioFields.push(fieldName);
-                }
-            }
-
-            if (audioFields.length > 0) {
-                note.audio = {
-                    url: definition.audio.url,
-                    filename: definition.audio.filename,
-                    skipHash: '7e2c2f954ef6051373ba916f000168dc', // hash of audio data that should be skipped
-                    fields: audioFields
-                };
-            }
-        }
-
         return note;
     }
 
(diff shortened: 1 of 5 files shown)


toasted-nutbread:
@FooSoft I'm looking through some of the audio code, and when loading an `Audio` object, there are two durations which cause the audio to be ignored: `5.694694` and `5.720718`. The first one corresponds to [invalid audio from the jpod101 source](https://assets.languagepod101.com/dictionary/japanese/audiomp3.php?kana=invalid) and the MD5 hash `7e2c2f954ef6051373ba916f000168dc`, but I'm not sure what the second one corresponds to. Do you know what the second one is/was for?

Originally added here: https://github.com/FooSoft/yomichan/commit/c8eb77cfd9e30625ae33739b7de5a86f59d0457c#diff-d5a318d27e03a2960299ccc82727e376R186

FooSoft:
Ah yes, I remember this. Firefox and Chrome were reporting different audio lengths for the same file for some strange reason. Both exist to catch the invalid audio from jpod101.

toasted-nutbread:
Thanks, I've updated the comments to clarify this behaviour.

ext/bg/js/anki-note-builder.js
```diff
@@ -126,6 +120,7 @@ class AnkiNoteBuilder {
         if (reading) { filename += `_${reading}`; }
         if (expression) { filename += `_${expression}`; }
         filename += '.mp3';
+        filename = filename.replace(/\]/g, '');
```
siikamiika, line 123: Since you added it, are there other characters that have to be escaped or sanitized from the filename? If it's actually stored in the filesystem under that name, then `/` and `\` can't be used either, but there are also other characters that are disallowed in filenames.

toasted-nutbread: The reason why "]" is removed is due to how Anki handles them, not necessarily file name restrictions. That is probably a valid thing to do however.

https://github.com/ankitects/anki/blob/913c26870529174b50f4207397ede7444a48dc6e/pylib/anki/utils.py#L94
https://github.com/ankitects/anki/blob/913c26870529174b50f4207397ede7444a48dc6e/pylib/anki/exporting.py#L67
https://github.com/ankitects/anki/blob/913c26870529174b50f4207397ede7444a48dc6e/pylib/anki/media.py#L35

siikamiika: Yeah, makes sense that it would remove the `]` character since it's used to terminate the `[audio:...]` block.
(offtopic, Anki seems to ignore empty `[audio:]` audio blocks in those functions)

toasted-nutbread: > are there other characters that have to be escaped or sanitized from the filename?

(Just FYI, I consider this out of scope for this change, but will gladly make the update in a follow-up PR.)

siikamiika: Thanks for clarifying, I was waiting for a new commit addressing this issue

siikamiika:
I swear this worked when I tested, but now I can't add anything with `{audio}`. Does it work for you?

toasted-nutbread:
#498?

PR #483, opened 2020-04-26
Api invoke with progress

toasted-nutbread:
In order to convert [some of the remaining `util*` functions](https://github.com/FooSoft/yomichan/blob/a49e4ccc4e5c7defb30bd88535c18137acf9812c/ext/bg/js/util.js#L69-L83) that use `utilBackend` into `api*` functions, it is necessary to have a mechanism for `api` calls with progress callbacks. This change implements such a mechanism.

The mechanism works using two steps:
1. The frontend sends a request to the backend to open a persistent communication [Port](https://developer.chrome.com/extensions/runtime#type-Port). The backend opens a port and initializes it with a handler map.
2. After the frontend receives this port, it sends an `{action, params}` message to that port. The port `ack`'s the message and then runs the handler. Progress messages are delivered to the port, as well as a completion or error message based on the execution of the handler.

This change does not implement any handlers, as I tried to keep the scope fairly limited.

### DIFF
--- a/ext/bg/js/backend.js
+++ b/ext/bg/js/backend.js
@@ -115,7 +115,8 @@ class Backend {
             ['purgeDatabase', {handler: this._onApiPurgeDatabase.bind(this), async: true}],
             ['getMedia', {handler: this._onApiGetMedia.bind(this), async: true}],
             ['log', {handler: this._onApiLog.bind(this), async: false}],
-            ['logIndicatorClear', {handler: this._onApiLogIndicatorClear.bind(this), async: false}]
+            ['logIndicatorClear', {handler: this._onApiLogIndicatorClear.bind(this), async: false}],
+            ['createActionPort', {handler: this._onApiCreateActionPort.bind(this), async: false}]
         ]);
 
         this._commandHandlers = new Map([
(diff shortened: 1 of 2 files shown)


toasted-nutbread:
For my test case, I used the following:

[backend.js, line 118](https://github.com/FooSoft/yomichan/blob/80adb0f4796f9b1bc9e3d563d969800648fa55a2/ext/bg/js/backend.js#L118):
```patch
-this._messageHandlersWithProgress = new Map();
+this._messageHandlersWithProgress = new Map([
+    ['test', {
+        handler: async (params, sender, onProgress) => {
+            console.log('sender', sender);
+            for (let i = 0; i < 20; ++i) {
+                onProgress({i, params});
+                await promiseTimeout(100);
+            }
+            return 44;
+        },
+        async: true
+    }]
+]);
```

[Add to api.js](https://github.com/FooSoft/yomichan/blob/80adb0f4796f9b1bc9e3d563d969800648fa55a2/ext/mixed/js/api.js#L291):
```patch
+async function apiTest() {
+    const v = await _apiInvokeWithProgress('test', {value: 55}, (data) => { console.log('progress', data); });
+    console.log('result=' + v);
+}
```

Then in a dev console on a Yomichan extension page (not background.html), just run `apiTest()`. It should print 20 progress messages followed by a result message.

ext/bg/js/backend.js
```diff
@@ -786,6 +787,26 @@ class Backend {
         this._updateBadge();
     }
 
+    _onApiCreateActionPort(params, sender) {
+        let tabId;
+        if (!(
+            sender &&
+            sender.tab &&
+            (typeof (tabId = sender.tab.id)) === 'number'
```
siikamiika, line 796: I wonder if we should use https://eslint.org/docs/rules/no-cond-assign. We are using strict comparisons everywhere so it's hard to mistype `=` when you actually want `===`, but it still feels like a side effect of the conditional statement.

siikamiika: Also for these types of cases there's a solution in the newest browser versions: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

toasted-nutbread: > Browser compatibility: Chrome 80, Firefox 74
😭

I frequently use the assign-and-test pattern elsewhere with the rationale being that it sometimes reduces lines-of-code, but I can change it here. It can be confusing, so perhaps the brevity isn't worth it.

toasted-nutbread: Style updated in b879ff83a491f0db74352eb3b789a5abe808daf2

PR #485, opened 2020-04-27
Object property accessor API update

toasted-nutbread:
Added `delete` and `swap`; simplified names; added documentation; updated tests.


toasted-nutbread:
For #335

PR #486, opened 2020-04-28
Remove Frontend inheritance

toasted-nutbread:
Follow-up of #484. `QueryParser` still inherits, which can be addressed in the future.

* Simplified `setEnabled` and `setOptions` APIs. `setOptions` now only takes one argument (`options`), and `setEnabled` now only takes one argument (`enabled`).
* Fixed a bug where `QueryParser` would not work when `options.general.enable` was false.
* `onWindowMessage` handler is now added, removed, and tracked by `Frontend`. The change takes place whenever `setEnabled` is changed.
* A few refactorings, including using `Array.push` instead of `concat`, making `Frontend.options` private, and using `.bind` instead of an arrow function.

### DIFF
--- a/ext/bg/js/search-query-parser.js
+++ b/ext/bg/js/search-query-parser.js
@@ -86,6 +86,7 @@ class QueryParser extends TextScanner {
 
     setOptions(options) {
         super.setOptions(options);
+        super.setEnabled(true);
         this.queryParser.dataset.termSpacing = `${options.parsing.termSpacing}`;
     }
 
--- a/ext/bg/js/settings/popup-preview-frame.js
+++ b/ext/bg/js/settings/popup-preview-frame.js
@@ -66,11 +66,10 @@ class SettingsPopupPreview {
         this.popup.setCustomOuterCss = this.popupSetCustomOuterCss.bind(this);
 
         this.frontend = new Frontend(this.popup);
-
         this.frontend.getOptionsContext = async () => this.optionsContext;
-        this.frontend.setDisabledOverride(true);
-
         await this.frontend.prepare();
+        this.frontend.setDisabledOverride(true);
+        this.frontend.canClearSelection = false;
 
         // Update search
         this.updateSearch();
--- a/ext/bg/js/settings/popup-preview-frame.js
+++ b/ext/bg/js/settings/popup-preview-frame.js
@@ -68,8 +68,7 @@ class SettingsPopupPreview {
         this.frontend = new Frontend(this.popup);
 
         this.frontend.getOptionsContext = async () => this.optionsContext;
-        this.frontend.setEnabled = () => {};
-        this.frontend.clearSelection = () => {};
+        this.frontend.setDisabledOverride(true);
 
         await this.frontend.prepare();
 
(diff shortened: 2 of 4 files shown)


siikamiika:
I'll review this before continuing on https://github.com/FooSoft/yomichan/pull/487

ext/bg/js/search-query-parser.js
```diff
@@ -86,6 +86,7 @@ class QueryParser extends TextScanner {
 
     setOptions(options) {
         super.setOptions(options);
+        super.setEnabled(true);
```
siikamiika, line 89: I actually allowed query parser to be disabled on purpose when scanning is disabled, but it would probably make more sense if the option only applied to website content because the search page still works normally. The option description says "Enable content scanning" which is a bit ambiguous.

ext/fg/js/frontend.js
```diff
@@ -272,6 +272,16 @@ class Frontend {
         return this._lastShowPromise;
     }
 
+    _updateTextScannerEnabled() {
+        const enabled = (
+            this.options !== null &&
```
siikamiika: I think this null check should probably be removed and `setDisabledOverride` called after prepare in preview frame.

ext/fg/js/frontend.js
```diff
@@ -41,15 +35,23 @@ class Frontend extends TextScanner {
...
         this._orphaned = false;
         this._lastShowPromise = Promise.resolve();
 
+        this._enabledEventListeners = new EventListenerCollection();
+        this._textScanner = new TextScanner(
+            window,
+            () => this.popup.isProxy() ? [] : [this.popup.getContainer()],
+            [(x, y) => this.popup.containsPoint(x, y)]
+        );
+        this._textScanner.onSearchSource = this.onSearchSource.bind(this);
```
siikamiika, line 51: It looks a bit inconsistent that the synchronous listener(s) use `this._textScanner.on(...)` but this asynchronous one is overridden. This is probably the simplest way to do it, but just mentioning

toasted-nutbread: I agree; I was trying to make the switch with minimal changes, and I think modifying `onSearchSource` would have been a bit more involved. Mainly because it affects `QueryParser` too. Eventually that class can be updated as well.

ext/bg/js/search-query-parser.js
toasted-nutbread: To me, it feels like there should be a distinction between the two. It's odd that the search page will still work and display the full text from the input, but you won't be able to scan it. Since the page is dedicated to do this task, IMO it would be fine to leave it always on. I can revert if you like, but that was my line of thinking.

siikamiika: I agree with you, I think it's ok to leave the exception there. It didn't feel like a big deal when I did it and I even thought it could be desirable sometimes to also disable scanning here, but it could also be very confusing without having any indication that it's disabled. This is a much better solution than implementing some kind of indicator for the unverified use case.

ext/fg/js/frontend.js
toasted-nutbread: bd717c087652dce4955f587d2e545419bf992a16

ext/bg/js/settings/popup-preview-frame.js
```diff
@@ -68,8 +68,7 @@ class SettingsPopupPreview {
         this.frontend = new Frontend(this.popup);
 
         this.frontend.getOptionsContext = async () => this.optionsContext;
-        this.frontend.setEnabled = () => {};
-        this.frontend.clearSelection = () => {};
```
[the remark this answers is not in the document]
toasted-nutbread: bd717c087652dce4955f587d2e545419bf992a16

Issue #412, opened 2020-03-13
Modifier key profile condition


siikamiika:
one other possible solution that came to my mind is to have different modifier keys trigger different profiles.(a new type of profile usage condition)
I don't know if this is easier to implement than the original request though.

_Originally posted by @Pintliz in https://github.com/FooSoft/yomichan/issues/320#issuecomment-574220642_

siikamiika:
Depends on https://github.com/FooSoft/yomichan/issues/452

siikamiika:
@toasted-nutbread I have a POC here: https://github.com/siikamiika/yomichan/commits/scanning-key-combinations-2

Any suggestions how to change the input type of the condition when changing between modifier selection and others? I created these templates first
```html
<template id="condition-input-text-template"><input type="text" class="form-control" /></template>
<template id="condition-input-select-template"><select class="form-control"></select></template>
```

but it would have been a bit complicated with the listeners. Another option would be to make the input listen to key events and choose the modifier based on the pressed key.

toasted-nutbread:
You could use `EventListenerCollection` so you don't have to manually track all the events which need to be removed. I guess it doesn't support jQuery's `on`/`off` style currently, but you can address that by using the native element or making `EventListenerCollection` support `on`/`off`. (Aside: moving away from jQuery/Bootstrap on the settings page in the v2 updates.)

If you're intending on supporting modifier combinations (e.g. `alt + shift`), then you probably want to use the text input + detect key strokes. `<select>` can do multiple, but it's layout is different. Make sure to use esc/backspace to clear. I think this will just have to be handled as a special case in the UI for now.

siikamiika:
Ok, thanks for the tips. I'll try what works the best. `<select>` support could be useful for other conditions in the future so I'll look at the `EventListenerCollection` approach first.

It's possible to support combinations with simple conditions too using `Modifier Key = alt AND Modifier Key = shift`. It's more complicated to implement stuff like `Modifier Key != alt + shift` on single row but might be possible with transforms.

siikamiika:
I had to do some refactoring anyway for `select` support so I was able to contain the event listener activation within a function that creates the `<input>` or `<select>` element. This doesn't use `EventListenerCollection`: https://github.com/siikamiika/yomichan/commit/6024e4981aea7df083ace4633c2ba85c32572a05

[...]

PR #487, opened 2020-04-29
Modifier key profile condition

siikamiika:
Fixes https://github.com/FooSoft/yomichan/issues/412

Adds a new profile condition that lets you specify a modifier key or an exact set of modifier keys that must or must not be pressed in order to activate a profile when scanning.

The set of active modifiers is changed when you move the mouse cursor and press or release Shift, Alt, Ctrl or Meta. If a popup is open when you do this, a boolean is stored that will cause the profile to be updated on the next popup close or scan.

I'm not completely sure about adding support for `Meta` because its support is so inconsistent across browsers. I only tested on Linux, and Chrome was the only browser where it worked. I still have a condition for Firefox/Linux but I'll remove it if I don't find a way to detect if it's pressed when scanning.

The modifier keys are checked from the mouse event because it's difficult to track pressed keys when focus shifts between popups and page content.

Not sure what should be done with query parser. I was planning to add some profile identifier query parameter for search page, but that's already kind of supported by the URL RegExp condition as long as unknown parameters aren't removed when dynamically changing the URL. It has some discoverability issues, though. Probably not a concern for this PR, currently modifers are just ignored on the search page.

### DIFF
--- a/ext/mixed/js/text-scanner.js
+++ b/ext/mixed/js/text-scanner.js
@@ -71,7 +71,7 @@ class TextScanner extends EventDispatcher {
         }
 
         const modifiers = DOM.getActiveModifiers(e);
-        this.trigger('updateActiveModifiers', {modifiers});
+        this.trigger('activeModifiersChanged', {modifiers});
 
         const scanningOptions = this.options.scanning;
         const scanningModifier = scanningOptions.modifier;
--- a/ext/mixed/js/text-scanner.js
+++ b/ext/mixed/js/text-scanner.js
@@ -70,6 +70,9 @@ class TextScanner extends EventDispatcher {
             return;
         }
 
+        const modifiers = this.getActiveModifiers(e);
+        this.trigger('updateActiveModifiers', {modifiers});
+
         const scanningOptions = this.options.scanning;
         const scanningModifier = scanningOptions.modifier;
         if (!(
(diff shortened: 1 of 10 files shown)


siikamiika:
This feature highlights an issue with the way `TextSourceRange` is used in Frontend.

```javascript
async findTerms(textSource, optionsContext) {
    this.setTextSourceScanLength(textSource, this.options.scanning.length);

    const searchText = textSource.text();
    if (searchText.length === 0) { return null; }

    const {definitions, length} = await apiTermsFind(searchText, {}, optionsContext);
    if (definitions.length === 0) { return null; }

    textSource.setEndOffset(length);

    return {definitions, type: 'terms'};
}
```

If the text is selected before going to this function, which will happen when options are updated, `options.scanning.length`characters  of text will be selected until `apiTermsFind` returns. This wasn't a big issue before, but will be a lot more noticeable with the new feature.

ext/mixed/js/text-scanner.js
```diff
@@ -60,6 +60,9 @@ class TextScanner extends EventDispatcher {
             return;
         }
 
+        const modifiers = DOM.getActiveModifiers(e);
+        this.trigger('updateActiveModifiers', {modifiers});
```
toasted-nutbread, line 64: Rename event to be more event-like: `activeModifiersChange` or `activeModifiersChanged`

(I forget if we have a consistent pattern for tense on event names; I have the feeling we don't.)

toasted-nutbread:
Don't see any big issues so far. I'll merge my PRs in so you can rebase and I'll check again.

ext/mixed/js/text-scanner.js
siikamiika: I've seen both, but I renamed it to the past tense https://github.com/FooSoft/yomichan/pull/487/commits/28afafd568a8e58d86c96ea38edc1ff8093ec94e

siikamiika:
I just realized that because the options are saved using their name separated with a ' + ', you cannot rename the modifiers based on the platform without breaking existing options unless you do a migration. I think this is probably something we want to address before releasing this.

toasted-nutbread:
> I just realized that because the options are saved using their name separated with a ' + ', you cannot rename the modifiers based on the platform without breaking existing options unless you do a migration. I think this is probably something we want to address before releasing this.

Wouldn't you just need to have a distinction between the name and the display name, since the event key fields are always the same?

siikamiika:
> Wouldn't you just need to have a distinction between the name and the display name, since the event key fields are always the same?

There is currently and I assumed that the transform functions in the descriptor would take care of converting the data before saving it to the options object, but I must have failed to do some step required to make that happen.

Issue #441, opened 2020-04-10
Improve Yomichan popup iframe security


toasted-nutbread:
Related: #353

This is a reminder issue to update the iframe to be dynamically loaded rather than using a `src` URL.

Old in-progress branch: https://github.com/toasted-nutbread/yomichan/tree/iframe-url-security

siikamiika:
I actually thought this was resolved earlier, but a while ago I tested that it's still possible to get the iframe `src` from the page. Were there some issues with getting blank popups?

toasted-nutbread:
There were some issues which weren't handled which made me have to revert it.

36605f74c3d48a98500d198636112454b36063f6

PR #488, opened 2020-04-30
Don't parent the popup frame to elements which cause unload

toasted-nutbread:
Fixes some bugs which cause the popup frame to unload when the fullscreen element is changed. The fullscreen element may be reported as an `<iframe>` or an element with a shadow DOM. If the popup frame is parented to one of these elements, it is unloaded. There are potentially others that this could happen on, but his handles a few edge cases in test-document2.html.

This does not handle elements with closed shadow DOMs, as those aren't (supposed to be) detectable. So it is still possible for the frame to unload. However, this issue needs to be addressed eventually for #441 and some other tasks I'm working on.

### DIFF
--- a/ext/fg/js/popup.js
+++ b/ext/fg/js/popup.js
@@ -281,7 +281,7 @@ class Popup {
     }
 
     _onFullscreenChanged() {
-        const parent = (DOM.getFullscreenElement() || document.body || null);
+        const parent = this._getFrameParentElement();
         if (parent !== null && this._container.parentNode !== parent) {
             parent.appendChild(this._container);
         }
@@ -373,6 +373,22 @@ class Popup {
         if (token === null || contentWindow === null) { return; }
 
         contentWindow.postMessage({action, params, token}, this._targetOrigin);
+    }
+
+    _getFrameParentElement() {
+        const defaultParent = document.body;
+        const fullscreenElement = DOM.getFullscreenElement();
+        if (fullscreenElement === null || fullscreenElement.shadowRoot) {
+            return defaultParent;
+        }
+
+        switch (fullscreenElement.nodeName.toUpperCase()) {
+            case 'IFRAME':
+            case 'FRAME':
+                return defaultParent;
+        }
+
+        return fullscreenElement;
     }
 
     static _getPositionForHorizontalText(elementRect, width, height, viewport, offsetScale, optionsGeneral) {


siikamiika:
I got this error when first scanning manual Test 2 on Firefox (it works for some reason) and while the popup is open scanning Test 1. It happened on both master and this branch.

```
Yomichan v20.4.18.0 has encountered a problem.
Originating URL: file:///home/siikamiika/koodi/yomichan/test/data/html/test-document2.html#
WrongDocumentError: Node cannot be used in a document other than the one in which it was created
equals@moz-extension://689faae7-541f-45a7-a011-773bbfe31bad/fg/js/source.js:97:31
searchAt@moz-extension://689faae7-541f-45a7-a011-773bbfe31bad/mixed/js/text-scanner.js:288:79
async*search@moz-extension://689faae7-541f-45a7-a011-773bbfe31bad/mixed/js/text-scanner.js:80:24
onMouseMove@moz-extension://689faae7-541f-45a7-a011-773bbfe31bad/mixed/js/text-scanner.js:83:9
EventListener.handleEvent*addEventListener@moz-extension://689faae7-541f-45a7-a011-773bbfe31bad/mixed/js/core.js:248:14
hookEvents@moz-extension://689faae7-541f-45a7-a011-773bbfe31bad/mixed/js/text-scanner.js:242:33
setEnabled@moz-extension://689faae7-541f-45a7-a011-773bbfe31bad/mixed/js/text-scanner.js:229:18
setOptions@moz-extension://689faae7-541f-45a7-a011-773bbfe31bad/mixed/js/text-scanner.js:269:14
updateOptions@moz-extension://689faae7-541f-45a7-a011-773bbfe31bad/fg/js/frontend.js:153:14
async*prepare@moz-extension://689faae7-541f-45a7-a011-773bbfe31bad/fg/js/frontend.js:65:24
applyOptions@moz-extension://689faae7-541f-45a7-a011-773bbfe31bad/fg/js/content-script-main.js:129:47
async*@moz-extension://689faae7-541f-45a7-a011-773bbfe31bad/fg/js/content-script-main.js:147:11
async*@moz-extension://689faae7-541f-45a7-a011-773bbfe31bad/fg/js/content-script-main.js:148:3
Data: null

Issues can be reported at https://github.com/FooSoft/yomichan/issues
```

#490 seems to fix another issue where you'll have an empty popup if you scan anything while Test 1 or 2 (anything non-iframe) is fullscreen and close the fullscreen while a popup is still open.

toasted-nutbread:
> I got this error when first scanning manual Test 2 on Firefox (it works for some reason) and while the popup is open scanning Test 1. It happened on both master and this branch.

Good catch. Apparently due to the shadow DOM and root window's documents being different.
Spec: https://dom.spec.whatwg.org/#dom-range-compareboundarypoints

> #490 seems to fix another issue where you'll have an empty popup if you scan anything while Test 1 or 2 (anything non-iframe) is fullscreen and close the fullscreen while a popup is still open.

This is actually the same issue being addressed in different ways. This PR is proactively preventing (most of) the situations which cause popup unload. The other PR is reactively "fixing" the popup (by hiding it) when an unload is detected.

What is actually happening when the popup goes blank is the following:
* Popup is open
* Fullscreen is enabled
* Popup is reparented to `<iframe>`, which causes the node to not be rendered.
* This causes the frame to unload. (about:blank)
* When fullscreen is exited, the popup is moved back into the `<body>` element.
* The popup is visible again, but since a refresh occurred, no content is displayed.

siikamiika:
Got it, it was hard to see what was actually going on with this PR because the popup would be hidden when you clicked the fullscreen button that caused the issue that's fixed by this. The #490 issue was easier to reproduce visually.

toasted-nutbread:
The other symptom these changes address is an error printed to the console:

```
popup.js:375 Failed to execute 'postMessage' on 'DOMWindow': The target origin provided does not match the recipient window's origin ('null').
```

Again, the underlying issue is the same: the iframe's content has changed to `about:blank`, so the `targetOrigin` doesn't match.

siikamiika:
Cool! I was wondering why those errors occurred sometimes.

Issue #465, opened 2020-04-19
Settings page UI/UX


toasted-nutbread:
So now that I have finished most of the tasks I had set out to do for database-related functionality, some of the next tasks I will be focusing on are related to updating the settings page. The first task is to design some updates to the UI/UX in order to accommodate the increase in the amount of settings.

Related: #296, #319, #335, #407, #435

I have an initial concept mocked up which is roughly based on the designs of the settings pages in Chrome and Firefox. The goals are to improve navigation, organization, and descriptions of the various options, as well as visually supporting future features such as profile inheritance. A preliminary screenshot is included below; the icons are mostly placeholders. Feedback and suggestions from anyone and everyone is welcome.

![image](https://user-images.githubusercontent.com/11037431/79702439-d9317080-8272-11ea-9083-995f89cfc3b4.png)

siikamiika:
This looks great just like https://github.com/FooSoft/yomichan/issues/402!

Some thoughts about possible UX and UI improvements:

- Use consistent language for feature names
- Add more descriptions and examples, maybe even visual (if you know Synaptics or macOS touchpad settings :smile:)
- Anchor links to settings that are mentioned in other settings (there are some examples of this already)
- Highlight the current category
- Make dictionary sorting easier (drag/drop or arrows like profile sorting)
- https://github.com/FooSoft/yomichan/issues/436#issuecomment-610613115

siikamiika:
> maybe even visual (if you know Synaptics or macOS touchpad settings 😄)

I'm wondering if these could be scripted and rendered in the popup preview frame, and serve as tests at the same time.

toasted-nutbread:
> I'm wondering if these could be scripted and rendered in the popup preview frame, and serve as tests at the same time.

Maybe, but it would probably be very easy to encounter issues with that due to browser and device differences, especially if we are trying to do things like animations which are largely based on timing. For example, there is that issue in Chrome where fetching indexedDB counts takes several seconds, which could easily impact preview animations.

siikamiika:
Yeah, probably good to start small first. I think the preview frame could be visible regardless of your scroll position because it will help you when choosing many options. The ones requiring an actual animation are a minority.

siikamiika:
Continuation for my comments on https://github.com/FooSoft/yomichan/issues/479 that belong here instead.

Advanced/normal options:
- `dictionaries.priority` from advanced to normal.
[...]

PR #489, opened 2020-05-01
Add helpful link for markers in settings.html

allanlw:
Add a link to the flashcard marker definitions on the yomichan homepage in settings.html to help users who are setting up Anki integration.

### DIFF
--- a/ext/bg/settings.html
+++ b/ext/bg/settings.html
@@ -847,6 +847,8 @@
                             As Anki requires the first field in the model to be unique, it is recommended
                             that you set it to <code>{expression}</code> for term flashcards and <code>{character}</code> for
                             Kanji flashcards. You can use multiple markers per field by typing them in directly.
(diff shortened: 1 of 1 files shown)


siikamiika:
Thanks for the contribution! We're currently redesigning the settings in https://github.com/FooSoft/yomichan/issues/465, and this is something to consider when implementing that. It's a long project so probably you'll see this as it is in some future release rather than converted straight into the new design.

allanlw:
Sounds good! Mostly I know this tripped me up enough times setting up yomichan that I finally got annoyed enough to fix it.

toasted-nutbread:
Yeah, as I work on the updated settings, I'll keep this in mind. Since the descriptions are already in the [README.md](https://github.com/FooSoft/yomichan/blob/master/README.md#flashcard-configuration) file, I think it would make sense to also have them directly in the project.

PR #490, opened 2020-05-01
Remove and unload the popup frame if an unexpected load occurs

toasted-nutbread:
If the popup emits an unexpected `load` event, the popup is removed from the page and unloaded. Such events can occur if the popup is unloaded (see: #441, #488), or new content is loaded, such as assigning `src` or `srcdoc`. This should, for the most part, prevent hijacking. Technically it would be even more secure to create a new iframe and forget about the old one, but that's not necessary at this point (and also out of scope for this change).

### DIFF
--- a/ext/fg/js/popup.js
+++ b/ext/fg/js/popup.js
@@ -201,13 +201,19 @@ class Popup {
     // Private functions
 
     _inject() {
-        if (this._injectPromise === null) {
-            this._injectPromise = this._createInjectPromise();
-            this._injectPromise.then(
-                () => { this._injectPromiseComplete = true; },
-                () => { this._resetFrame(); });
+        let injectPromise = this._injectPromise;
+        if (injectPromise === null) {
+            injectPromise = this._createInjectPromise();
+            this._injectPromise = injectPromise;
+            injectPromise.then(
+                () => {
+                    if (injectPromise !== this._injectPromise) { return; }
+                    this._injectPromiseComplete = true;
+                },
+                () => { this._resetFrame(); }
+            );
         }
-        return this._injectPromise;
+        return injectPromise;
     }
 
     async _createInjectPromise() {
--- a/ext/fg/js/popup.js
+++ b/ext/fg/js/popup.js
@@ -261,8 +261,6 @@ class Popup {
         this._container.removeAttribute('src');
         this._container.removeAttribute('srcdoc');
 
-        this._containerSecret = null;
-        this._containerToken = null;
         this._injectPromise = null;
         this._injectPromiseComplete = false;
     }
--- a/ext/fg/js/popup.js
+++ b/ext/fg/js/popup.js
@@ -255,7 +255,7 @@ class Popup {
     }
 
     _onFrameLoad() {
-        if (!(this._injectPromise !== null && this._injectPromiseComplete)) { return; }
+        if (!this._injectPromiseComplete) { return; }
         this._resetFrame();
     }
 
--- a/ext/fg/js/popup.js
+++ b/ext/fg/js/popup.js
@@ -31,6 +31,7 @@ class Popup {
         this._child = null;
         this._childrenSupported = true;
         this._injectPromise = null;
+        this._injectPromiseComplete = false;
         this._visible = false;
         this._visibleOverride = null;
         this._options = null;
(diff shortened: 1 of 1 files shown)


ext/fg/js/popup.js
```diff
@@ -243,6 +248,25 @@ class Popup {
         return popupPreparedPromise;
     }
 
+    _onFrameLoad() {
+        if (!(this._injectPromise !== null && this._injectPromiseComplete)) { return; }
```
siikamiika, line 252: `this._injectPromise` should never be `null` when `this._injectPromiseComplete` is `true`

ext/fg/js/popup.js
```diff
@@ -243,6 +248,25 @@ class Popup {
...
+    _resetFrame() {
+        const parent = this._container.parentNode;
+        if (parent !== null) {
+            parent.removeChild(this._container);
+        }
+        this._container.removeAttribute('src');
+        this._container.removeAttribute('srcdoc');
+
+        this._containerSecret = null;
+        this._containerToken = null;
```
siikamiika, line 265: These don't seem to be used anywhere else, did you mean `_messageToken` perhaps?

toasted-nutbread: Oops; leftovers from a different commit I was based on. You'll probably see these in a future commit, but for now, they aren't used.
25243f33207ba9d853afdbbbc9633ff02a117a79

ext/fg/js/popup.js
```diff
@@ -201,6 +203,9 @@ class Popup {
     _inject() {
         if (this._injectPromise === null) {
             this._injectPromise = this._createInjectPromise();
+            this._injectPromise.then(
+                () => { this._injectPromiseComplete = true; },
+                () => { this._resetFrame(); });
```
siikamiika, line 208: Should the closing paren be on a new line? Don't mind this too much but I've seen that kind of style more.

ext/fg/js/popup.js
toasted-nutbread: Looking at it again, the `this._injectPromise` check may actually not be necessary, since `this._injectPromiseComplete` being `true` is predicated on `this._injectPromise` existing.

The way the code is used currently doesn't allow for this, but technically `this._injectPromiseComplete` could be set to `true` if `this._injectPromise` was reassigned due to anything other than that promise resolving. I may add a guard for this, since the code may potentially allow that case in the future; easier to remember now than later.

ext/fg/js/popup.js
toasted-nutbread: Oops again; yes it should. Different language styles mixing me up.

ext/fg/js/popup.js
toasted-nutbread: Guard added in b13957ea8efa58bd2d88615262282fecad151f61.

ext/fg/js/popup.js
toasted-nutbread: Style fixed in b13957ea8efa58bd2d88615262282fecad151f61.

siikamiika: Are you secretly writing Lisp somewhere )))

siikamiika: The new style looks good :+1:

toasted-nutbread: Thankfully no; it stems from whether languages/conventions prefer brackets on the same line as the expression, or on a new line.

```js
// Same line
if (
    something1 &&
    something2
) {
    // ...
}

// New line
if (
    something1 &&
    something2)
{
    // ...
}

// Yuck
if (
    something1 &&
    something2
)
{
    // ..
}
```

ext/fg/js/popup.js
siikamiika: Thanks!

toasted-nutbread: Redundant check removed in ed8b438836eda6ec4532bd98c2d7b425c0870393.

ext/fg/js/popup.js
siikamiika: Ok, I've seen lots of C code to put brackets on a new line. Java/JS/PHP etc seems to prefer the same line. Not sure how I'd handle multi line ifs with the new line style. Maybe I'd create a variable for the boolean before testing it in more complex cases.

PR #491, opened 2020-05-01
Catch WrongDocumentError thrown by compareBoundaryPoints

toasted-nutbread:
Fixes https://github.com/FooSoft/yomichan/pull/488#issuecomment-622607016
https://dom.spec.whatwg.org/#dom-range-compareboundarypoints

### DIFF
--- a/ext/fg/js/source.js
+++ b/ext/fg/js/source.js
@@ -97,9 +97,11 @@ class TextSourceRange {
             try {
                 return this.range.compareBoundaryPoints(Range.START_TO_START, other.range) === 0;
             } catch (e) {
-                // Potential WrongDocumentError due to the ranges being in different documents.
-                // This can happen with shadow DOMs.
-                return false;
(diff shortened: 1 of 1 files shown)


ext/fg/js/source.js
```diff
@@ -94,7 +94,13 @@ class TextSourceRange {
...
             );
         } else {
-            return this.range.compareBoundaryPoints(Range.START_TO_START, other.range) === 0;
+            try {
+                return this.range.compareBoundaryPoints(Range.START_TO_START, other.range) === 0;
+            } catch (e) {
+                // Potential WrongDocumentError due to the ranges being in different documents.
+                // This can happen with shadow DOMs.
+                return false;
+            }
```
siikamiika, line 103: Should this check `e.name` to avoid silencing other errors?

toasted-nutbread: c4b992408b5b16b9d4a3a199cad66cf20d73d6a9

PR #494, opened 2020-05-02
Style fixes

toasted-nutbread:
Fixed a few styles with the same issue pointed out in https://github.com/FooSoft/yomichan/pull/490#discussion_r418788009. Searched for an eslint rule but didn't find one. Added a few related eslint rules instead.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -53,6 +53,7 @@
         "comma-spacing": ["error", { "before": false, "after": true }],
         "computed-property-spacing": ["error", "never"],
         "func-call-spacing": ["error", "never"],
+        "function-paren-newline": ["error", "multiline-arguments"],
         "generator-star-spacing": ["error", "before"],
         "key-spacing": ["error", {"beforeColon": false, "afterColon": true, "mode": "strict"}],
         "keyword-spacing": ["error", {"before": true, "after": true}],
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -23,6 +23,7 @@
         "eqeqeq": "error",
         "func-names": ["error", "always"],
         "guard-for-in": "error",
+        "new-parens": "error",
         "no-case-declarations": "error",
         "no-const-assign": "error",
         "no-constant-condition": "off",
@@ -31,6 +32,7 @@
         "no-prototype-builtins": "error",
         "no-shadow": ["error", {"builtinGlobals": false}],
         "no-undef": "error",
+        "no-unexpected-multiline": "error",
         "no-unneeded-ternary": "error",
         "no-unused-vars": ["error", {"vars": "local", "args": "after-used", "argsIgnorePattern": "^_", "caughtErrors": "none"}],
         "no-unused-expressions": "error",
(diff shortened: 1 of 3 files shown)


.eslintrc.json
```diff
@@ -31,6 +32,7 @@
         "no-prototype-builtins": "error",
         "no-shadow": ["error", {"builtinGlobals": false}],
         "no-undef": "error",
+        "no-unexpected-multiline": "error",
```
siikamiika, line 35: Tested that code like

```javascript
[1, 2, 3]
    .filter((i) => i > 1)
    .map((i) => `[ ${i} ]`)
    .join('; ');
```

works with this rule. Could there be other styles that unexpectedly don't work?

.eslintrc.json
```diff
@@ -53,6 +56,7 @@
         "comma-spacing": ["error", { "before": false, "after": true }],
         "computed-property-spacing": ["error", "never"],
         "func-call-spacing": ["error", "never"],
+        "function-paren-newline": ["error", "multiline-arguments"],
```
siikamiika, line 59: `search.js`:

```javascript
this._onKeyDownIgnoreKeys = new Map([
    ['ANY_MOD', new Set([
        'Tab', 'ArrowUp', 'ArrowDown', 'ArrowLeft', 'ArrowRight', 'PageDown', 'PageUp', 'Home', 'End',
        'F1', 'F2', 'F3', 'F4', 'F5', 'F6', 'F7', 'F8', 'F9', 'F10',
        'F11', 'F12', 'F13', 'F14', 'F15', 'F16', 'F17', 'F18', 'F19', 'F20',
        'F21', 'F22', 'F23', 'F24'
    ])],
    ['Control', new Set(['C', 'A', 'Z', 'Y', 'X', 'F', 'G'])],
    ['Meta', new Set(['C', 'A', 'Z', 'Y', 'X', 'F', 'G'])],
    ['OS', new Set()],
    ['Alt', new Set()],
    ['AltGraph', new Set()],
    ['Shift', new Set()]
]);
```

Do you consider this something that can be done even if function argument newline consistency is enforced?

.eslintrc.json
toasted-nutbread: Guess we'll know if we ever encounter one ¯\\\_(ツ)\_/¯. I just know there's been a few times where I've been duplicating lines of code and done something like this. It's usually caught by other rules related to indent, but may as well explicitly enforce it.

```js
const x = [
    [something],
    [something],
    [something],
    // ...
    [something]
    [something] // <= copy-pasted
]

.eslintrc.json
toasted-nutbread: All of that is inside `[]` brackets, so that's all fine. I consider functions to be a special case, since they typically have a limited and explicit number of arguments.

Whenever I'm doing multiline function arguments, I prefer to have one per line. There was only the one place in the project where this wasn't being done, so it felt consistent enough to enable the rule.

.eslintrc.json
siikamiika: Ok. That specific issue would also be solved by having `comma-dangle` set to `always`.

.eslintrc.json
siikamiika: Yeah I agree that this is a bit different from functions because in functions the position usually has a meaning and the amount is limited like you said. Here it's just about grouping related things together. Speaking of grouping, I think it would make more sense if it was grouped by 12 because that's the amount of function keys on the keyboard.

PR #497, opened 2020-05-02
Popup init update

toasted-nutbread:
This PR changes the way that the popup and frame are initialized. The goal is to remove the use of [`Backend.messageToken`](https://github.com/FooSoft/yomichan/blob/d4ae9aa501ece99ea6c5e6b8fb01c3005f5b7f03/ext/bg/js/backend.js#L76) as the mechanism for ensuring message authenticity. There are two reasons for this:
* Manifest v3 discourages storing state in the service worker, as it will be started/stopped without notice. `messageToken` is effectively a state, and if it's unexpectedly changed, that could have negative consequences. (#455)
* Since the popup can be unloaded, the initialization sequence will need to be re-executed. This could also cause issues with a stale state. (#441)

Follow-up of #488, #490

(Apologies in advance for yet another lengthy description.)

## Changes

* Adds an `sendMessageToFrame` API, which sends a message to a specific frameId in the same tab.
* Removes the `getMessageToken` API.
* `DisplayFloat._windowMessageHandlers` now stores additional information about the handler. The main change is that message authentication is opt-out. If the `authenticate` field is set to `false`, the sender is not validated before running the handler.
* The old `Popup.prepare` has been split into three different steps: `prepare`, `initialize`, and `configure`.
  * The new `prepare` sets up event listeners and is called in `float-main.js`'s entry point.
  * `initialize` is an authentication step (described in more detail later) to ensure messages are coming from Yomichan and nothing else.
  * `configure` assigns various settings and information.

## Initialization

The new initialization process replaces the old `messageToken` mechanism. It is effectively a handshake operation, which is described below.

(Note: I have tried to lay the steps out sequentially, but it is really a two-part parallel operation, since both the content script and the frame are running simultaneously.)

1. `Popup` sets up the frame to load, i.e. by setting `src` and adding the `<iframe>` element to the document.
2. In the frame, `DisplayFloat` now has a new unique ID field named `secret` which is constant for its lifetime.
3. Once the `DisplayFloat` has been fully `prepare`'d, it uses `apiBroadcastTab` to broadcast a `prepareCompleted` message with its `secret` data. (This broadcast is only seen by the extension code for that tab.)
4. `Popup` waits for a `prepareCompleted` message.
   * When `Popup` receives a `prepareCompleted` message, it posts an `initialize` message directly to the frame's `contentWindow`. This message contains both the `secret` another new unique ID called `token`. (This message is only seen by the extension code for that `frameId`.)
5. When `DisplayFloat` receives an `initialize` message, it checks if the `secret` matches its secret.
   * If the `secret` does not match, the message is discarded and processing stops.
   * If the `secret` does match, `DisplayFloat` stores the `token` from the message. `DisplayFloat` sends an `initializeCompleted` message to the `frameId` which sent the the `initialize` message, with both the `secret` and `token`. **`DisplayFloat` is now authenticated.**
6. When `Popup` receives `initializeCompleted`, it checks to see if the `secret` and `token` matches.
   * If the `secret` and `token` don't both match, the message is discarded and processing stops.
   * If the `secret` and `token` both match, `Popup` stores both values. **`Popup` is now authenticated.**
7. Initialization is now complete. All future message exchanges use these unique IDs to verify the message is coming from the right source.
8. There is also a timeout and handling for error/unexpected states.

## Explanation

The reason why this works is due to the secure messaging channel provided by the WebExtension framework. The underlying content page never sees any messages posted to it, so it would never be able to determine what the `secret`/`token` is. Therefore, it will never be able to send messages to the frame.

The use of `contentWindow.postMessage` in step 4 is used as filtering mechanism in case multiple popups are initializing at the same time. This way, the message is guaranteed to _only_ be sent to the frame which `Popup` is attempting to initialize. Contrasted with a `apiBroadcastTab` or `sendMessageToFrame`, both of those functions have no way of knowing that the `frameId` corresponds to the `<iframe>` element.

(Aside: this may also make it possible to improve the `FrameOffsetForwarder`, since there is now a direct mapping of `frameId` to `<iframe>`. Out of scope for this change though.)

### DIFF
--- a/ext/fg/js/float.js
+++ b/ext/fg/js/float.js
@@ -30,8 +30,6 @@ class DisplayFloat extends Display {
         this._secret = yomichan.generateId(16);
         this._token = null;
 
-        this._popupId = null;
-
         this._orphaned = false;
         this._prepareInvoked = false;
 
@@ -48,11 +46,11 @@ class DisplayFloat extends Display {
 
         this._windowMessageHandlers = new Map([
             ['initialize', {handler: this._initialize.bind(this), authenticate: false}],
+            ['configure', {handler: this._configure.bind(this)}],
             ['setOptionsContext', {handler: ({optionsContext}) => this.setOptionsContext(optionsContext)}],
             ['setContent', {handler: ({type, details}) => this.setContent(type, details)}],
             ['clearAutoPlayTimer', {handler: () => this.clearAutoPlayTimer()}],
             ['setCustomCss', {handler: ({css}) => this.setCustomCss(css)}],
-            ['prepare', {handler: ({popupInfo, optionsContext, childrenSupported, scale}) => this.prepare(popupInfo, optionsContext, childrenSupported, scale)}],
             ['setContentScale', {handler: ({scale}) => this.setContentScale(scale)}]
         ]);
 
@@ -60,28 +58,6 @@ class DisplayFloat extends Display {
         window.addEventListener('message', this.onMessage.bind(this), false);
 
         apiBroadcastTab('popupPrepared', {secret: this._secret});
-    }
-
-    async prepare(popupInfo, optionsContext, childrenSupported, scale) {
-        if (this._prepareInvoked) { return; }
-        this._prepareInvoked = true;
-
-        const {id, parentFrameId} = popupInfo;
-        this._popupId = id;
-
-        this.optionsContext = optionsContext;
-
-        await super.prepare();
-        await this.updateOptions();
-
-        if (childrenSupported) {
-            const {depth, url} = optionsContext;
-            popupNestedInitialize(id, depth, parentFrameId, url);
-        }
-
-        this.setContentScale(scale);
-
-        apiBroadcastTab('popupPrepareCompleted', {targetPopupId: this._popupId});
     }
 
     onError(error) {
@@ -181,6 +157,25 @@ class DisplayFloat extends Display {
         this._token = token;
 
         apiSendMessageToFrame(frameId, 'popupInitialized', {secret, token});
+    }
+
+    async _configure({messageId, frameId, popupId, optionsContext, childrenSupported, scale}) {
+        if (this._prepareInvoked) { return; }
+        this._prepareInvoked = true;
+
+        this.optionsContext = optionsContext;
+
+        await super.prepare();
+        await this.updateOptions();
+
+        if (childrenSupported) {
+            const {depth, url} = optionsContext;
+            popupNestedInitialize(popupId, depth, frameId, url);
+        }
+
+        this.setContentScale(scale);
+
+        apiSendMessageToFrame(frameId, 'popupConfigured', {messageId});
     }
 
     _isMessageAuthenticated(message) {
--- a/ext/fg/js/float.js
+++ b/ext/fg/js/float.js
@@ -86,15 +86,27 @@ class DisplayFloat extends Display {
 
     onMessage(e) {
         const data = e.data;
-        if (typeof data !== 'object' || data === null) { return; } // Invalid data
+        if (typeof data !== 'object' || data === null) {
+            this._logMessageError(e, 'Invalid data');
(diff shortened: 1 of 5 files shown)


ext/fg/js/popup.js
```diff
@@ -287,8 +315,11 @@ class Popup {
             }, timeout);
 
             chrome.runtime.onMessage.addListener(onMessage);
-            frame.addEventListener('load', () => containerLoadedResolve());
-            frame.addEventListener('error', () => containerLoadedReject(new Error('Failed to load container')));
```
siikamiika: Why was `error` handler removed here?

ext/fg/js/float.js
```diff
@@ -193,4 +161,45 @@ class DisplayFloat extends Display {
...
+        if (childrenSupported && !this._initializedNestedPopups) {
+            const {depth, url} = optionsContext;
+            popupNestedInitialize(popupId, depth, frameId, url);
+            this._initializedNestedPopups = true;
+        }
+
+        this.setContentScale(scale);
+
+        apiSendMessageToFrame(frameId, 'popupConfigured', {messageId});
+    }
```
siikamiika, line 196: Because the `_initializedNestedPopups` guard was moved after `await`, the function can be called multiple times. None of the lines except the block initializing nested popups seem to misbehave when called multiple times, but it could be more clear to rename the guard and put it to the beginning.

ext/fg/js/popup.js
```diff
@@ -216,40 +216,154 @@ class Popup {
...
+                                    cleanup();
+                                    resolve({secret, token});
+                                }
+                            }
+                            break;
+                    }
+                } catch (e) {
+                    cleanup();
+                    reject(e);
+                }
```
siikamiika, line 281: Are there exceptions other than the explicit ones in `_initializeFrame` that can trigger this code? `cleanup` will remove the listener completely.

siikamiika:
The PR appears to work fine with all manual tests when rebased to master. Nice work cleaning up the naturally evolved `popupId` identification and replacing it with a proper handshake! I can try to do further development on `FrameOffsetForwarder` after this is merged unless you already have something in mind.

siikamiika:
There's at least one other place in `SettingsPopupPreview` that adds an event listener in the constructor and has prepare called from another frame using the listener. If I understand this PR correctly, the event listener should be activated in `prepare`, but because the popup preview is on the settings page it doesn't need protection against hijacking and the current `prepare` could be renamed to just `initialize` or `configure`. The handshake could also be extracted into a class in the future if it needs to be reused in places that actually need it.

ext/fg/js/popup.js
toasted-nutbread: I don't think "error" is a real event for frames. You get load events even for invalid content.

ext/fg/js/float.js
toasted-nutbread: It was an intentional choice. I didn't see any reason to prevent "configure" from being called more than once, but there was an explicit reason to prevent popup initialization multiple times. My thought was that it acts sort of like a multi-setter/`setVariousOptionsAtOnce` type function. In practice, it's currently not called more than once, but it could be in case it ever becomes necessary.

ext/fg/js/popup.js
toasted-nutbread: `postMessage` can fail. It's intended to remove all listeners, timers, etc. because the main promise is rejected right after.

toasted-nutbread:
> There's at least one other place in `SettingsPopupPreview` that adds an event listener in the constructor and has prepare called from another frame using the listener. If I understand this PR correctly, the event listener should be activated in `prepare`, but because the popup preview is on the settings page it doesn't need protection against hijacking and the current `prepare` could be renamed to just `initialize` or `configure`. The handshake could also be extracted into a class in the future if it needs to be reused in places that actually need it.

Yep; no need for a big handshake there, just a clarification of function names.

PR #500, opened 2020-05-03
Extract textSource text inside TextScanner

siikamiika:
Resolves https://github.com/FooSoft/yomichan/pull/487#issuecomment-621438659

### DIFF
--- a/ext/mixed/js/text-scanner.js
+++ b/ext/mixed/js/text-scanner.js
@@ -318,21 +318,26 @@ class TextScanner extends EventDispatcher {
         }
     }
 
-    setTextSourceScanLength(textSource, length) {
+    getTextSourceContent(textSource, length) {
+        const originalLength = textSource.text().length;
+
         textSource.setEndOffset(length);
-        if (this.ignoreNodes === null || !textSource.range) {
-            return;
+
+        if (this.ignoreNodes !== null && textSource.range) {
+            length = textSource.text().length;
+            while (textSource.range && length > 0) {
+                const nodes = TextSourceRange.getNodesInRange(textSource.range);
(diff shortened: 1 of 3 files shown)


ext/mixed/js/text-scanner.js
```diff
@@ -318,21 +318,26 @@ class TextScanner extends EventDispatcher {
...
-            return;
-        }
 
-        length = textSource.text().length;
-        while (textSource.range && length > 0) {
-            const nodes = TextSourceRange.getNodesInRange(textSource.range);
-            if (!TextSourceRange.anyNodeMatchesSelector(nodes, this.ignoreNodes)) {
-                break;
+        if (this.ignoreNodes !== null && textSource.range) {
+            length = textSource.text().length;
```
toasted-nutbread, line 327: `length = textSource.text().length;` => `length = originalLength;`

ext/mixed/js/text-scanner.js
```diff
@@ -318,21 +318,26 @@ class TextScanner extends EventDispatcher {
         }
     }
 
-    setTextSourceScanLength(textSource, length) {
+    getTextSourceContent(textSource, length) {
```
toasted-nutbread, line 321: Is this function intended to leave textSource having the same state as what it entered the function with? If so, probably best to operate on a clone.

siikamiika: Yep, that's the idea. I didn't know that clones would have their text selection detached from the page. Thanks! https://github.com/FooSoft/yomichan/pull/500/commits/d6567a3c1d4b3dae0b05ace3b11aad5ceeefba1a

PR #501, opened 2020-05-03
Add apiModifySettings

toasted-nutbread:
Replace `apiOptionsSet` with new `apiModifySettings`, which uses `ObjectPropertyAccessor` and the options schema. This new function will be used as the basis for #335.

### DIFF
--- a/ext/bg/js/backend.js
+++ b/ext/bg/js/backend.js
@@ -964,9 +964,19 @@ class Backend {
     }
 
     // Utilities
+
+    _getModifySettingObject(target) {
+        switch (target.scope) {
+            case 'profile':
+                if (!isObject(target.optionsContext)) { throw new Error('Invalid optionsContext'); }
+                return this.getOptions(target.optionsContext, true);
+            case 'global':
+                return this.getFullOptions(true);
+        }
+    }
+
     async _modifySetting(target) {
-        const optionsContext = target.optionsContext;
-        const options = isObject(optionsContext) ? this.getOptions(optionsContext, true) : this.getFullOptions(true);
+        const options = this._getModifySettingObject(target);
         const accessor = new ObjectPropertyAccessor(options);
         const action = target.action;
         switch (action) {
--- a/ext/bg/js/backend.js
+++ b/ext/bg/js/backend.js
@@ -276,16 +276,17 @@ class Backend {
     }
 
     getProfile(optionsContext) {
-        const profiles = this.options.profiles;
+        const options = this.getFullOptions();
+        const profiles = options.profiles;
         if (typeof optionsContext.index === 'number') {
             return profiles[optionsContext.index];
         }
-        const profile = this.getProfileFromContext(optionsContext);
-        return profile !== null ? profile : this.options.profiles[this.options.profileCurrent];
+        const profile = this.getProfileFromContext(options, optionsContext);
+        return profile !== null ? profile : options.profiles[options.profileCurrent];
     }
 
-    getProfileFromContext(optionsContext) {
-        for (const profile of this.options.profiles) {
+    getProfileFromContext(options, optionsContext) {
+        for (const profile of options.profiles) {
             const conditionGroups = profile.conditionGroups;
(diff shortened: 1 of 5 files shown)


ext/bg/js/backend.js
```diff
@@ -967,6 +944,50 @@ class Backend {
     }
 
     // Utilities
+    async _modifySetting(target) {
+        const optionsContext = target.optionsContext;
+        const options = isObject(optionsContext) ? this.getOptions(optionsContext, true) : this.getFullOptions(true);
```
siikamiika, line 949: Maybe full options should be explicitly enabled in case the API user forgets to supply context and breaks things

ext/bg/js/backend.js
```diff
@@ -809,6 +772,20 @@ class Backend {
...
+        const results = [];
+        for (const target of targets) {
+            try {
+                this._modifySetting(target);
+                results.push({result: true});
+            } catch (e) {
+                results.push({error: errorToJson(e)});
+            }
+        }
+        await this._onApiOptionsSave({source});
```
siikamiika, line 785: Is the `source` argument actually used anywhere?

ext/bg/js/backend.js
```diff
@@ -969,6 +985,50 @@ class Backend {
...
+                    if (typeof path2 !== 'string') { throw new Error('Invalid path2'); }
+                    accessor.swap(ObjectPropertyAccessor.getPathArray(path1), ObjectPropertyAccessor.getPathArray(path2));
+                }
+                break;
+            case 'splice':
+                {
+                    const {path, start, deleteCount, items} = target;
+                    if (typeof path !== 'string') { throw new Error('Invalid path'); }
+                    if (typeof start !== 'number' || Math.floor(start) !== start) { throw new Error('Invalid start'); }
+                    if (typeof deleteCount !== 'number' || Math.floor(deleteCount) !== start) { throw new Error('Invalid deleteCount'); }
```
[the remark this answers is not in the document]
toasted-nutbread: Correct. 444da26cae343056d976c9f5f7d8a208cbde2309

ext/bg/js/backend.js
toasted-nutbread: Updated. c12887daf3e4b81233a5fdc39af951de53267fe2

ext/bg/js/backend.js
toasted-nutbread: It's used on the settings page to determine what the source was for a settings change. It uses this value to not get into infinite loops of updating values due to a change that was triggered by itself.

Would be good to get rid of if possible, but the current setup made that difficult at the time.

ext/bg/js/backend.js
```diff
@@ -968,6 +945,60 @@ class Backend {
...
 
+    _getModifySettingObject(target) {
+        switch (target.scope) {
+            case 'profile':
+                if (!isObject(target.optionsContext)) { throw new Error('Invalid optionsContext'); }
+                return this.getOptions(target.optionsContext, true);
+            case 'global':
+                return this.getFullOptions(true);
+        }
+    }
```
siikamiika, line 956: This is good for correctness, but looking at the object "signature" it feels like there could be a separate `apiModifySettingsGlobal` in addition to `apiModifySettingsProfile`. Alternatively classes that use `apiModifySettings` could wrap the function so that it automatically uses the `optionsContext`, `action`, `scope` and `source` that's needed. Don't have to do this now but if this starts to get used more somewhere it could be useful.

Also this could throw by default.

toasted-nutbread: > Also this could throw by default.

Forgot to do that. b2908abea1418da85521fc739ab4ed28f89ca9e6

> This is good for correctness, but looking at the object "signature" it feels like there could be a separate `apiModifySettingsGlobal` in addition to `apiModifySettingsProfile`.

I think the use of this will probably make a bit more sense once it's used on the settings page. Effectively there will be a `<input ... data-scope="global">` for some of the options. It could be done the way you mentioned as well, but this might be more simple. We'll see as it becomes used.

PR #502, opened 2020-05-03
Support meta as scanning modifier

siikamiika:
It was added to profile conditions in https://github.com/FooSoft/yomichan/pull/487, so also support it as the scanning modifier key. Same thing applies here, it doesn't work on Firefox, so the option is conditionally hidden.

### DIFF
--- a/ext/bg/data/options-schema.json
+++ b/ext/bg/data/options-schema.json
@@ -351,7 +351,7 @@
                                     },
                                     "modifier": {
                                         "type": "string",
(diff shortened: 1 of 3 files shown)


siikamiika:
A thing I noticed while testing events on macOS: Firefox actually supports both `Meta` key event and `MouseEvent/KeyEvent.metaKey`. The support is only removed for Windows and Linux. I'll do something about that later, not in this PR.

Issue #503, opened 2020-05-03
Save modifier key profile condition value as a list when the operator is "are" or "areNot"


siikamiika:
https://github.com/FooSoft/yomichan/pull/487#issuecomment-623107824

Prevents changing the names of the modifier keys without breaking existing settings.

PR #506, opened 2020-05-03
Fix modifier key condition data

siikamiika:
Fixes https://github.com/FooSoft/yomichan/issues/503.

Saves the transformed value of the input as a profile condition instead of the transformed and reversed value. Since the default value changes from `''` to `[]`, all default values have to be isolated.

Can be used in the future by supplying `transformInput` into the descriptor in place of `transform`.

### DIFF
--- a/ext/bg/js/conditions.js
+++ b/ext/bg/js/conditions.js
@@ -16,12 +16,14 @@
  */
 
 
-function conditionsValidateOptionValue(object, value) {
+function conditionsValidateOptionValue(object, value, isInput=false) {
     if (hasOwn(object, 'validate') && !object.validate(value)) {
         throw new Error('Invalid value for condition');
     }
 
-    if (hasOwn(object, 'transform')) {
+    if (isInput && hasOwn(object, 'transformInput')) {
+        value = object.transformInput(value);
+    } else if (hasOwn(object, 'transform')) {
         value = object.transform(value);
 
         if (hasOwn(object, 'validateTransformed') && !object.validateTransformed(value)) {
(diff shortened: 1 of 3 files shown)


siikamiika:
I'll try to refactor this a bit because I only realized after creating the PR that it broke existing profile condition saving and did some quick fixes.

siikamiika:
Should be more clear after https://github.com/FooSoft/yomichan/pull/506/commits/e96bbb088225118528346c17c5aa5c49967ab410. Sorry for the messy commits.

siikamiika:
A short explanation about the usage of `transformInput`.

If a condition value is validated with `ConditionsUI.Condition.validateValue(value, isInput=true)`, the normalization function will first check for `transformInput` in `conditionDescriptor` and `operatorDescriptor`. They're executed in order and the result is accumulated like with `transform`. If either of the descriptors had `transformInput`, the intermediate value from the result will be returned separately and saved as the condition value in options. The intermediate value will go through `transform` and `transformReverse` as usual and the result from that will be returned as usual.

When the condition value is read back from the options and validated before showing, `isInput` will not be true and `transformInput` is skipped even if it exists. The result will be the same because the value has already gone through `transformInput`.

-----------------

Should there be some tests to ensure that this doesn't break anything and works as expected? This adds some complexity but it was the simplest solution I was able to come up with.

Maybe the schema should also define the types for each condition separately so that you could enforce correct values for each condition.

toasted-nutbread:
> Should there be some tests to ensure that this doesn't break anything and works as expected? This adds some complexity but it was the simplest solution I was able to come up with.

I've been wondering for a while if the conditions functionality could be replaced with the JSON schema validation functionality. If extended with the appropriate metadata, I think it could, but I haven't looked deeply into the idea. This would make it easier to test some stuff since it would just be using an existing component which already has tests.

siikamiika:
> I've been wondering for a while if the conditions functionality could be replaced with the JSON schema validation functionality. If extended with the appropriate metadata, I think it could, but I haven't looked deeply into the idea. This would make it easier to test some stuff since it would just be using an existing component which already has tests.

Interesting, I hope this PR doesn't complicate things too much for that. I agree that reusing existing tests would be nice.

PR #507, opened 2020-05-04
Text scanning update

toasted-nutbread:
* Fixes a test with a missing parameter.
* Updates `docSentenceExtract` to not rescan text after moving the starting cursor. Instead, it scans bidirectionally away from the initial cursor. (This is related to what I commented on here: https://github.com/FooSoft/yomichan/pull/458#issuecomment-619281704)
* `TextSourceRange.setEndOffset` now has a `fromEnd` parameter to support the above. Since there are many other places using `setEndOffset`, I chose not to change those for now.

### DIFF
--- a/ext/fg/js/document.js
+++ b/ext/fg/js/document.js
@@ -159,7 +159,7 @@ function docSentenceExtract(source, extent) {
 
     const sourceLocal = source.clone();
     const position = sourceLocal.setStartOffset(extent);
-    sourceLocal.setEndOffset(position + extent);
+    sourceLocal.setEndOffset(extent * 2 - position, true);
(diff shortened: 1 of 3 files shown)


siikamiika:
This might have broken `{sentence}` extraction, I tested scanning 思いやられる in  https://github.com/FooSoft/yomichan/pull/524#issuecomment-626194778 and the word itself was missing from the sentence. I'll test a bit more soon.

siikamiika:
Yeah, it works on `d6ae322^`

toasted-nutbread:
Secret new feature: automatically cloze the word from the sentence!

I'll look into it.

PR #510, opened 2020-05-05
Japanese util refactor

toasted-nutbread:
This PR merges the two japanese.js files into a single file. The wanakana dependency is now optional and will throw a meaningful error if any methods try to use it when it's not present.

#413, #463


ext/mixed/js/japanese.js
```diff
@@ -89,63 +139,410 @@ const jp = (() => {
...
-                return false;
+        // Character code testing functions
+
+        isCodePointKanji(codePoint) {
+            return isCodePointInRanges(codePoint, CJK_UNIFIED_IDEOGRAPHS_RANGES);
+        }
+
+        isCodePointKana(codePoint) {
+            return isCodePointInRanges(codePoint, KANA_RANGES);
+        }
```
siikamiika, line 168: This is not used anymore since `isStringEntirelyKana` uses `isCodePointInRanges` directly

toasted-nutbread: True, I didn't notice that. I think it's probably fine to leave in though, since it's so simple.

toasted-nutbread:
I just realized, CI didn't run for this change... 🤔

siikamiika:
> I just realized, CI didn't run for this change... 🤔

You can try to rebase, I merged a PR

PR #515, opened 2020-05-07
Text scanner refactor

toasted-nutbread:
Basically just making things private. Similar to #513, I think the reorganization change should be done separately.

### DIFF
--- a/ext/mixed/js/text-scanner.js
+++ b/ext/mixed/js/text-scanner.js
@@ -57,14 +57,14 @@ class TextScanner extends EventDispatcher {
         this._canClearSelection = value;
     }
 
-    onMouseOver(e) {
+    _onMouseOver(e) {
         if (this.ignoreElements().includes(e.target)) {
-            this.scanTimerClear();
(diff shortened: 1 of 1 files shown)


ext/mixed/js/text-scanner.js
```diff
@@ -116,76 +116,76 @@ class TextScanner extends EventDispatcher {
...
 
-        this.primaryTouchIdentifier = null;
-        this.preventScroll = false;
-        this.preventNextClick = false;
+        this._primaryTouchIdentifier = null;
+        this._preventScroll = false;
+        this._preventNextClick = false;
         // Don't revert context menu and mouse down prevention,
         // since these events can occur after the touch has ended.
         // this.preventNextContextMenu = false;
```
siikamiika, line 203: The variables in these comments didn't get renamed yet

toasted-nutbread: I've updated this comment in general to not look like it accidentally commented out code.
29acbd6d0391ab9e68ece96b9f69a8f87110799b

PR #517, opened 2020-05-08
Text scanner refactor

toasted-nutbread:
* Fixes a bug with missing return value.
* Changes `TextScanner.onSearchSource` to a constructor argument.
* Simplifies `ignorePoints` function; is now optional.
* Groups public functions together.

### DIFF
--- a/ext/mixed/js/text-scanner.js
+++ b/ext/mixed/js/text-scanner.js
@@ -296,10 +296,8 @@ class TextScanner extends EventDispatcher {
                 return;
             }
 
-            for (const ignorePointFn of this._ignorePoints) {
-                if (await ignorePointFn(x, y)) {
-                    return;
-                }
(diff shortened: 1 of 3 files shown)


ext/mixed/js/text-scanner.js
```diff
@@ -296,10 +296,8 @@ class TextScanner extends EventDispatcher {
                 return;
             }
 
-            for (const ignorePointFn of this._ignorePoints) {
-                if (await ignorePointFn(x, y)) {
-                    return;
-                }
+            if (typeof this._ignorePoints === 'function' && await this._ignorePoints(x, y)) {
```
siikamiika: This could be called `ignorePoint` now that there's only one.

toasted-nutbread: 45d6cf49b3423dfdc8f6d7c19553b5a65c38bb58

siikamiika: Looks good :+1:

PR #521, opened 2020-05-09
Dynamic loader load style

toasted-nutbread:
Addresses https://github.com/FooSoft/yomichan/pull/518#issue-415011950:

> I am also thinking that it might be good to move `_injectStylesheet` into `dynamicLoader`.

### DIFF
--- a/ext/fg/js/popup.js
+++ b/ext/fg/js/popup.js
@@ -19,6 +19,7 @@
  * DOM
  * apiInjectStylesheet
  * apiOptionsGet
+ * dynamicLoader
  */
 
 class Popup {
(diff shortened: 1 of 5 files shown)


ext/fg/js/popup.js
```diff
@@ -713,71 +708,6 @@ class Popup {
...
-            if (styleNode !== null && styleNode.parentNode !== null) {
-                styleNode.parentNode.removeChild(styleNode);
-            }
-            styleNode = document.createElement(tagName);
-            styleNode.id = id;
-        }
-
-        // Update node style
-        if (isFile) {
-            styleNode.rel = value;
```
siikamiika, line 766: Was this broken before? It's changed to `styleNode.href = value;` in `loadStyle` which makes more sense.

siikamiika:
(I just moved, so I'll be a bit less active until the dust settles)

ext/fg/js/popup.js
toasted-nutbread: Yes; it was overlooked since that particular code path was never used.

PR #522, opened 2020-05-09
Add support for additional types of event listeners

toasted-nutbread:
There are some times where it would be convenient for `EventListenerCollection` to support different event listener styles, so this adds that. Not used yet, but may be soon.

### DIFF
--- a/ext/mixed/js/core.js
+++ b/ext/mixed/js/core.js
@@ -250,15 +250,35 @@ class EventListenerCollection {
         return this._eventListeners.length;
     }
 
-    addEventListener(node, type, listener, options) {
-        node.addEventListener(type, listener, options);
-        this._eventListeners.push([node, type, listener, options]);
+    addEventListener(object, ...args) {
+        object.addEventListener(...args);
+        this._eventListeners.push(['removeEventListener', object, ...args]);
+    }
+
+    addListener(object, ...args) {
+        object.addListener(args);
+        this._eventListeners.push(['addListener', object, args]);
+    }
+
+    on(object, ...args) {
+        object.on(args);
+        this._eventListeners.push(['off', object, args]);
     }
 
     removeAllEventListeners() {
         if (this._eventListeners.length === 0) { return; }
-        for (const [node, type, listener, options] of this._eventListeners) {
-            node.removeEventListener(type, listener, options);
+        for (const [type, object, ...args] of this._eventListeners) {
+            switch (type) {
+                case 'removeEventListener':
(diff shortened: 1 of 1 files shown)


ext/mixed/js/core.js
```diff
@@ -250,15 +250,35 @@ class EventListenerCollection {
...
+    on(object, ...args) {
+        object.on(args);
+        this._eventListeners.push(['off', object, args]);
     }
 
     removeAllEventListeners() {
         if (this._eventListeners.length === 0) { return; }
-        for (const [node, type, listener, options] of this._eventListeners) {
-            node.removeEventListener(type, listener, options);
+        for (const [type, object, ...args] of this._eventListeners) {
```
siikamiika, line 270: `type` might not be the best name for the name of the function to remove the event listener. Unless they have some established names, could just go with `removeFunctionName` or `addFunctionName`.

ext/mixed/js/core.js
```diff
@@ -250,15 +250,35 @@ class EventListenerCollection {
...
-        node.addEventListener(type, listener, options);
-        this._eventListeners.push([node, type, listener, options]);
+    addEventListener(object, ...args) {
+        object.addEventListener(...args);
+        this._eventListeners.push(['removeEventListener', object, ...args]);
+    }
+
+    addListener(object, ...args) {
+        object.addListener(args);
+        this._eventListeners.push(['addListener', object, args]);
```
siikamiika, line 260: Others use the remove function name, and the switch case below is for `removeListener`.

ext/mixed/js/core.js
```diff
@@ -250,15 +250,35 @@ class EventListenerCollection {
...
+                case 'removeEventListener':
+                    object.removeEventListener(...args);
+                    break;
+                case 'removeListener':
+                    object.removeListener(...args);
+                    break;
+                case 'off':
+                    object.off(...args);
+                    break;
+            }
```
siikamiika, line 281: `default: throw`

toasted-nutbread: a8b39b7e1726c288f52fdc7d9255803de33ce7e3

ext/mixed/js/core.js
toasted-nutbread: a8b39b7e1726c288f52fdc7d9255803de33ce7e3

ext/mixed/js/core.js
toasted-nutbread: a8b39b7e1726c288f52fdc7d9255803de33ce7e3

PR #527, opened 2020-05-09
Text source refactor

toasted-nutbread:
Addresses https://github.com/FooSoft/yomichan/pull/526#issue-415603889:

> The API of TextSourceElement needs to be updated to match TextSourceRange, but that's a bit of a larger task.

After both of these PRs, I will probably split `TextSourceRange/TextSourceElement.setEndOffset` into two functions:
* `setEndOffset`, which follows the behaviour of `fromEnd=true`
* `setLength`, which follows the behaviour of `fromEnd=false`

### DIFF
--- a/ext/fg/js/source.js
+++ b/ext/fg/js/source.js
@@ -358,13 +358,16 @@ class TextSourceRange {
  */
 
 class TextSourceElement {
-    constructor(element, content='') {
+    constructor(element, fullContent=null, start=0, end=0) {
         this.element = element;
-        this.content = content;
+        this.fullContent = (typeof fullContent === 'string' ? fullContent : TextSourceElement.getElementContent(element));
+        this.start = start;
+        this.end = end;
+        this.content = this.fullContent.substring(this.start, this.end);
     }
 
     clone() {
-        return new TextSourceElement(this.element, this.content);
+        return new TextSourceElement(this.element, this.fullContent, this.start, this.end);
     }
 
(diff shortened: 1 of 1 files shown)


siikamiika:
Since `TextSourceElement` is only used for buttons and images, I'll use this PR as a forum to say something I've thought of for a while.

While the text that's found in these elements is commonly shorter than that found in regular text-containing elements like `<p>`, it's still rare that you'll be able to scan everything that's in the element without moving the starting position. It's especially confusing on touch devices if you try to swipe a page from an image and get a popup.

So, either the feature could have a toggle, or there could be an imposter element of sorts that actually allowed text selection.

toasted-nutbread:
> Since `TextSourceElement` is only used for buttons and images, I'll use this PR as a forum to say something I've thought of for a while.
> 
> While the text that's found in these elements is commonly shorter than that found in regular text-containing elements like `<p>`, it's still rare that you'll be able to scan everything that's in the element without moving the starting position. It's especially confusing on touch devices if you try to swipe a page from an image and get a popup.
> 
> So, either the feature could have a toggle, or there could be an imposter element of sorts that actually allowed text selection.

Potentially a use case for `QueryParser` inside of the popup, just like the search page.

siikamiika:
> Potentially a use case for `QueryParser` inside of the popup, just like the search page.

Sounds like a good idea, especially for images. I can create a new issue for this.

PR #528, opened 2020-05-09
Use CSS variables for themes

toasted-nutbread:
CSS [`var()`](https://developer.mozilla.org/en-US/docs/Web/CSS/var) is [supported by our minimum browser versions](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties#Browser_compatibility), so we can use that. I plan on using that for settings as well. This change is only moving all the theme styles into display.css.


siikamiika:
> CSS var() is supported by our minimum browser versions, so we can use that.

Our minimum version for Firefox is 53, last increased when adding [`Node.isConnected`](https://developer.mozilla.org/en-US/docs/Web/API/Node/isConnected). However, it seems that `var()` has been experimental until version 55 although it would let you understand that it's been supported since 31:

> From version 29 until version 55 (exclusive): this feature is behind the layout.css.variables.enabled preference (needs to be set to true). To change preferences in Firefox, visit about:config.

toasted-nutbread:
Well, I definitely misread some versions then. However, I think bumping the min Firefox version to 55 should be fine. CSS variables are a very compelling feature that greatly simplifies a lot of things.

Unfortunately, we don't really have any user statistics about what versions of browsers users are on, so we can't really make judgements based on that. The general assumption I typically follow for is that if a feature has been available for 3-4 years, it should be pretty widely supported/adopted by the common browser userbases. (Firefox 55's release date was August 8, 2017)

siikamiika:
The lowest version of normal Firefox I can somehow imagine people be using is 56 because legacy extension support was dropped in 57. I don't think that's a reason to avoid increasing the minimum version number though because there are forks like Waterfox addressing the issue that are based on a newer version. Using an unsupported version you'll probably also be used to things looking broken and places telling you to upgrade your browser.

toasted-nutbread:
Updated to 55 in 96125589694f9744dca8d1987d29bf303ef145c1.

siikamiika:
There seems to be a difference in the dark theme popular term color:

dark-diff
light-diff

toasted-nutbread:
That was actually intentional, but perhaps misguided. I was matching the value of `--tag-popular-background-color`, which is darker than the light theme, but since that's for a background color, it doesn't work well for text. I've reverted it to the previous value.

Here's an overlaid comparison of the light (top) vs dark (bottom) colors:
image

e3fcc7ba0775ae4ec0390ff34321c3b41970f008

siikamiika:
I agree it was maybe a bit too dark for text color, althought that would have been consistent. I initially copied the color straight from the popular tag background here: https://github.com/FooSoft/yomichan/pull/95/files#diff-604777a6212329d0b021db678429e058R135

The color itself is not really light or dark but something in between, so it works for both themes. Probably not optimal but still. Maybe they should be reviewed at some point along with tag colors which I also changed somewhat arbitrarily at some point to distinguish parts of speech from other tag types.

PR #529, opened 2020-05-09
Load frame content using contentDocument.location.href instead of src attribute

### LINKED ISSUE #441: Improve Yomichan popup iframe security

toasted-nutbread:
Resolves #441 and part of #353.

The changes made to the handshake were made with this eventuality in mind, so the final change is just swapping a single line. Effectively an un-revert of 36605f74c3d48a98500d198636112454b36063f6.

### DIFF
--- a/ext/fg/js/popup.js
+++ b/ext/fg/js/popup.js
@@ -336,9 +336,9 @@ class Popup {
         const {secret, token} = await this._initializeFrame(this._frame, this._targetOrigin, this._frameId, (frame) => {
             frame.removeAttribute('src');
             frame.removeAttribute('srcdoc');
-            frame.setAttribute('src', chrome.runtime.getURL('/fg/float.html'));
             this._observeFullscreen(true);
             this._onFullscreenChanged();
+            frame.contentDocument.location.href = chrome.runtime.getURL('/fg/float.html');
         });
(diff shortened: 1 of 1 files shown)


siikamiika:
Actually there's one behavioral difference with this on Firefox, but it didn't even work properly in the first place. If you make manual test 2 (plain shadow dom) fullscreen and scan text in that, then close the fullscreen with esc, a blank popup will stay there. Before it would disappear when pressing esc. Anyway, it was never visible in fullscreen.

toasted-nutbread:
> Actually there's one behavioral difference with this on Firefox, but it didn't even work properly in the first place. If you make manual test 2 (plain shadow dom) fullscreen and scan text in that, then close the fullscreen with esc, a blank popup will stay there. Before it would disappear when pressing esc. Anyway, it was never visible in fullscreen. I think the security benefit is worth more in this case.

I'll check that out. I've noticed that test 2 works on Firefox but not on Chrome, so I wasn't too worried about how fullscreen didn't work properly for it. Probably worth checking out though.

PR #531, opened 2020-05-10
Cross frame communication

toasted-nutbread:
These new classes are intended to eventually replace `FrontendApiSender`, `FrontendApiReceiver`, and `BackendApiForwarder`. They serve the same need, but they are more robust and general-purpose.

The implementation is very similar as the classes listed above. The main differences are:
* `FrontendApiReceiver` and `FrontendApiSender` are now the same class (`CrossFrameAPI`), and should be a singleton. The reason is that they should share the same port for bidirectional communication.
* Improved error handling for some potential odd edge cases.
* `BackendApiForwarder` is just part of `Backend`.

This will be used to address the following:
* Moving `PopupProxy.getUrl`: https://github.com/FooSoft/yomichan/pull/516#discussion_r421844037
* Providing a more meaningful error if `setTimeout` fails: https://github.com/FooSoft/yomichan/issues/392#issuecomment-626187449
* It should handle disconnects/reconnects a little better: #392. This may become particularly useful for Manifest v3 where the service handler might not persist the forwarding port if it's not in use: #455.

PR #532, opened 2020-05-10
API refactor

toasted-nutbread:
Converts `api*` functions into a singleton class. This has a few benefits:

* Only need to do one `global api` instead of every single function.
* Will make some unit tests easier, since only `api` will need to be specified. (I've been putting off making some tests for this reason.)
* Removes global private functions.
* Function names now match the `action` string passed to `_invoke`. (Easier to search maybe?)
* State is now stored in a class instance, which should be more logical/consistent than global state variables (`_forwardLogsToBackendEnabled`).
* May be used to store #531's `CrossFrameAPI` instance when that's ready, which would be a stateful variable.

The [first commit](https://github.com/FooSoft/yomichan/commit/9d877b52e7c555ca96cc2616bf242c5af19cc3ba) may be easier to read in some parts if the "Hide whitespace changes" is enabled.

The [second commit](https://github.com/FooSoft/yomichan/commit/52c467cd4c1b02875e8b923b0aa9878365de9f1c) was automated using this script:
```js
const fs = require('fs');
const {getAllFiles} = require('./test/yomichan-test');

for (const f of getAllFiles('.', (fn) => /\.js$/.test(fn) && !/node_modules/.test(fn))) {
    let source = fs.readFileSync(f, {encoding: 'utf8'});
    source = source.replace(/(?<!\.)(?<!(?:async |const |let |))(?:window\.)?\bapi([A-Z])(\w+)\b(?!\.)/g, (g0, g1, g2) => `api.${g1.toLowerCase()}${g2}`);
    source = source.replace(/(?:window\.)api([A-Z])(\w+)\b(?!\.)/g, (g0, g1, g2) => `api.${g1.toLowerCase()}${g2}`);
    source = source.replace(/(\bapi\.\w+)(?!\w)(?!\()(?!\.bind)(?! = )/g, '$1.bind(api)');
    source = source.replace(/ \* api\.\w+(\r?\n)/g, ' * api$1');
    source = source.replace(/(\r?\n \* api){2,}/g, '$1');
    fs.writeFileSync(f, source);
}
```

### DIFF
--- a/ext/bg/js/search.js
+++ b/ext/bg/js/search.js
@@ -50,7 +50,7 @@ class DisplaySearch extends Display {
         this.introVisible = true;
         this.introAnimationTimer = null;
 
-        this.clipboardMonitor = new ClipboardMonitor({getClipboard: api.clipboardGet});
+        this.clipboardMonitor = new ClipboardMonitor({getClipboard: api.clipboardGet.bind(api)});
 
         this._onKeyDownIgnoreKeys = new Map([
             ['ANY_MOD', new Set([
--- a/ext/bg/js/search.js
+++ b/ext/bg/js/search.js
@@ -20,9 +20,7 @@
  * DOM
  * Display
  * QueryParser
- * apiClipboardGet
- * apiModifySettings
- * apiTermsFind
+ * api
  * wanakana
  */
 
@@ -52,7 +50,7 @@ class DisplaySearch extends Display {
         this.introVisible = true;
         this.introAnimationTimer = null;
 
-        this.clipboardMonitor = new ClipboardMonitor({getClipboard: apiClipboardGet});
+        this.clipboardMonitor = new ClipboardMonitor({getClipboard: api.clipboardGet});
 
         this._onKeyDownIgnoreKeys = new Map([
             ['ANY_MOD', new Set([
@@ -234,7 +232,7 @@ class DisplaySearch extends Display {
             this.setIntroVisible(!valid, animate);
             this.updateSearchButton();
             if (valid) {
-                const {definitions} = await apiTermsFind(query, details, this.getOptionsContext());
+                const {definitions} = await api.termsFind(query, details, this.getOptionsContext());
                 this.setContent('terms', {definitions, context: {
                     focus: false,
                     disableHistory: true,
@@ -258,7 +256,7 @@ class DisplaySearch extends Display {
         } else {
             wanakana.unbind(this.query);
         }
-        apiModifySettings([{
+        api.modifySettings([{
             action: 'set',
             path: 'general.enableWanakana',
             value,
@@ -274,7 +272,7 @@ class DisplaySearch extends Display {
                 (granted) => {
                     if (granted) {
                         this.clipboardMonitor.start();
-                        apiModifySettings([{
+                        api.modifySettings([{
                             action: 'set',
                             path: 'general.enableClipboardMonitor',
                             value: true,
@@ -288,7 +286,7 @@ class DisplaySearch extends Display {
             );
         } else {
             this.clipboardMonitor.stop();
-            apiModifySettings([{
+            api.modifySettings([{
                 action: 'set',
                 path: 'general.enableClipboardMonitor',
                 value: false,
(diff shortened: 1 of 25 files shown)


siikamiika:
This PR has some conflicts. Sorry that it's been in review for so long. I'll check again after they're resolved.

ext/bg/js/search.js
```diff
@@ -52,7 +50,7 @@ class DisplaySearch extends Display {
         this.introVisible = true;
         this.introAnimationTimer = null;
 
-        this.clipboardMonitor = new ClipboardMonitor({getClipboard: apiClipboardGet});
+        this.clipboardMonitor = new ClipboardMonitor({getClipboard: api.clipboardGet});
```
siikamiika, line 53: `api.clipboardGet.bind(api)`

ext/bg/js/settings/anki-templates.js
```diff
@@ -97,15 +94,15 @@ async function ankiTemplatesValidate(infoNode, field, mode, showSuccessResult, i
...
             const context = {
                 document: {
                     title: document.title
                 }
             };
             let templates = options.anki.fieldTemplates;
-            if (typeof templates !== 'string') { templates = await apiGetDefaultAnkiFieldTemplates(); }
-            const ankiNoteBuilder = new AnkiNoteBuilder({renderTemplate: apiTemplateRender});
+            if (typeof templates !== 'string') { templates = await api.getDefaultAnkiFieldTemplates(); }
+            const ankiNoteBuilder = new AnkiNoteBuilder({renderTemplate: api.templateRender});
```
siikamiika, line 105: `api.templateRender.bind(api)`

ext/bg/js/settings/popup-preview-frame.js
```diff
@@ -20,14 +20,13 @@
...
- * apiFrameInformationGet
- * apiOptionsGet
+ * api
  */
 
 class SettingsPopupPreview {
     constructor() {
         this.frontend = null;
-        this.apiOptionsGetOld = apiOptionsGet;
+        this.apiOptionsGetOld = api.optionsGet;
```
siikamiika, line 29: `api.optionsGet.bind(api)`

toasted-nutbread:
Resolved by updating the automation script:
```js
source = source.replace(/(\bapi\.\w+)(?!\w)(?!\()/g, '$1.bind(api)');
```
cd9b180128cd244f21cd79bb9e60892fa7398583

ext/mixed/js/api.js
```diff
@@ -15,307 +15,212 @@
...
+        getAnkiModelFieldNames(modelName) { return this._invoke('getAnkiModelFieldNames', {modelName}); }
+        getDictionaryInfo() { return this._invoke('getDictionaryInfo'); }
+        getDictionaryCounts(dictionaryNames, getTotal) { return this._invoke('getDictionaryCounts', {dictionaryNames, getTotal}); }
+        purgeDatabase() { return this._invoke('purgeDatabase'); }
+        getMedia(targets) { return this._invoke('getMedia', {targets}); }
+        log(error, level, context) { return this._invoke('log', {error, level, context}); }
+        logIndicatorClear() { return this._invoke('logIndicatorClear'); }
+        importDictionaryArchive(archiveContent, details, onProgress) { return this._invokeWithProgress('importDictionaryArchive', {archiveContent, details}, onProgress); }
+        deleteDictionary(dictionaryName, onProgress) { return this._invokeWithProgress('deleteDictionary', {dictionaryName}, onProgress); }
+        modifySettings(targets, source) { return this._invoke('modifySettings', {targets, source}); }
```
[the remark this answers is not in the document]
toasted-nutbread: I did it this way because:
* Using newlines (same layout as before) basically requires 4x the additional lines.
* Spaces between `)` and `{` would require changes to every line if a new function is added with long arguments, or if a function's arguments are changed.

I'll probably just revert back to the newline layout.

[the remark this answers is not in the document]
toasted-nutbread: 930cdac51fc30586981ecbd7b038473638d13928
f864a58f5e7fd82b2c70b8ca2846d9a567eea309

siikamiika:
> For some reason, popup preview `#example-text` gets selected on this branch but not on master, even with the suggested fixes.

Any idea why this happens?

toasted-nutbread:
Automation script missed `window.apiSomething` overrides. Updated script in main post.
7bac8cfb8ce95ed44413494aae430bb8ff5bdc22

Issue #538, opened 2020-05-11
Unable to add term's tags to Anki w/ {tags} marker


aiueo97:
I'm able to add a term's tags to an Anki card via the `{glossary}` marker, as shown here: 
![blah1](https://user-images.githubusercontent.com/65163733/81550735-f395d000-9380-11ea-9b90-96923f1c1acd.png)

But I want to add just the tags to another field in my Anki card, preferrably using the `{tags}` marker. When I try using the `{tags}` marker, the field ends up blank.

I've tried updating the template to use `definitionTags` instead of `definition.tags` as it is used in the `glossar`/`glossary-single` inline template, but the field still ends up blank.

Is this a bug, or am I using the template incorrectly?

For reference, here is a gist of my current template: https://gist.github.com/aiueo97/229189d8bf05159fc32ad010380d5138

I'm using JMdict.

siikamiika:
It seems to be an issue with the result grouping mode. When you set the mode to "split definitions to their own results", `{tags}` works. Otherwise it doesn't, presumably because it's not clear how to group the tags when they point to different definitions.

I think it's a bug because at least it should inform the user somehow that the marker is unsupported with the current result grouping mode. The implementation could also be changed to just merge all tags together and remove duplicates.

PR #539, opened 2020-05-11
Fix "tags" template

siikamiika:
Fixes https://github.com/FooSoft/yomichan/issues/538. Yet another issue caused by the data format issue of https://github.com/FooSoft/yomichan/issues/337.

### DIFF
--- a/ext/bg/js/backend.js
+++ b/ext/bg/js/backend.js
@@ -150,7 +150,7 @@ class Backend {
             await profileConditionsDescriptorPromise;
 
             this.optionsSchema = await requestJson(chrome.runtime.getURL('/bg/data/options-schema.json'), 'GET');
-            this.defaultAnkiFieldTemplates = await requestText(chrome.runtime.getURL('/bg/data/default-anki-field-templates.handlebars'), 'GET');
+            this.defaultAnkiFieldTemplates = (await requestText(chrome.runtime.getURL('/bg/data/default-anki-field-templates.handlebars'), 'GET')).trim();
             this.options = await optionsLoad();
             this.options = JsonSchema.getValidValueOrDefault(this.optionsSchema, this.options);
 
--- a/ext/bg/js/options.js
+++ b/ext/bg/js/options.js
@@ -111,19 +111,31 @@ const profileOptionsVersionUpdates = [
     },
     (options) => {
         // Version 14 changes:
-        //  Changed template for Anki audio.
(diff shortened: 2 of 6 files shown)


ext/bg/js/options.js
```diff
@@ -111,19 +111,31 @@ const profileOptionsVersionUpdates = [
...
-        fieldTemplates = fieldTemplates.replace(/\{\{#\*inline "audio"\}\}\{\{\/inline\}\}/g, () => {
-            replaced = true;
-            return replacement;
-        });
-
-        if (!replaced) {
-            fieldTemplates += '\n\n' + replacement;
+        const replacements = [
+            [
+                '{{#*inline "audio"}}\n    {{~#if definition.audioFileName~}}\n        [sound:{{definition.audioFileName}}]\n    {{~/if~}}\n{{/inline}}',
```
toasted-nutbread, line 120: 1. Flip the order of the `[replacement, pattern]` values, since it's kind of a `pattern => replacement` style.
2. Move `new RegExp(escapeRegExp(` and `), 'g')` inside the `for`.

siikamiika: https://github.com/FooSoft/yomichan/pull/539/commits/cf0f4a6a576f01d75b8bcae87e1c2cb90554985d

ext/bg/js/backend.js
```diff
@@ -150,7 +150,7 @@ class Backend {
             await profileConditionsDescriptorPromise;
 
             this.optionsSchema = await requestJson(chrome.runtime.getURL('/bg/data/options-schema.json'), 'GET');
-            this.defaultAnkiFieldTemplates = await requestText(chrome.runtime.getURL('/bg/data/default-anki-field-templates.handlebars'), 'GET');
+            this.defaultAnkiFieldTemplates = (await requestText(chrome.runtime.getURL('/bg/data/default-anki-field-templates.handlebars'), 'GET')).trim();
```
siikamiika, line 153: Is this something that could cause issues in the future? I think it should work fine as long as you don't compare anything with the raw result from the `.handlebars` file. I did it this way because I've configured my editor to ensure a newline at the end of the file and it was added when I modified the `tags` template.

toasted-nutbread: It should probably be fine.

toasted-nutbread:
Merging this on your behalf in your (semi-)absence.

PR #540, opened 2020-05-17
Add icons

toasted-nutbread:
Icons will be used for #465, among other things. Open for suggestion if anyone notices any issues, and they can always be changed if we notice any issues during implementation.

Commands used to generate the below image:
```bat
for %f in (*.svg) do @inkscape -z -w 128 -h 128 %f -e %~nf.png
magick montage -label "%t" -tile 6x *.png -geometry +2+2 all.png
```

128x128
icons128

32x32
icons32

16x16
icons16


siikamiika:
Nice :+1: I like the changes in (what I assume) are for appearance, scanning and keyboard shortcuts.

- Could there be trademark or licensing issues with the Anki icon?
- `text-parsing`: Maybe something like  (from http://icon-library.com/icon/icon-data-14.html) could work better, or at least I don't instantly get the idea of text parsing from the icon.
- Google Translate and Wikipedia (mobile) use an icon that contains 文 and A next to each other, maybe something with あ and A for translation options

siikamiika:
>* Could there be trademark or licensing issues with the Anki icon?

https://github.com/FortAwesome/Font-Awesome/blob/master/LICENSE.txt

> \# Brand Icons
All brand icons are trademarks of their respective owners. The use of these
trademarks does not indicate endorsement of the trademark holder by Font
Awesome, nor vice versa. **Please do not use brand logos for any purpose except
to represent the company, product, or service to which they refer.**

toasted-nutbread:
> Could there be trademark or licensing issues with the Anki icon?

This is a weird gray area that maybe it's best to try to avoid. On one hand, the icon I made is just composed of a rounded rect and two star cutouts. It wasn't traced, so the positions/sizes of everything are unlikely to be identical. On the other hand, I don't know where the line begins and ends for being "close enough".

The copyright/licensing for the logo is mentioned in anki's [README](https://github.com/ankitects/anki/blob/master/LICENSE).


> `text-parsing`: Maybe something like [[img](https://camo.githubusercontent.com/541f6475d04c68b1635cc87ca5621d0d0b589abe/68747470733a2f2f6c68352e676f6f676c6575736572636f6e74656e742e636f6d2f70726f78792f35486e68466a685666444d64325349356734426c6b5948446d42423149726c4e3079314c31333364485959494b797655337373757458736e7a53396f4f6d597661446c5a574c6445494a7341314c5f4d37567a4b5172594573445076384a4376424d55595351)] (from http://icon-library.com/icon/icon-data-14.html) could work better, or at least I don't instantly get the idea of text parsing from the icon.

True; there are a few things to bear in mind on that front:
1. The actual resolution of the icons will be 32x32, with the realistic work area likely being closer to 30x30.
2. The icons will be accompanied by a text label, at least when there is sufficient room and the text doesn't need to be collapsed.

So at the expected size, the icon you linked would look like this: 
image

I tried to keep it somewhat abstract and simple. It's often challenging to perfectly represent an uncommon, abstract process with a 32x32 icon. I can keep thinking on it.


> Google Translate and Wikipedia (mobile) use an icon that contains 文 and A next to each other, maybe something with あ and A for translation options

This is probably a good idea, assuming it looks fine for the resolution. I'll check that out some more.

toasted-nutbread:
Updated concepts: (left: old, right: new)
image

PR #541, opened 2020-05-17
Settings refactor

toasted-nutbread:
Cleans up a few things in preparation for some larger refactoring and changes.

* Debug info block is removed since the same data and more can be found by using the backup system.
* Updated some of the controls for options with special handling. This includes some of the Anki settings and `general.enableClipboardPopups` handling.
* This also fixes a bug where the Anki card fields would not be updated when changing the modifying profile.

### DIFF
--- a/ext/bg/js/settings/main.js
+++ b/ext/bg/js/settings/main.js
@@ -33,7 +33,6 @@
  * storageInfoInitialize
  * utilBackend
  * utilBackgroundIsolate
- * utilIsolate
  */
 
 function getOptionsMutable(optionsContext) {
--- a/ext/bg/js/settings/main.js
+++ b/ext/bg/js/settings/main.js
@@ -48,22 +48,6 @@ function getOptionsFullMutable() {
 
 async function formRead(options) {
     options.general.enable = $('#enable').prop('checked');
-    const enableClipboardPopups = $('#enable-clipboard-popups').prop('checked');
-    if (enableClipboardPopups) {
-        options.general.enableClipboardPopups = await new Promise((resolve, _reject) => {
-            chrome.permissions.request(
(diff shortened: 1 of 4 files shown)


ext/bg/js/settings/main.js
```diff
@@ -265,6 +250,9 @@ async function onOptionsUpdated({source}) {
 
     const optionsContext = getOptionsContext();
     const options = await getOptionsMutable(optionsContext);
+
+    onAnkiOptionsChanged();
```
siikamiika, line 261: `await`

ext/bg/js/settings/main.js
```diff
@@ -260,6 +259,7 @@ async function onOptionsUpdated({source}) {
     const options = await getOptionsMutable(optionsContext);
 
     document.querySelector('#enable-clipboard-popups').checked = options.general.enableClipboardPopups;
+    ankiTemplatesUpdateValue();
```
siikamiika, line 259: Does this not `await` on purpose?

toasted-nutbread: I believe all the awaits are omitted because they are unnecessary. The anki settings are modified independently, so there is no reason to await.

toasted-nutbread:
> There's still a related issue with Anki template preview that I noticed when doing #539. If you change result grouping mode, the object used to generate the preview isn't updated before you reload the settings page or edit the preview text. Maybe not in scope for this PR unless it reveals other issues that should be fixed.

In general, most of the files in the settings folder need to be refactored. I am trying to do incremental changes to get to that point. The layout/design changes for the updated settings page also necessitate some changes. So tasks like that will probably be addressed during those changes.

PR #551, opened 2020-05-23
Settings backup refactor

toasted-nutbread:
Converts backup.js into a class. Moves `utilReadFileArrayBuffer` out of `util.js` and into this new class.

### DIFF
--- a/ext/bg/js/settings/backup.js
+++ b/ext/bg/js/settings/backup.js
@@ -22,7 +22,6 @@
  * utilBackend
  * utilBackgroundIsolate
  * utilIsolate
- * utilReadFileArrayBuffer
  */
 
 class SettingsBackup {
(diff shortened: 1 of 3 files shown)


ext/bg/js/settings/backup.js
```diff
@@ -24,355 +24,366 @@
...
+        a.rel = 'noopener';
+        a.target = '_blank';
 
-    const date = new Date(Date.now());
+        const revoke = () => {
+            URL.revokeObjectURL(blobUrl);
+            a.href = '';
+            this._settingsExportRevoke = null;
+        };
+        this._settingsExportRevoke = revoke;
```
siikamiika, line 108: I think this should be fine without `.bind`, but could double check

toasted-nutbread: Should be fine since it's an arrow function & not on the prototype.

PR #553, opened 2020-05-23
Use cross frame API

toasted-nutbread:
Follow-up of #531. Replaces the old system with the new one. Will want to rebase this PR after #532 is merged in order to make `crossFrame` a member of the `api` object.


toasted-nutbread:
> Will want to rebase this PR after #532 is merged in order to make `crossFrame` a member of the `api` object.

66932515cf1cf66f65a3bfa6492c35147af8d725

PR #555, opened 2020-05-23
Immediate backend event handlers

toasted-nutbread:
Addresses https://github.com/FooSoft/yomichan/issues/455#issuecomment-633134977

### DIFF
--- a/ext/bg/js/backend.js
+++ b/ext/bg/js/backend.js
@@ -181,15 +181,6 @@ class Backend {
 
             this.onOptionsUpdated('background');
 
-            if (isObject(chrome.commands) && isObject(chrome.commands.onCommand)) {
-                chrome.commands.onCommand.addListener(this._runCommand.bind(this));
-            }
-            if (isObject(chrome.tabs) && isObject(chrome.tabs.onZoomChange)) {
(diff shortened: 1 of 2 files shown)


ext/bg/js/backend.js
```diff
@@ -137,7 +143,26 @@ class Backend {
...
+                },
+                (error) => {
+                    this._prepareError = true;
+                    this._prepareCompleteReject(error);
+                }
+            );
+            promise.finally(() => this._updateBadge());
+            this._preparePromise = promise;
+        }
+        return this._prepareCompletePromise;
```
siikamiika, line 162: It appears that this is not used (`background-main.js` doesn't await); is this intentional?

toasted-nutbread: The intent is to have it match a call to `.prepareComplete()` in terms of what is returned. So it is intentional, but current functionality doesn't use it. The reason `await` was removed from background-main.js is because there was nothing after it, and having the entry point function be non-`async` makes the semantics more clear. I.e. for anyone who isn't perfectly familiar with how/when promises invoke, it will be clear that there are no deferrals before `BackendEventHandler` is prepared.

siikamiika: Ok, in that case the PR looks good to me.

Issue #568, opened 2020-05-25
Anki decks/models not updated when toggling "Enable Anki integration"


toasted-nutbread:
Noticed while working on #567.

PR #573, opened 2020-05-25
Generic settings controller + clipboard popups controller

toasted-nutbread:
`GenericSettingsController` will be superseded by #542 in the future.


toasted-nutbread:
I believe this is the last settings controller class refactor. After this will come some updates to use #565.

So long as there are no objections, I will probably self-test/review and merge these changes in. The changes made in the recent PRs put up don't drastically change the implementations, but the diffs are a mess due to whitespace, naming, and ordering changes. It will probably be less headache and faster to test for issues after they are all merged in together. There will still be more changes coming in the future as well (e.g. #568) which can be reviewed more thoroughly.

PR #597, opened 2020-06-10
Handle cases where platform info is not available

### LINKED ISSUE #595: Yomichan Chrome extension isn't working with the android Kiwi Browser since version 20.5.22.0
Not sure if this is the best place to post this, but I'll try.

I've been using the Chromium based Kiwi Browser for the sole reason that Yomichan works excellent with it. But since 20.5.22.0 and 20.5.22.1 it stopped working. It turns on in the extension manager, but when I try to turn it on via the Yomichan On/Off button it always goes back to Off. If I go into the setting page the Active Profile box is empty, along with almost everything else, and no buttons work.
Version 20.4.10.1 and back still works fine, but since the extension automatically updates I can't use it for very long before it updates to the newest version. I've tried the latest update on 2 phones, but both were Samsung S10e.

That is all I know about this issue, I'm not 100% sure if it's not a Yomichan issue, but I ran out of ideas to try and get it fixed so now I wrote here.

toasted-nutbread:
Fixes #595. `chrome.runtime.getPlatformInfo` was generating an error and returning `undefined`.

### DIFF
--- a/ext/mixed/js/environment.js
+++ b/ext/mixed/js/environment.js
@@ -32,15 +32,34 @@ class Environment {
 
     async _loadEnvironmentInfo() {
         const browser = await this._getBrowser();
-        const platform = await new Promise((resolve) => chrome.runtime.getPlatformInfo(resolve));
-        const modifierInfo = this._getModifierInfo(browser, platform.os);
+        const os = await this._getOperatingSystem();
+        const modifierInfo = this._getModifierInfo(browser, os);
         return {
             browser,
-            platform: {
-                os: platform.os
-            },
+            platform: {os},
             modifiers: modifierInfo
         };
+    }
+
+    async _getOperatingSystem() {
+        try {
+            return await this._getPlatformInfo();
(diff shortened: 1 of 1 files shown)


ext/mixed/js/environment.js
```diff
@@ -32,17 +32,36 @@ class Environment {
...
-                os: platform.os
-            },
+            platform: {os},
             modifiers: modifierInfo
         };
     }
 
+    async _getOperatingSystem() {
+        try {
+            return await this._getPlatformInfo();
```
siikamiika, line 46: This is an object, but we only need the `os` property if it exists

toasted-nutbread: 9fad3029d093d9bef210b1436c985e9f9be32ebc

toasted-nutbread:
Agreed; it's worth a refactor at some point. (Unfortunately manifest v3's stateless backend will come eventually and ruin everything, but such is life.)

PR #604, opened 2020-06-13
Ignore getZoom errors

toasted-nutbread:
Unclear how this happened, but I saw the following error occur on Firefox:
```
Yomichan v20.5.22.1 has encountered a problem.
Originating URL: about:blank
Error: Invalid tab
_onApiGetZoom@moz-extension://.../bg/js/backend.js:687:35
onMessage@moz-extension://.../bg/js/backend.js:222:44
```

This change should fix it.

### DIFF
--- a/ext/fg/js/frontend.js
+++ b/ext/fg/js/frontend.js
@@ -65,8 +65,12 @@ class Frontend {
     async prepare() {
         try {
             await this.updateOptions();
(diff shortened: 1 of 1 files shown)


toasted-nutbread:
Also noticed this happen on a non-about:blank website, so maybe it occurred due to an invalid state.

PR #605, opened 2020-06-13
Set up initial manifest v3 support

toasted-nutbread:
This PR is not intended to be merged, it is for testing and evaluation purposes only.

The service work can be inspected in Chrome by opening `chrome://serviceworker-internals/` and inspecting the extension's service worker.

The manifest is not enabled by default, it is configured as a new build target: `chrome-mv3`. To update the manifest, run `build --manifest chrome-mv3`.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -156,6 +156,46 @@
                 "node": true,
                 "webextensions": false
             }
+        },
+        {
+            "files": [
+                "ext/mixed/js/core.js",
+                "ext/mixed/js/yomichan.js",
+                "ext/mixed/js/environment.js",
+                "ext/mixed/js/japanese.js",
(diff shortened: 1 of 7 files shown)


toasted-nutbread:
Note: On the current version of Chrome Dev (87.0.4252.0), there is a bug which can prevent the browser from starting when using this branch. To reproduce:

* Switch to this branch
* Reload the extension
* Close the browser
* Switch branches to a non-manifest-v3 branch
* Open the browser
* Browser will crash immediately

This appears to be due to an issue with the file `C:\Users\<user>\AppData\Local\Google\Chrome Dev\User Data\Default\Secure Preferences`. This crash can be resolved by switching back to the manifest-v3 branch, launching the browser, changing to a non-manifest-v3 branch, and then reloading the extension.

PR #612, opened 2020-06-17
Handlebars sandbox

toasted-nutbread:
Follow-up of #574. Implements a sandbox frame for rendering templates. The frame is currently created on the background page, but as with most things, this won't work with manifest v3. This is another intermediate step.

Another point is that apparently the sandboxing feature is not supported in Firefox WebExtensions. This means that in order to fully secure the default `content_security_policy`, there would need to be two versions of the extension: one for Firefox (which keeps the `unsafe-eval` permission on the default `content_security_policy`) and one for Chrome (which uses `sandbox.content_security_policy`).

https://developer.chrome.com/apps/sandboxingEval

* [x] Buttons don't appear grayed out when Anki isn't running.
* [x] Fields filled out incorrectly. To reproduce: `Front: {expression}; Back: {screenshot}`.

### DIFF
--- a/dev/data/manifest-variants.json
+++ b/dev/data/manifest-variants.json
@@ -67,6 +67,12 @@
             "page": "bg/settings.html",
             "open_in_tab": true
         },
+        "sandbox": {
+            "pages": [
+                "bg/template-renderer.html"
+            ],
+            "content_security_policy": "sandbox allow-scripts; script-src 'self' 'unsafe-eval'; object-src 'self'"
+        },
         "permissions": [
             "<all_urls>",
             "storage",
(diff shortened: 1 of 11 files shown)


toasted-nutbread:
One side effect of this change is that the handlebars sandbox can be used by non-Yomichan webpages since it is now a `web_accessible_resources`. However, this shouldn't have any negative side effects, since the webpage only responds to an input message and replies with an output message.

This may be able to be addressed by setting up a service worker to intercept and modify requests to the URL to have a secret, but there was difficulty and inconsistency setting that up. May be worth revisiting later.

toasted-nutbread:
Technically it would also be doable by doing something similar to this:
```js
const baseUrl = chrome.runtime.getURL('/bg/search.html');
const templateRendererUrl = chrome.runtime.getURL('/bg/template-renderer.html');
const callback = ({url}) => {
    if (url === templateRendererUrl) {
        const redirectUrl = `${url}?some-secret`;
        return {redirectUrl};
    }
    return {};
};
const filter = {urls: [baseUrl]};
chrome.webRequest.onBeforeRequest.addListener(callback, filter, ['blocking']);
```
But that would add a lot of complexity. I will leave that as a task for later, if necessary.

PR #668, opened 2020-07-18
Check lastError in _updateSearchQuery

toasted-nutbread:
Handles cases where the tab is invalid.


toasted-nutbread:
Oops, lastError is already checked in `getMessageResponseResult`.

PR #797, opened 2020-09-10
Use Anki classes directly in Display

toasted-nutbread:
This change moves the usage of `AnkiConnect` and `AnkiNoteBuilder` into `Display` (and the respective settings controllers). This change is necessary to create the sandboxed iframe with manifest v3 support, since the background page (service worker) will no longer support DOM.

### DIFF
--- a/ext/bg/background.html
+++ b/ext/bg/background.html
@@ -23,8 +23,6 @@
         <script src="/mixed/js/environment.js"></script>
         <script src="/mixed/js/japanese.js"></script>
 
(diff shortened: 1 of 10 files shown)


toasted-nutbread:
This implementation causes "issues" due to the `AnkiConnect` instance being used directly from the popup: unpreventable error messages show up in the webpage console due to no Anki connection. This is undesirable since it adds clutter and is an expected use case, but the error cannot be suppressed.

toasted-nutbread:
Superseded by #804.

PR #943, opened 2020-10-19
Custom audio URL random

### LINKED ISSUE #898: [Feature Request] Wildcard with custom audio URL
It seems like oftentimes a custom URL source will include the expression and/or reading of a word in its path, but will also include a random string or identifier of some sort. To be able to use these sources, having a wildcard like an asterisk, or being able to specify a regex would be convenient. Does yomichan already something like this?

toasted-nutbread:
Adds support for generating random strings as part of custom audio URLs. The format is:

`{random:CHARACTERS,COUNT}`

* `CHARACTERS` - a sequence of characters or character ranges.
  * A single character is just that character, e.g. `X`.
  * Characters ranges are defined in the format `X-Z`, which includes the characters `X`, `Z`, and all characters between.
  * There is no delimiter between character/range groups.
  * Characters can optionally be prefixed with a `\`, which can be used to include a raw `-` character without forming a group.
* `COUNT` - non-negative base-10 integer.

Example: `{random:a-zA-Z0-9\-,32}` to generate a 32 character string including alphanumeric characters and a dash.

Resolves #898?

### DIFF
--- a/ext/bg/js/audio-downloader.js
+++ b/ext/bg/js/audio-downloader.js
@@ -31,6 +31,7 @@ class AudioDownloader {
             ['text-to-speech-reading', this._getInfoTextToSpeechReading.bind(this)],
             ['custom', this._getInfoCustom.bind(this)]
         ]);
+        this._customUrlFormatPattern = /\{([^}]*)\}/g;
     }
 
     async getInfo(source, expression, reading, details) {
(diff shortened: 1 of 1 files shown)


toasted-nutbread:
Closing as it does not resolve the original issue and does not seem to be useful otherwise.

PR #1199, opened 2021-01-03
Translator regex replacements

### LINKED ISSUE #1195: [Feature Request] ignore parentheses
Hello, i had an idea while using netflix. Often a website or netflix subtitles use parentheses instead of furigana :  e.g " なんかあったら 遠慮(えんりょ)なく言(い)いなさい "
Having an option that ignore the parentheses so we can search 言いなさい  will be useful because currently yomichan will block at 言 and willn't make the link with いんさい. 
Thank for developing this extension that I use every day ! I hope my explanation is understandable 😄

toasted-nutbread:
Resolves #1195.

Example pattern: `[(]([^)]*)[)]`

### DIFF
--- a/ext/bg/css/settings2.css
+++ b/ext/bg/css/settings2.css
@@ -1744,7 +1744,8 @@ code.anki-field-marker {
     grid-template-rows: auto;
     grid-template-areas:
         "index pattern-label     pattern     button"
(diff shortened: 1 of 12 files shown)


Thermospore:
This looks interesting! About time I learned regex... looking forward to messing around with this

PR #1307, opened 2021-01-24
Update Basic Usage section of the readme

toasted-nutbread:
This change updates the README.md file to have more up-to-date descriptions and images for the _Basic Usage_ section. Image URLs are currently local URLs, and files are in the project's `resources` folder.

This change does not include updates to the rest of the readme, or updates to the screenshots at the end; those will be updated in a different change.

@FooSoft Requesting a review from you, since this may affect how the images need to be handled on foosoft.net.

[Preview](https://github.com/FooSoft/yomichan/blob/58fd5641b4907ca3c944a6bf6675642c3f0d12b7/README.md#basic-usage)


FooSoft:
Very nice! Not that it matters that much, but is there any reason that you're using HTML instead of markdown in the MD file? Does the markdown renderer screw it up or something? 

Regarding images, I think it's fine to have them be in the repo itself. The only reason they are not is that I was initially using a script to convert my website to the `README.md` file automatically (so I wouldn't have to update documentation in both places), but there is no real reason to continue doing this as I have a different process now.

README.md
```diff
@@ -84,32 +84,35 @@ primary language is not English, you may consider also importing the English ver
 
 ## Basic Usage ##
 
-1.  Click on the ![](https://foosoft.net/projects/yomichan/img/yomichan-icon.png) icon in the browser toolbar to open the Yomichan actions dialog.
+1.  Click the <img src="ext/mixed/img/yomichan-icon.svg" alt="" width="16" height="16"> _Yomichan_ button in the browser bar to open the quick-actions popup.
```
FooSoft, line 87: Why raw HTML?

toasted-nutbread: Forcing the width/height to 16, since they are SVGs. Should prevent any layout issues that could occur if these sizes are ever changed. It should also make their load placeholder have a constant size and not change the layout after the content is loaded.

FooSoft:
@toasted-nutbread site is now in sync with the repo

PR #1310, opened 2021-01-25
Enable automatic updates for the firefox-dev release

### LINKED ISSUE #830: Add support for auto-updating of Firefox testing builds
https://stackoverflow.com/questions/40541561/firefox-web-extension-auto-update-unlisted-and-self-hosted-extension

toasted-nutbread:
@FooSoft Let me know if you think we should add this field to the standard `firefox` build target. I assume it won't affect the store deployment since the [Firefox information page](https://extensionworkshop.com/documentation/manage/updating-your-extension/) states:

> Firefox supports automated updates to add-ons using JSON update manifests. Add-ons hosted on addons.mozilla.org (AMO) automatically receive updates to new versions posted there. Other add-ons must specify the location of their update manifests.

Resolves #830.

### DIFF
--- a/dev/data/manifest-variants.json
+++ b/dev/data/manifest-variants.json
@@ -235,6 +235,11 @@
                     "action": "set",
                     "path": ["browser_specific_settings", "gecko", "id"],
                     "value": "alex.testing@foosoft.net"
+                },
+                {
+                    "action": "set",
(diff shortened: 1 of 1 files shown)


FooSoft (approved):
Let's go with this, as it's the most straightforward approach. If there are any conflicts with the store, we can address them then.

PR #1343, opened 2021-02-04
え deinflection

toasted-nutbread:
This list is probably not comprehensive, but it covers a lot of common cases.

https://japanese.stackexchange.com/questions/29099/changing-of-diphthongs-at-the-end-of-words-to-%E3%81%88-in-exclamations


byjtje:
https://github.com/FooSoft/yomichan/blob/9520d5d677661e495307f0bf05b10e02770ef234/ext/bg/lang/deinflect.json#L598-L599
For what it's worth those are also seen just as ええ, like こええ or つええ

PR #1524, opened 2021-03-13
Remove old settings page

toasted-nutbread:
This has been in the works for a long time, and the new settings page is finally ready to completely replace the old one. This removes the dependency on jQuery and Bootstrap.

FYI @FooSoft, mostly just as a noteworthy milestone.


FooSoft:
Very nice! It's awesome that we are finally dropping these heavy dependencies (and that you actually know your CSS well enough to allow this to happen) 👍🏻

PR #1575, opened 2021-03-31
Definition collapsing when overflowing

### LINKED ISSUE #1051: Collapsible Dictionaries+Words
It would be super nice to have them. Some dictionary entries from Kenkyusha take a long time to scroll past. One might want to have collapsed J-E dictionaries by default to reveal only when the J-J definition is unclear or has too many new words.

Collapsible word entries would be cool too. First result not what you wanted? Click a button to hide instead of scrolling through several dictionary entries of a different word.

Being able to collapse right in the middle of a long entry seems important too. You may not realize how long the entry is until midway.

toasted-nutbread:
Resolves #1051.

### DIFF
--- a/ext/css/display.css
+++ b/ext/css/display.css
@@ -102,6 +102,7 @@
     --animation-duration2: calc(2 * var(--animation-duration));
 
     --collapsible-definition-line-count: 3;
+    --collapsible-definition-test-offset: 0.2em;
 
     /* Colors */
     --background-color: #ffffff;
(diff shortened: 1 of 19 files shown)


oakkitten:
Probably I'm doing something wrong, but I find that sometimes the collapse test fails. For instance, if I reset all the settings, then turn on only `大辞林`, and then `Compact glossaries` and collapsing for all dictionaries, I get this:

image

Also, when collapse test does work, adding either of these CSS rules seems to break it

```css
.tag[data-category=dictionary] {
  display: none
}

:root {
  --line-height-no-units: 19;
}
```

toasted-nutbread:
@oakkitten Please create a new issue for visibility; I tried reproducing the issue following your steps to no avail.

Issue #1580, opened 2021-04-01
March 2021 Release


toasted-nutbread:
https://github.com/FooSoft/yomichan/releases/tag/21.3.31.0

@FooSoft 

### General
* [x] Create [release tag](https://github.com/FooSoft/yomichan/tree/21.3.31.0).
* [x] Create [GitHub release](https://github.com/FooSoft/yomichan/releases/tag/21.3.31.0).

### Testing
* [ ] Update Chrome testing release.
* [ ] Update Firefox testing release.
* [ ] Update Firefox [updates.json](https://github.com/FooSoft/foosoft.github.io/blob/master/projects/yomichan/dl/updates.json) and [verify](https://foosoft.net/projects/yomichan/dl/updates.json).
* [ ] Upload testing builds to the [GitHub release](https://github.com/FooSoft/yomichan/releases/tag/21.3.31.0).

### Stable
* [ ] Update Chrome stable release.
* [ ] Update Firefox stable release.
* [ ] Upload stable builds to the [GitHub release](https://github.com/FooSoft/yomichan/releases/tag/21.3.31.0).

FooSoft:
Testing release up!

Thermospore:
I'd like to try using the new option hotkey feature to toggle a dictionary. Is there an easy way to find the path for a given option? Might be nice to add a little explanation text in the menu there

Also does that only work for these kinds of options?
![image](https://user-images.githubusercontent.com/35350819/113380952-209cf980-932a-11eb-875a-2a77380a0bc0.png)
or can it cycle through dropdown lists as well

toasted-nutbread:
Check [options-schema.json](https://github.com/FooSoft/yomichan/blob/master/ext/data/schemas/options-schema.json) or export your settings file and look at the paths. Paths start relative to profile[current].options, and the box will outline in red if the path is invalid. It only works on booleans.

grospoulpe:
Just installed, with all the dictionaries I use (or not ;p), keeping the stable version (I suppose we have to, and eventually will become that new version)
So far, so good - and be able to play different audio source with a keyboard shortcut is great / timesaving :)
And the possibility to collapse dictionaries (especially with 研究社　新和英大辞典　第５版 , which has a lot of example displayed) is a useful feature, too :)

Aridow:
I noticed it's no longer possible to drag and drop kanjis from kanjidics dictionaries on Firefox. Is it expected?
Here's what I mean on Chrome : 
[...]

PR #1616, opened 2021-04-18
Add missing sequence and sequenceDictionary Info for termGrouped

qharscoet:
Added sequence info for groupedSequence, as #1591 added it in merged but not Grouped.

Also updated test data


toasted-nutbread:
I think the reason why the fields were empty for grouped mode is because the sequence is not used for merging, and specifying a single `sequence`/`sequenceDictionary` is misleading, since the entries may be coming from different dictionaries and/or have multiple `sequence` IDs.

Based on the changes here, only the `sequence`/`sequenceDictionary` that is encountered first would be exposed.

I assume this is to address more compatibility issues mentioned in https://github.com/FooSoft/yomichan/issues/1580#issuecomment-812904279?

qharscoet:
Hmm, in that case couldn't we still expose the sequence in the `defintions` array and keep -1 in the root ?

Yes, grouped mode is the mode I use myself to make my anki cards and I'd like to keep this functionality as it was present prior to the March update 😅

Maybe I'll switch to merged mode then, but I'd like to keep only the reading I encounter instead of having them all in the same card.

toasted-nutbread:
Yeah, adding the sequence to the definitions makes sense. The reason I didn't originally do that was because I didn't think anyone used that for Anki data, so I omitted it.

Check out #1617, that PR should fix the issues. I may also consider a follow-up PR which removes the `sequence` field from the internal data format and derive it for the Anki template renderer data format, since it's slightly redundant.

And since you're the first person in a while to open a PR (besides myself), let me know if you noticed anything in the codebase that would have made it easier to understand or figure out where/how to make changes; I know a lot of stuff isn't really documented. Appreciate the contribution

qharscoet:
> The reason I didn't originally do that was because I didn't think anyone used that for Anki data, so I omitted it.

Yeah it's just so that I can use a link in my cards to pull up more info about the words and the kanjis used in the Kanji Study app while reviewing on Android. Kind of specific but maybe someone else could use this info.

> And since you're the first person in a while to open a PR (besides myself), let me know if you noticed anything in the codebase that would have made it easier to understand or figure out where/how to make changes; I know a lot of stuff isn't really documented. Appreciate the contribution

I mostly looked at what you edited to know where to look at, so I can't really speak about the general organization of the code.
First thing that comes to mind from what I saw is in the name of the variables, things like `dictionaryEntries2` in `_groupDictionaryEntriesByHeadword` that is named `groups` in `_getRelatedDictionaryEntries` , stuff like that.

```javascript  
const newDictionaryEntries = [];
for (const group of groupedDictionaryEntries) {
    newDictionaryEntries.push(this._createGroupedDictionaryEntry(group.dictionaryEntries, group.sequence, group.sequenceDictionary, true));
}
```

```javascript
const results = [];
for (const dictionaryEntries2 of groups.values()) {
    const dictionaryEntry = this._createGroupedDictionaryEntry(dictionaryEntries2.entries, dictionaryEntries2.sequence, dictionaryEntries2.sequenceDictionary, false);
    results.push(dictionaryEntry);
}
```

toasted-nutbread:
Thanks for the feedback, I'll make a PR to improve the consistency of those two cases also.

Issue #1626, opened 2021-04-26
[Feature Request] Show anki tags and/or flags on words that already have cards created


soriac:
When reviewing cards I often notice that the context I added is bad, insufficient, or maybe I even forgot to add the actual context. I like to tag and flag these words so that whenever I want to clean my decks up I can quickly find "bad cards" and deal with them.

Sometimes, when reading japanese text I'll see one of the words I have flagged and remember that it's flagged, so I can delete the old card or update its context. However most of the time I simply come across these words, see that they already have a card and then not fix it because I don't know that the card was bad.

It would be nice to be able to show all tags and/or flags for any words that have an anki card associated with them, so that I can quickly see "oh, this is a red card tagged with `insufficient_context`, let's fix it" and not have to be on a constant lookout while reading.

Thermospore:
Maybe not what you are looking for but personally I would
 -[make cards out of sentences instead of words](https://refold.la/roadmap/stage-2/a/basic-sentence-mining#Mine-Sentences-Not-Words)
 -add a screenshot or photo of whatever you were consuming for further context
 
This should prevent the issue from the root
 
Here are what my sentence cards look like for example

front:
![image](https://user-images.githubusercontent.com/35350819/116037097-17930580-a614-11eb-9be1-81f209b20003.png)

back (usually I only learn one word per card but these were easy):
![image](https://user-images.githubusercontent.com/35350819/116037392-7eb0ba00-a614-11eb-93b9-3b0a97b93a0e.png)

soriac:
Yes! I actually make Audio Cards, with sentences, pictures and audio, mining from visual novels. 

The problem is sometimes, in the future, I realize the card I made wasn't that good, or I misunderstood the word or context. For example, I created a card for 融通 where the sentence actually uses 融通が利く.

Sometimes while reading I'll be too focused and just hit the plus button but then forget to add the rest of the context.

Sometimes there's a word I really want to add, because it comes up over and over and I need some help remembering the reading, but I just can't find a good context for within the novel itself.

And finally, sometimes the context is NSFW and while I'm fine with that, I'd rather keep my decks as SFW as possible.

In those cases, I'd rather not have perfect be the enemy of good and just make a "meh" card, that I can maybe fix later. Right now I just tag and flag those cards, and whenever I have some free time try to either delete or unflag all flagged cards (by finding some other context for them).

I thought of moving those cards to a separate deck, but then the new card order gets messed up and they either always go to the very end or the very start of the queue.

[...]

PR #1628, opened 2021-04-26
Show any custom tags on words that have anki cards created

soriac:
This PR adds a small button next to the "add card/view added card" button on the display template. This button will only appear if:

1. The word has an anki card.
2. That card has tags.
3. At least one of those tags is not in the "Card tags" anki setting.

This button does nothing when clicked, but instead when hovered over will show a tooltip with all of the custom tags that belong to this card:

ImageGlass_rnfeHloBC6

ImageGlass_Ph0TTHqidr

This PR would solve issue #1626. This feature would help my workflow, as I tend to add cards that I want to later edit or fix if/when I find a better context for them. I go into more details of my workflow in the issue.

There are a few things that still need to be considered:

- I'm not sure about code style/conventions/the way I'm handling data in `backend.js`. Particularly how I'm accessing user config, using anki-connect to get card tags, and also assuming that I will only ever get one ID for a particular note when calling `findNoteIds`. I go into detail with comments in lines 488-505 in `backend.js`.
- A different icon should probably be used. I quickly picked one at random from the icon folder. I'm not sure if there are any design guidelines or any other factor to decide on icons, but the current magnifying glass is a bit hard to see against a dark background.
- We could consider adding an option to enable/disable this behaviour, especially if fetching the required data from anki-connect turns out to be expensive.

### DIFF
--- a/ext/css/display.css
+++ b/ext/css/display.css
@@ -691,6 +691,9 @@ button.action-button[data-icon=play-audio]::before {
 button.action-button[data-icon=source-term]::before {
     background-image: url('/images/source-term.svg');
 }
+button.action-button[data-icon=magnifying-glass]::before {
+    background-image: url('/images/magnifying-glass.svg');
+}
 :root[data-result-output-mode=merge] .entry[data-type=term] .actions>button.action-button.action-play-audio {
     display: none;
 }
--- a/ext/css/display.css
+++ b/ext/css/display.css
@@ -644,7 +644,7 @@ button.action-button[hidden] {
         -webkit-filter var(--animation-duration) linear,
         background-color var(--animation-duration) linear;
 }
-button.collapsable-action-button[hidden] {
+button.collapsible-action-button[hidden] {
     display: none;
 }
 button.action-button:disabled {
--- a/ext/css/display.css
+++ b/ext/css/display.css
@@ -667,7 +667,7 @@ button.action-button:active {
     background-color: var(--action-button-active-color);
     box-shadow: none;
 }
-button.action-button::before {
+button.action-button[data-icon]::before {
     content: '';
     width: var(--action-button-size);
     height: var(--action-button-size);
@@ -693,6 +693,12 @@ button.action-button[data-icon=source-term]::before {
 }
 button.action-button[data-icon=magnifying-glass]::before {
     background-image: url('/images/magnifying-glass.svg');
+}
+.action-view-tags>.icon {
+    display: block;
+    width: var(--action-button-size);
+    height: var(--action-button-size);
+    background-color: var(--text-color);
 }
 :root[data-result-output-mode=merge] .entry[data-type=term] .actions>button.action-button.action-play-audio {
     display: none;
--- a/ext/css/display.css
+++ b/ext/css/display.css
@@ -691,9 +691,6 @@ button.action-button[data-icon=play-audio]::before {
 button.action-button[data-icon=source-term]::before {
     background-image: url('/images/source-term.svg');
 }
-button.action-button[data-icon=magnifying-glass]::before {
-    background-image: url('/images/magnifying-glass.svg');
-}
 .action-view-tags>.icon {
     display: block;
     width: var(--action-button-size);
--- a/ext/css/display.css
+++ b/ext/css/display.css
@@ -644,6 +644,9 @@ button.action-button[hidden] {
         -webkit-filter var(--animation-duration) linear,
         background-color var(--animation-duration) linear;
 }
+button.collapsable-action-button[hidden] {
+    display: none;
+}
 button.action-button:disabled {
     pointer-events: none;
     cursor: default;
--- a/ext/js/background/backend.js
+++ b/ext/js/background/backend.js
@@ -480,7 +480,6 @@ class Backend {
             const noteIdsArray = await this._anki.findNoteIds(cannotAddNotes);
             for (let i = 0, ii = Math.min(cannotAdd.length, noteIdsArray.length); i < ii; ++i) {
                 const noteIds = noteIdsArray[i];
-
                 if (noteIds.length > 0) {
                     cannotAdd[i].info.noteIds = noteIds;
                     cannotAdd[i].info.noteInfos = await this._anki.notesInfo(noteIds);
--- a/ext/js/background/backend.js
+++ b/ext/js/background/backend.js
@@ -469,7 +469,6 @@ class Backend {
             const valid = AnkiUtil.isNoteDataValid(note);
             if (!valid) { canAdd = false; }
             const info = {canAdd, valid, noteIds: null};
-
             results.push(info);
             if (!canAdd && valid) {
                 cannotAdd.push({note, info});
@@ -484,29 +483,7 @@ class Backend {
 
                 if (noteIds.length > 0) {
                     cannotAdd[i].info.noteIds = noteIds;
-
-                    /* TODO: should i be selecting the first ID like this?
-                        looking at findNoteIds I can't see how we could find
-                        more than one ID, but an example that proves me wrong
-                        would be cool (and also how to get the right ID) */
-                    const noteId = noteIds[0];
-
-                    /* TODO: should I be catching any errors here? findNoteIds
-                        is not wrapped in try/catch, but I can't find a catch
-                        anywhere in the code that calls this function */
-                    const noteInfo = (await this._anki.notesInfo([noteId]))[0];
-
-                    /* TODO: is this the right way to get user settings?
-                        I couldn't find any examples in the code */
-                    const settingsTags = this._getSetting({
-                        path: 'anki.tags',
-                        scope: 'profile',
-                        optionsContext: {current: true}
-                    });
-
-                    if (settingsTags && noteInfo) {
-                        cannotAdd[i].info.tags = noteInfo.tags.filter((tag) => !settingsTags.includes(tag));
-                    }
+                    cannotAdd[i].info.noteInfos = await this._anki.notesInfo(noteIds);
                 }
             }
         }
--- a/ext/js/background/backend.js
+++ b/ext/js/background/backend.js
@@ -458,7 +458,7 @@ class Backend {
         return await this._anki.addNote(note);
     }
 
-    async _onApiGetAnkiNoteInfo({notes}) {
+    async _onApiGetAnkiNoteInfo({notes, fetchAdditionalInfo}) {
         const results = [];
         const cannotAdd = [];
         const canAddArray = await this._anki.canAddNotes(notes);
@@ -482,7 +482,9 @@ class Backend {
                 const noteIds = noteIdsArray[i];
                 if (noteIds.length > 0) {
                     cannotAdd[i].info.noteIds = noteIds;
-                    cannotAdd[i].info.noteInfos = await this._anki.notesInfo(noteIds);
+                    if (fetchAdditionalInfo) {
+                        cannotAdd[i].info.noteInfos = await this._anki.notesInfo(noteIds);
+                    }
                 }
             }
         }
--- a/ext/js/background/backend.js
+++ b/ext/js/background/backend.js
@@ -469,6 +469,7 @@ class Backend {
             const valid = AnkiUtil.isNoteDataValid(note);
             if (!valid) { canAdd = false; }
             const info = {canAdd, valid, noteIds: null};
+
             results.push(info);
             if (!canAdd && valid) {
                 cannotAdd.push({note, info});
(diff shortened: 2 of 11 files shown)


ext/js/background/backend.js
```diff
@@ -480,8 +481,32 @@ class Backend {
...
+                    /* TODO: should i be selecting the first ID like this?
+                        looking at findNoteIds I can't see how we could find
+                        more than one ID, but an example that proves me wrong
+                        would be cool (and also how to get the right ID) */
+                    const noteId = noteIds[0];
+
+                    /* TODO: should I be catching any errors here? findNoteIds
+                        is not wrapped in try/catch, but I can't find a catch
+                        anywhere in the code that calls this function */
+                    const noteInfo = (await this._anki.notesInfo([noteId]))[0];
```
toasted-nutbread, line 497: Omitting try/catch is fine, let the frontend handle (or not) the error.

ext/js/background/backend.js
```diff
@@ -480,8 +481,32 @@ class Backend {
...
                 const noteIds = noteIdsArray[i];
+
                 if (noteIds.length > 0) {
                     cannotAdd[i].info.noteIds = noteIds;
+
+                    /* TODO: should i be selecting the first ID like this?
+                        looking at findNoteIds I can't see how we could find
+                        more than one ID, but an example that proves me wrong
+                        would be cool (and also how to get the right ID) */
+                    const noteId = noteIds[0];
```
toasted-nutbread, line 492: Should probably return note info for all notes, i.e. just return an array that matches `noteIds`. Would probably be best to use the `'multi'` API of AnkiConnect (`findNoteIds` does this).

You probably don't need to worry about too much optimization, but the ideal situation is that only one API call is made. I can optimize that later if you like.

ext/js/display/display.js
```diff
@@ -1112,6 +1112,17 @@ class Display extends EventDispatcher {
                 }
                 button.disabled = !canAdd;
                 button.hidden = false;
+
+                if (button.disabled && tags && tags.length >= 1) {
```
toasted-nutbread, line 1116: Based on comments in backend.js, the settings checks would go here; use `this._options`. Would also recommend just putting this new section of logic into a private function like `this._setupTagsIndicator(i, noteInfos);`

For the `if` statement:
* `if (Array.isArray(tags) && tags.length > 0) {`
* (`disabled` check likely unnecessary)

toasted-nutbread (commented):
One other thing that would be good: the tag button currently doesn't do anything and is only used to show a tooltip on mouse hover. For accessibility, it should also show a snackbar notification; see `_ankiNoteNotification` in display.js for an example.

> I'm not sure about code style/conventions/the way I'm handling data in `backend.js`. Particularly how I'm accessing user config, using anki-connect to get card tags, and also assuming that I will only ever get one ID for a particular note when calling `findNoteIds`. I go into detail with comments in lines 488-505 in `backend.js`.

PR comments mostly cover this.
`findNoteIds` _can_ return multiple results, even if display.js doesn't use them (currently). It's probably rare, but it's permitted if a user disables note duplication checks.

> A different icon should probably be used. I quickly picked one at random from the icon folder. I'm not sure if there are any design guidelines or any other factor to decide on icons, but the current magnifying glass is a bit hard to see against a dark background.

We don't have a tag SVG currently, but I could make one pretty easily.

> We could consider adding an option to enable/disable this behaviour, especially if fetching the required data from anki-connect turns out to be expensive.

Agree, there are two considerations here:
1. What you mentioned
2. Unexpected changes to the UI that users may not expect, potentially not working well with their workflow (maybe they tag for different reasons)

soriac:
Wow... this is the best reply to a PR I've ever gotten. I just pushed two commits, the first should resolve most of the comments, and the second adds two toggles to the options menu, to enable/disable the tag button and enable/disable filtering out tags set by the user above.

ext/js/display/display.js
```diff
@@ -1112,13 +1113,64 @@ class Display extends EventDispatcher {
...
+
+        let optionTags = this._options.anki.tags;
+        if (!Array.isArray(optionTags)) {
+            optionTags = [];
+        }
+
+        // in case a user has disabled note duplication check we look at the tags for all notes
+        // that were found. this seems to fit the purpose of showing tags better
+        // (i already have this word added but i still want to pay special attention to it for some reason)
+        let tags = noteInfos.reduce((ts, noteInfo) => ts.concat(noteInfo.tags), []);
```
toasted-nutbread, line 1143: Since this is a highly-traversed code path, we should optimize the tags intersection check, something like this:
```js
const noteTags = new Set();
for (const {tags} of noteInfos) {
  for (const tag of tags) {
    noteTags.add(tag);
  }
}
for (const tag of optionTags) {
  noteTags.delete(tag);
}
```

(Rationale: standard for..of loops with a `Set` should be faster than `Array`'s `reduce/concat/filter/includes`, and use less memory)

ext/settings.html
```diff
@@ -1581,6 +1581,47 @@ <h1>Yomichan Settings</h1>
...
+            <div class="settings-item-inner">
+                <div class="settings-item-left">
+                    <div class="settings-item-label">Show custom card tags</div>
+                    <div class="settings-item-description">
+                        Show an icon next to the "view added note" button that shows all tags on that note (if it has
+                        any).
+                    </div>
+                </div>
+                <div class="settings-item-right">
+                    <label class="toggle">
```
toasted-nutbread, line 1594: Use the single-line layout for toggles (used elsewhere), otherwise it adds a lot of extra noise. (It would be nice if [web components](https://developer.mozilla.org/en-US/docs/Web/Web_Components) could be used so it's just a single element, but the minimum browser versions prevent this sadly, hence the repetition)

(My comment above in this file may render this feedback unnecessary)

ext/settings.html
```diff
@@ -1581,6 +1581,47 @@ <h1>Yomichan Settings</h1>
                 <button class="icon-button"><span class="icon-button-inner"><span class="icon" data-icon="material-right-arrow"></span></span></button>
             </div>
         </div></div>
+        <div class="settings-item advanced-only">
```
toasted-nutbread, line 1584: It would probably be better to use a single enum dropdown option for this instead of two toggles, something like:

**Show Anki card tags**

* Never
* Always
* Non-standard tags

Maybe add a More.../(?) link to explain what _non-standard tags_ are, if you think it's unclear.

ext/css/display.css
```diff
@@ -691,6 +691,15 @@ button.action-button[data-icon=play-audio]::before {
 button.action-button[data-icon=source-term]::before {
     background-image: url('/images/source-term.svg');
 }
+button.action-button[data-icon=magnifying-glass]::before {
```
toasted-nutbread, line 694: `button.action-button[data-icon=magnifying-glass]::before` rule no longer necessary

ext/js/background/backend.js
```diff
@@ -480,8 +480,10 @@ class Backend {
             const noteIdsArray = await this._anki.findNoteIds(cannotAddNotes);
             for (let i = 0, ii = Math.min(cannotAdd.length, noteIdsArray.length); i < ii; ++i) {
                 const noteIds = noteIdsArray[i];
+
```
toasted-nutbread, line 483: Remove stray newline

toasted-nutbread (commented):
I hope it's not too overbearing, since I'm leaving a lot of feedback. A few more things:

* You have the tag SVG graphic merged into this PR; rebase onto FooSoft/master so that you get the latest changes. I think this should automatically remove the duplicate code from your branch. I don't know your level of familiarity with git, so I'll include the following just in case:
  ```sh
  # Ensure you have a remote pointing for the main repo
  git remote add foosoft https://github.com/FooSoft/yomichan.git
  # Fetch
  git fetch foosoft
  # Rebase onto the main repo's master; this should be automatic
  git rebase foosoft/master
  # Force push afterwards
  git push -f
  ```

* The new option(s) should be added to the options schema, and tests updated (test-options-util.js; run `npm test`)
  https://github.com/FooSoft/yomichan/blob/763c6c76aa5bceb1eb929aa3a87bacf91e37407f/ext/data/schemas/options-schema.json#L797

soriac:
No, this is not overbearing at all. (A bit OT but i've only worked at (way too agile) small startups and as a freelancer, so I'm used to very lax PRs and code requirements that always turn into broken code. But this is way nicer.)

I ended up adding a _More..._ link to explain the new setting in general, and non-standard tags in particular, but it might not be clear enough... I was thinking of throwing in a real use-case, maybe something like:

_This could be useful for cards where you realize that your current context is not enough, or maybe you forgot to grab a piece of context, so you can tag them and then easily tell that this card needs fixing the next time you come across that word._

But that might be going into too much detail.

ext/js/display/display.js
```diff
@@ -1112,13 +1114,60 @@ class Display extends EventDispatcher {
...
+        if (displayTags === 'non-standard') {
+            for (const tag of optionTags) {
+                noteTags.delete(tag);
+            }
+        }
+
+        if (noteTags.size > 0) {
+            tagsIndicator.disabled = false;
+            tagsIndicator.hidden = false;
+            tagsIndicator.title = `Card tags: ${Array.from(noteTags).join(', ')}`;
```
toasted-nutbread, line 1150: `[...noteTags]` instead of `Array.from(noteTags)` (that's the more usual convention for yomichan)

ext/js/background/backend.js
```diff
@@ -482,6 +482,7 @@ class Backend {
                 const noteIds = noteIdsArray[i];
                 if (noteIds.length > 0) {
                     cannotAdd[i].info.noteIds = noteIds;
+                    cannotAdd[i].info.noteInfos = await this._anki.notesInfo(noteIds);
```
toasted-nutbread, line 485: Since this information is not used in the default use case, let's add another argument like `getNoteInfos` to the signature of the `getAnkiNoteInfo`/`_onApiGetAnkiNoteInfo` API functions. Can probably start the passing it from here:

https://github.com/FooSoft/yomichan/blob/937e118da23da3e87c0c8433f5a2a94dee326864/ext/js/display/display.js#L1085-L1086

ext/settings.html
```diff
@@ -1581,6 +1581,34 @@ <h1>Yomichan Settings</h1>
...
+                    <select data-setting="anki.displayTags">
+                        <option value="never">Never</option>
+                        <option value="always">Always</option>
+                        <option value="non-standard">Non-Standard</option>
+                    </select>
+                </div>
+            </div>
+            <div class="settings-item-children more" hidden>
+                <p>
+                    When coming across a word that is already in an anki deck, show an icon that lets you quickly see
```
toasted-nutbread, line 1602: Some slight wording tweaks:

* > When coming across a word that is already in an anki deck, show an icon that lets you quickly see which tags that card has.

   to 
   
  > When coming across a word that is already present in an Anki deck, a button will appear that shows the tags the card has.
  
  Rationale:
  * Anki capitalized
  * Slight rewording to omit the word "you", which is avoided for most descriptions.

<br>

* > If set to &lt;em&gt;Non-Standard&lt;/em&gt; then all tags ...

  to
  
  > If set to &lt;em&gt;Non-Standard&lt;/em&gt;, all tags ...
  
  Rationale: comma should be there anyway, but removing "then" results in less repetition with the following sentence.

toasted-nutbread:
* `options-util.js` needs to be updated; add the change to `_updateVersion11` by adding:
   ```js
   //  Added anki.displayTags.
   ```
   ```js
   profile.options.anki.displayTags = 'never';
   ```
   Check to make sure everything is set up correctly by running `npm run`. (Github runs CI, but I guess it doesn't run for PRs that aren't approved yet.)
* The new button should likely be set to `display:none` when it has the hidden attribute. Currently, the other buttons are just set to `visibility:none`, which means they still take up layout space. I would consider this new button a different case, since it's a non-standard and potentially uncommon workflow, at least currently.

soriac:
I tried to run `npm run` but it just prints a list of commands. I've been running `npm run test` before every push and it seems to work fine. It's not printing any error and building for all 5 targets.

ext/css/display.css
```diff
@@ -644,6 +644,9 @@ button.action-button[hidden] {
         -webkit-filter var(--animation-duration) linear,
         background-color var(--animation-duration) linear;
 }
+button.collapsable-action-button[hidden] {
```
toasted-nutbread, line 647: This is a super minor nitpick, but change `collapsable` to `collapsible` (using an "i"). Search the code for both words and you'll see the one with the "i" is mainly used; there are a few other cases with the "a" that I'll fix up soon.

ext/js/data/options-util.js
```diff
@@ -792,6 +792,7 @@ class OptionsUtil {
         // Version 11 changes:
         //  Changed dictionaries to an array.
         //  Changed audio.customSourceUrl's {expression} marker to {term}.
+        //  Added anki.displayTags
```
toasted-nutbread, line 795: Another very minor nitpick: add a period at the end for consistency with other note comments.

ext/js/pages/settings/anki-controller.js
```diff
@@ -61,6 +61,8 @@ class AnkiController {
         this._ankiEnableCheckbox = document.querySelector('[data-setting="anki.enable"]');
         this._ankiCardPrimary = document.querySelector('#anki-card-primary');
         const ankiCardPrimaryTypeRadios = document.querySelectorAll('input[type=radio][name=anki-card-primary-type]');
+        this._ankiShowCustomTagsToggle = document.querySelector('#show-tags-enabled');
```
toasted-nutbread, line 64: I don't think the changes in this file are necessary any more.

toasted-nutbread:
My bad, I did mean `npm test`/`npm run test`; there was a minor issue previously with the settings upgrade which is fixed now. Looking good so far, a few last things:
* On the settings page, move the _Show card tags_ option below the _Suspend new cards_ option (i.e. above the two _Configure..._ options at the end).
* I commented a few minor nitpicks that I would have skipped if I didn't have other feedback.

After that, I should be be able to merge this in tomorrow.

toasted-nutbread:
Merged in, thanks for the contribution! If you have any other feedback related to any issues you had with the codebase or anything, let me know.

Issue #1824, opened 2021-07-13
[Feature Request] Indication for Nasal (鼻濁音) and Devoiced (無声化) Sounds


Shudouken:
The NHK accent dictionary provides accent information for nasal and devoiced sounds, however those seem to be unsupported by yomichan.
Nasal sounds are indicated using a red 半濁点 ゜ over the が行, and devoicing of vowels is indicated with a red ◌ surrounding the mora.
![nhk](https://user-images.githubusercontent.com/1529740/125476882-4167a801-f926-4140-ba28-f58f23da0efd.png)
This accent information could be displayed inside of the yomichan popup window, in combination with the pitch accent information.
Mockup:
![yomi](https://user-images.githubusercontent.com/1529740/125484199-37e2bfba-9779-4aa2-8f63-e7dd1adee722.png)

I'm also trying to replicate this notation in anki with two custom classes.
It would be great if i could let the yomichan handlebars do the work of adding those classes instead of having to add them manually.
![file](https://user-images.githubusercontent.com/1529740/125477357-e284f383-bb10-42bf-952b-4d1a9562762b.png)
<details>
  <summary>My Anki CSS (simplified)</summary>
  
  ```javascript
.museika:before {
    content: "◌";
    color: #cc0033;
    position: absolute;
}

.bidakuon:after {
    content: "ﾟ";
    color: #cc0033;
    position: relative;
}
  ```
  
</details>

I tried playing around and making a custom dictionary with position data for nasal and devoiced sounds, but it seems yomichan ignores json fields it is not aware of (please correct me if I'm wrong).
Or I simply couldn't figure out how to access them in the handlebars section of yomichan.

I would suggest adding it to the pitch accent dictionary syntax, something like this:
`[["番号","pitch",{"reading":"ばんごう","pitches":[{"position":3}],"nasal":[{"position":3}]}]]`
Then it could be used in the handlebars to wrap the ご mora with a custom class or style for nasal sounds.
Similarly for devoiced moras.

For reference, this [anki addon](https://github.com/javdejong/nhk-pronunciation) has somewhat similar functionality and uses a csv file with the NHK accent information, however it would be nice to have this functionality inside of yomichan as well.

PR #1832, opened 2021-07-16
Devoice and nasal pronunciation info

toasted-nutbread:
This change adds support for devoiced and nasal characters in dictionary pronunciation info. [term_meta_bank_1.json](https://github.com/toasted-nutbread/yomichan/blob/99a593bcf865297dec2f2fdeb9841e0dd9f14067/test/data/dictionaries/valid-dictionary1/term_meta_bank_1.json#L45-L74) contains examples of the data format.

Addresses part of #1824.

### DIFF
--- a/docs/interfaces/dictionary-entry.ts
+++ b/docs/interfaces/dictionary-entry.ts
@@ -355,6 +355,14 @@ namespace Translation {
          */
         position: number;
         /**
+         * Positions of moras that have a .
+         */
+        nasalPositions: number[];
+        /**
+         * Position of the downstep, as a number of mora.
+         */
+        devoicePositions: [];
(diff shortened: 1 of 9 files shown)


Shudouken:
For nasal sounds it might look better to replace が for か etc. in the display generator since the ゛isn't always in the same position for every mora and also might depend on the font.

I use this on my anki template for example

```
var s = document.getElementsByClassName("nasal");

[].forEach.call(s, function (el) {el.innerHTML = el.innerHTML
  .replace("が","か")
  .replace("ぎ","き")
  .replace("ぐ","く")
  .replace("げ","け")
  .replace("ご","こ")
});
```

PR #1843, opened 2021-07-19
List `vz` rule identifier in term bank schema

Thermospore:
Also could you describe the purpose of the `iru` rule identifier further?


toasted-nutbread:
`iru` is an internal flag used to form the progressive "tense" and should not be used in dictionary definitions. It is used by the deinflector to mark deinflections that end with a "te" form.

PR #1878, opened 2021-08-07
Add support for manifest V3 script injection

toasted-nutbread:
Adds support for script injection when using Manifest V3. Note that the behaviour is slightly different, maybe because MV3 is not ready yet:
* [`chrome.tabs.executeScript`](https://developer.chrome.com/docs/extensions/reference/tabs/#method-executeScript) supports `runAt = 'document_start'`, [`chrome.scripting.executeScript`](https://developer.chrome.com/docs/extensions/reference/scripting/#method-executeScript) does not.
* MV2's injection of a `<script>` element uses the extension's CSP; MV3's uses the webpage's CSP. Therefore, `google-docs.js` currently does not work in MV3.

Follow-up of #1875.

### DIFF
--- a/ext/js/background/backend.js
+++ b/ext/js/background/backend.js
@@ -2164,6 +2164,61 @@ class Backend {
         });
     }
 
+    _injectScript(file, tabId, frameId) {
+        if (isObject(chrome.tabs) && typeof chrome.tabs.executeScript === 'function') {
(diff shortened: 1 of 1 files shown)


toasted-nutbread:
Related:
https://bugs.chromium.org/p/chromium/issues/detail?id=1217895
https://bugs.chromium.org/p/chromium/issues/detail?id=1137396

PR #1938, opened 2021-09-08
Frequency dictionary sort

### LINKED ISSUE #1742: [Feature Request] Handle frequency ranking on a global basis.
Currently the frequency ranking is handled within the dictionary format itself. This means that you either have to always have the provided jmdict dictionary imported and activated or have to manually export the freq info of the jmdict into the dictionary you want to actually use. This behaviours is fine if you use jmdict but it's quite annoying when trying to use monolingual dictionaries, or even when making new dictionaries as it means you have to re-match all the same frequencies.

Wouldn't it be much better to just handle the freq info for all dictionaries at the addon level?

I don't know if it's useful in any ways but I've compiled a tsv with all the possible reading/writing combos, their frequency and ID based on the available jmdict dictionary.
https://github.com/epistularum/japanese-word-frequency

I've done some small changes like:

- Removing duplicate reading/writing combos (only keeping the highest freq for each combo)
- Adding more granular freq for reading/writing combos with the same ID and same freq (was an issue for things like 鳩尾[みずおち] vs 鳩尾[みぞおち] vs 水落[みぞおち] vs 水落[みずおち] which all had the same freq). This was done by prioritizing the order those entries are found in JMdict. Decimals were used in those cases in order to not mess with other possible matches.
- Getting rid of negative values by scaling up all frequencies. This is more of a practical change, I find it hard to handle negative values because they don't get sorted correctly when doing numerical sorts.

toasted-nutbread:
This change adds support for using a frequency dictionary to sort search results.

Resolves #1742.

### DIFF
--- a/ext/css/settings.css
+++ b/ext/css/settings.css
@@ -2259,9 +2259,13 @@ input[type=number].dictionary-priority {
 }
 .horizontal-flex.horizontal-flex-nowrap {
     flex-wrap: nowrap;
+    margin-left: 0;
 }
 .horizontal-flex>* {
     margin-left: 0.375em;
+}
+.horizontal-flex.horizontal-flex-nowrap>*:first-child {
+    margin-left: 0;
 }
 .horizontal-flex-fill {
     flex-grow: 1;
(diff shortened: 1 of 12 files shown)


Ceynou:
Can the parser result overwrite the frequency sorting? for exemple mecab giving a less used reading of a compound word, appearing first, before the most frequent reading?

toasted-nutbread:
Sentence parsing does not affect term sorting, it is an entirely separate (and optional) process that occurs at a different time.

Issue #2046, opened 2022-01-04
Send AnkiConnect requests with `application/json` as the content type


archiif:
Yomichan sends AnkiConnect requests with `text/plain` as the content type instead of `application/json`.
My script expects POST requests containing json data to have `application/json` as the content type, so this quirk caused requests from Yomichan to be parsed incorrectly by the script.
In any case, I think having `application/json` as the content type should be the correct way of performing this kind of requests anyways.

PR #2047, opened 2022-01-04
Set `Content-Type` header of AnkiConnect requests to `application/json`

archiif:
Fixes https://github.com/FooSoft/yomichan/issues/2046


FooSoft:
Looks good, thanks!

Issue #1165, opened 2020-12-27
[Feature Request] add support for JMdict `note`, `see`, and `ant`


Thermospore:
These often contain critical information. I don't think JMdict should be considered complete without them

Here are some entries that use these, for reference (though there are probably better entries to use as an example of their importance):
http://www.edrdg.org/jmdictdb/cgi-bin/entr.py?svc=jmdict&sid=&q=1465580.1
http://www.edrdg.org/jmdictdb/cgi-bin/entr.py?svc=jmdict&sid=&q=1456360.1
http://www.edrdg.org/jmdictdb/cgi-bin/entr.py?svc=jmdict&sid=&q=1339460.1

Screenshots of what I'm referring to specifically
![image](https://user-images.githubusercontent.com/35350819/103169276-b6cefa00-47e6-11eb-8bbb-ed08eb83a809.png)
![image](https://user-images.githubusercontent.com/35350819/103169324-06adc100-47e7-11eb-9863-e023bbb984bb.png)
![image](https://user-images.githubusercontent.com/35350819/103169332-1cbb8180-47e7-11eb-9029-560d83276701.png)

(I think this should technically be in the yomichan import github, but this seems to be where the main activity happens)

Thermospore:
Looks like they store the notes as `s_inf` in the jmdict xml file. Judging from https://godoc.org/github.com/FooSoft/jmdict I think if you just slap `term.addDefinitionTags(sense.Information...)` in line 181 of https://github.com/FooSoft/yomichan-import/blob/master/edict.go then notes would show up in the definition tags in yomichan?

FooSoft:
@toasted-nutbread I am going to make a pass on yomichan-import to clean up some of the dictionary processing (especially for EPWING), let me know if you have a list of format annoyances to take care of while I am at it.

toasted-nutbread:
@FooSoft [https://github.com/FooSoft/yomichan/labels/dictionary format](https://github.com/FooSoft/yomichan/labels/dictionary%20format) lists all the issues I am aware of. Most of them related to Kenkyusha, with some being somewhat cosmetic (due to large size of definitions), whereas a few others impact functionality (terms not marked as verbs, or unexpected reading/expression form).

Related to this issue: if we want to support see/note/ant, it may be best to present them differently than standard definitions (as seen in the images, e.g. brackets, crosslinks). This may require some metadata updates to the dictionary format Yomichan uses, perhaps something similar to how image definitions work.

[...]

PR #2129, opened 2022-05-13
Add new structured content features: lists and the HTML `lang` attribute

stephenmk:
## Structured Content `lang` Attribute 

Authors of dictionaries for Yomichan may want to include content from various languages, so I think this is a valuable feature. JMDict, for example, includes Japanese loanword source-words from over 60 different languages.

<details>
  <summary>Characters 直次茶冷 displayed in structured content glosses</summary>

⚠The current version of Yomichan seems to apply `lang="ja"` attributes to standard glosses which contain Japanese characters, but not to structured content glosses. By default, most browsers will render the characters "直次茶冷" in simplified Chinese. 
html_lang
</details>

## Structured Content Lists
The current way in which I have inserted supplemental information into JMdict glossaries (see: https://github.com/FooSoft/yomichan/issues/1165) is a little awkward. When the "compact glossaries" option is enabled, all of the information is grouped and compacted together. It will be easier for the user to parse the information if the different types of information are broken into separate sections. My idea is to break them up into separate unordered lists, each with its own `list-style-type`.

<details>
  <summary>読む</summary>

yomu
</details>

<details>
  <summary>読む (compact glossaries)</summary>

yomu
</details>

<details>
  <summary>元 ("compact glossaries" and "group related terms" mode)</summary>

moto
</details>

<details>
  <summary>アルバイト</summary>

arubaito
</details>

<details>
  <summary>欠席</summary>

kesseki
</details>

<details>
  <summary>ちりも積もれば山となる</summary>

chiri
</details>

I've included tests for these new structured content features in the file `test/data/dictionaries/valid-dictionary1/term_bank_1.json`.

Additionally, here is a new version of JMdict for Yomichan which uses the new features.
[jmdict_english_info_glosses_2022_05_13.zip](https://github.com/FooSoft/yomichan/files/8690986/jmdict_english_info_glosses_2022_05_13.zip)

This version takes over 30 minutes to validate during the import process, so it is probably not viable for distribution unless Yomichan's validation procedure is optimized. Glosses that do not contain supplemental information are not inserted into structured content containers (they are formatted identically to the current production version of the dictionary), so I don't think there are any additional optimizations I can make to the dictionary without cutting content.

Here is a version of the new JMdict dictionary that does not contain external reference notes (i.e. notes that indicate when an entry is referenced by another entry). On my PC it takes about 10 minutes to validate.
[jmdict_english_info_glosses_no_ext_xrefs_2022_05_13.zip](https://github.com/FooSoft/yomichan/files/8691017/jmdict_english_info_glosses_no_ext_xrefs_2022_05_13.zip)

I'm open to suggestions on how to improve the appearance of the new JMdict dictionary. I still need to clean up the code in my branch of yomichan-import a bit, but I think I'm out of ideas for additional features to add.

### DIFF
--- a/dev/data/structured-content-overrides.css
+++ b/dev/data/structured-content-overrides.css
@@ -61,3 +61,16 @@
     display: none;
     /* remove-property background-color vertical-align width height margin-left background-color position */
 }
+.gloss-sc-ol,
+.gloss-sc-ul {
+    /* remove-property padding-left */
+}
+:root[data-glossary-layout-mode=compact] .gloss-sc-ul[data-sc-content=glossary] {
+    /* remove-rule */
+}
+:root[data-glossary-layout-mode=compact] .gloss-sc-ul[data-sc-content=glossary] .gloss-sc-li {
+    /* remove-rule */
+}
+:root[data-glossary-layout-mode=compact] .gloss-sc-ul[data-sc-content=glossary] .gloss-sc-li:not(:first-child)::before {
+    /* remove-rule */
+}
--- a/ext/data/schemas/dictionary-term-bank-v3-schema.json
+++ b/ext/data/schemas/dictionary-term-bank-v3-schema.json
@@ -95,7 +95,7 @@
                             "properties": {
                                 "tag": {
                                     "type": "string",
-                                    "enum": ["span", "div"]
+                                    "enum": ["span", "div", "ol", "ul", "li"]
                                 },
                                 "content": {
                                     "$ref": "#/definitions/structuredContent"
@@ -276,6 +276,10 @@
                 "marginBottom": {
                     "type": "number",
                     "default": 0
+                },
+                "listStyleType": {
+                    "type": "string",
+                    "default": "disc"
                 }
             }
         }
--- a/ext/data/schemas/dictionary-term-bank-v3-schema.json
+++ b/ext/data/schemas/dictionary-term-bank-v3-schema.json
@@ -54,7 +54,7 @@
                                 },
                                 "lang": {
                                     "type": "string",
-                                    "description": "Defines the language of an element in the format defined by RFC 5646"
+                                    "description": "Defines the language of an element in the format defined by RFC 5646."
                                 }
                             }
                         },
@@ -89,7 +89,7 @@
                                 },
                                 "lang": {
                                     "type": "string",
-                                    "description": "Defines the language of an element in the format defined by RFC 5646"
+                                    "description": "Defines the language of an element in the format defined by RFC 5646."
                                 }
                             }
                         },
@@ -116,7 +116,7 @@
                                 },
                                 "lang": {
                                     "type": "string",
-                                    "description": "Defines the language of an element in the format defined by RFC 5646"
+                                    "description": "Defines the language of an element in the format defined by RFC 5646."
                                 }
                             }
                         },
@@ -221,7 +221,7 @@
                                 },
                                 "lang": {
                                     "type": "string",
-                                    "description": "Defines the language of an element in the format defined by RFC 5646"
+                                    "description": "Defines the language of an element in the format defined by RFC 5646."
                                 }
                             }
                         }
--- a/ext/data/schemas/dictionary-term-bank-v3-schema.json
+++ b/ext/data/schemas/dictionary-term-bank-v3-schema.json
@@ -51,6 +51,10 @@
                                 },
                                 "data": {
                                     "$ref": "#/definitions/structuredContentData"
+                                },
+                                "lang": {
+                                    "type": "string",
+                                    "description": "Defines the language of an element in the format defined by RFC 5646"
                                 }
                             }
                         },
@@ -82,6 +86,10 @@
                                 },
                                 "style": {
                                     "$ref": "#/definitions/structuredContentStyle"
+                                },
+                                "lang": {
+                                    "type": "string",
+                                    "description": "Defines the language of an element in the format defined by RFC 5646"
                                 }
                             }
                         },
@@ -105,6 +113,10 @@
                                 },
                                 "style": {
                                     "$ref": "#/definitions/structuredContentStyle"
+                                },
+                                "lang": {
+                                    "type": "string",
+                                    "description": "Defines the language of an element in the format defined by RFC 5646"
                                 }
                             }
                         },
@@ -206,6 +218,10 @@
                                     "type": "string",
                                     "description": "The URL for the link. URLs starting with a ? are treated as internal links to other dictionary content.",
                                     "pattern": "^(?:https?:|\\?)[\\w\\W]*"
+                                },
+                                "lang": {
+                                    "type": "string",
+                                    "description": "Defines the language of an element in the format defined by RFC 5646"
                                 }
                             }
                         }
--- a/ext/css/structured-content.css
+++ b/ext/css/structured-content.css
@@ -235,3 +235,23 @@
     padding: 0.25em;
     vertical-align: top;
 }
+.gloss-sc-ol {
+    padding-left: var(--list-padding2);
+}
+.gloss-sc-ul {
+    padding-left: var(--list-padding2);
+}
+:root[data-glossary-layout-mode=compact] .gloss-sc-ul[data-sc-content=glossary] {
+    display: inline;
+    list-style: none;
+    padding-left: 0;
+}
+:root[data-glossary-layout-mode=compact] .gloss-sc-ul[data-sc-content=glossary] .gloss-sc-li {
+    display: inline;
+}
+:root[data-glossary-layout-mode=compact] .gloss-sc-ul[data-sc-content=glossary] .gloss-sc-li:not(:first-child)::before {
+    white-space: pre-wrap;
+    content: var(--compact-list-separator);
+    display: inline;
+    color: var(--text-color-light3);
+}
--- a/ext/css/structured-content.css
+++ b/ext/css/structured-content.css
@@ -235,9 +235,7 @@
     padding: 0.25em;
     vertical-align: top;
 }
-.gloss-sc-ol {
-    padding-left: var(--list-padding2);
-}
+.gloss-sc-ol,
 .gloss-sc-ul {
     padding-left: var(--list-padding2);
 }
--- a/ext/js/display/sandbox/structured-content-generator.js
+++ b/ext/js/display/sandbox/structured-content-generator.js
@@ -56,6 +56,9 @@ class StructuredContentGenerator {
                 return this._createStructuredContentElement(tag, content, dictionary, 'table-cell', true, true);
             case 'div':
             case 'span':
+            case 'ol':
+            case 'ul':
+            case 'li':
                 return this._createStructuredContentElement(tag, content, dictionary, 'simple', true, true);
             case 'img':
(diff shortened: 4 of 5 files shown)


ext/data/schemas/dictionary-term-bank-v3-schema.json
```diff
@@ -51,6 +51,10 @@
                                 },
                                 "data": {
                                     "$ref": "#/definitions/structuredContentData"
+                                },
+                                "lang": {
+                                    "type": "string",
+                                    "description": "Defines the language of an element in the format defined by RFC 5646"
```
toasted-nutbread, line 57: Nitpick: add period to the end of description value (other descriptions are punctuated).

Apply to other locations in the file also.

toasted-nutbread:
Looks good, just a few comments on CSS:

The way that Anki cards have styling information generated for structured content is a special case which needs to be handled.

1. Modify [dev/data/structured-content-overrides.css](https://github.com/FooSoft/yomichan/blob/master/dev/data/structured-content-overrides.css) to create overrides for new additions to the [ext/css/structured-content.css](https://github.com/stephenmk/yomichan/blob/5d57f587f37a43a39c91d347187542f2f3bcdc10/ext/css/structured-content.css#L238-L257) file. You'll want to hardcode `var()` values (just use its computed value) and to flag the `:root...` properties using <br> `/* remove-rule */`, since they aren't necessary for Anki.
2. You'll have to run this to update the [ext/data/structured-content-style.json](https://github.com/FooSoft/yomichan/blob/master/ext/data/structured-content-style.json) file:
   `node dev/generate-css-json.js`

toasted-nutbread:
> ⚠The current version of Yomichan seems to apply `lang="ja"` attributes to standard glosses which contain Japanese characters, but not to structured content glosses. By default, most browsers will render the characters "直次茶冷" in simplified Chinese.

Out of scope for this PR, but this may be something that's worth updating at some point also, since this seems like an oversight on my part. (not saying this is something you should do)

toasted-nutbread:
> This version takes over 30 minutes to validate during the import process, so it is probably not viable for distribution unless Yomichan's validation procedure is optimized. Glosses that do not contain supplemental information are not inserted into structured content containers (they are formatted identically to the current production version of the dictionary), so I don't think there are any additional optimizations I can make to the dictionary without cutting content.

And to this point, yes I agree that the validation is unfortunately slow. This is partially due to how non-specific JSON schemas are technically allowed to be, and I made a few optimizations a while back, but I should revisit this.

It's also potentially a motivation for having a different way to represent the type of data mentioned in #1165. While structured content can accomplish it, it is by no means the optimal way of doing it. Compare something like:

<details>
  <summary>Current</summary>

```jsonc
[
  "読む",
  "よむ",
  "5 v5m vt",
  "v5",
  1440001,
  [
    {
      "content": [
        {
          "content": {
            "content": "now mostly used in idioms",
            "tag": "li"
          },
          "data": {
            "content": "notes"
          },
          "lang": "ja",
          "style": {
            "listStyleType": "'📝 '"
          },
          "tag": "ul"
        },
        {
          "content": [
            {
              "content": "to count",
              "tag": "li"
            },
            {
              "content": "to estimate",
              "tag": "li"
            }
          ],
          "data": {
            "content": "glossary"
          },
          "lang": "en",
          "style": {
            "listStyleType": "circle"
          },
          "tag": "ul"
        },
        {
          "content": {
            "content": [
              "see: ",
              {
                "content": "さばを読む",
                "href": "?query=さばを読む\u0026wildcards=off",
                "lang": "ja",
                "tag": "a"
              },
              {
                "content": " to manipulate figures to one's advantage; to count wrongly on purpose; to inflate or deflate one's age",
                "data": {
                  "content": "refGlosses"
                },
                "style": {
                  "fontSize": "x-small",
                  "verticalAlign": "middle"
                },
                "tag": "span"
              }
            ],
            "tag": "li"
          },
          "data": {
            "content": "references"
          },
          "lang": "en",
          "style": {
            "listStyleType": "'➡ '"
          },
          "tag": "ul"
        }
      ],
      "type": "structured-content"
    }
  ],
  1456360,
  "P ichi news6k"
]
```
</details>

<details>
  <summary>Minimized</summary>

```jsonc
[
  "読む",
  "よむ",
  "5 v5m vt",
  "v5",
  1440001,
  [
    {
      "content": "now mostly used in idioms",
      "type": "note"
    },
    "to count",
    "to estimate",
    {
      "content": "さばを読む",
      "brief": "to manipulate figures to one's advantage; to count wrongly on purpose; to inflate or deflate one's age", // optional
      "href": "?query=さばを読む\u0026wildcards=off", // optional
      "type": "references"
    }
  ],
  1456360,
  "P ichi news6k"
]
```
</details>

---

For comparison, I also ran the validation on the dictionary you provided and it took roughly 18 minutes:
```
node dev\dictionary-validate.js jmdict_english_info_glosses_2022_05_13.zip
Validating jmdict_english_info_glosses_2022_05_13.zip...
No issues detected (1102.21s)
```

stephenmk:
Thanks for the detailed feedback. The files have now been updated.

>You'll want to hardcode `var()` values (just use its computed value) and to flag the `:root...` properties using  `/* remove-rule */`, since they aren't necessary for Anki.

I added the `--padding-left` rule to the structured content lists so that they would align with the regular glosses within the Yomichan popup, but it doesn't seem that the regular glosses have this padding when exported to Anki. So I think the correct move is to just drop the rule (rather than hardcode a value). If that doesn't work for some reason I didn't think of, it's no problem to update it again. I made a card to test and it seemed to come out alright.

<details>
  <summary>読む glossary in Anki</summary>

yomu_anki
</details>

dev/data/structured-content-overrides.css
```diff
@@ -61,3 +61,16 @@
     display: none;
     /* remove-property background-color vertical-align width height margin-left background-color position */
 }
+.gloss-sc-ol,
+.gloss-sc-ul {
+    /* remove-property padding-left */
+}
+:root[data-glossary-layout-mode=compact] .gloss-sc-ul[data-sc-content=glossary] {
```
toasted-nutbread, line 68: Group these together for simplicity:
```css
:root[data-glossary-layout-mode=compact] .gloss-sc-ul[data-sc-content=glossary],
:root[data-glossary-layout-mode=compact] .gloss-sc-ul[data-sc-content=glossary] .gloss-sc-li,
:root[data-glossary-layout-mode=compact] .gloss-sc-ul[data-sc-content=glossary] .gloss-sc-li:not(:first-child)::before {
    /* remove-rule */
}
```

stephenmk: Ha, that's actually what I tried first. But the tests do not like that. I think that only works if they are defined that way (as a group) in the `ext/css/structured-content.css` file. But the rules are all different there, so they can't be combined.

```
Running test-css-json.js...
Error: Could not find rule with matching selectors
    at generateRules (/home/stephen/Code/yomichan/dev/css-to-json-util.js:139:48)
    at main (/home/stephen/Code/yomichan/test/test-css-json.js:28:42)
    at testMain (/home/stephen/Code/yomichan/dev/util.js:127:15)
    at Object.<anonymous> (/home/stephen/Code/yomichan/test/test-css-json.js:35:5)
    at Module._compile (node:internal/modules/cjs/loader:1105:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1159:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Module._load (node:internal/modules/cjs/loader:827:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:77:12)
    at node:internal/main/run_main_module:17:47
```

toasted-nutbread: Hmm, I thought I handled that differently, but I guess not; my bad! (I should really be more familiar with my own code)

toasted-nutbread:
> Out of scope for this PR, but this may be something that's worth updating at some point also, since this seems like an oversight on my part. (not saying this is something you should do)

#2131

stephenmk:
> It's also potentially a motivation for having a different way to represent the type of data mentioned in #1165. While structured content can accomplish it, it is by no means the optimal way of doing it. 

I think this is an interesting idea. It seems simple enough that even I could probably implement it, and it would probably cut back on a lot of validation time. I think a major drawback to the idea (as I understand it) is that it would very tightly couple the dictionary's display logic with yomichan, which would require [1] yomichan to be updated whenever major changes occur in the dictionary format and [2] yomichan to maintain compatibility with all prior dictionary formats (to support users who do not upgrade to the newest version of the dictionary).

If the new data type representations are sufficiently modular, perhaps this would not be an issue. However, I'm not sure we're going to be able to strike a good balance between modularity and conciseness with this JMdict data that would be worth the effort. Here's an example:

<details>
  <summary>Minimized JMdict term bank entry with hypothetical data types</summary>

```json
[
  "読む",
  "よむ",
  "5 v5m vt",
  "v5",
  1440001,
  {
    "type": "jmdictglossary",
    "sources": [
      {
        "lang": "zh",
        "language": "Chinese",
        "wasei": false,
        "content": "讀・dú",
        "type": "partial"
      }
    ],
    "notes": [
      "now mostly used in idioms"
    ],
    "glosses": [
      "to count",
      "to estimate"
    ],
    "infoGlosses": [
      {
        "type": "expl",
        "content": "The Yomi in Yomichan comes from 読む"
      },
      {
        "type": "tm",
        "content": "FooSoft® 2022"
      }
    ],
    "references": [
      {
        "kanji": "さばを読む",
        "brief": " to manipulate figures to one's advantage; to count wrongly on purpose; to inflate or deflate one's age",
        "type": "reference"
      },
      {
        "kanji": "欠",
        "reading": "あくび",
        "brief": " 欠 can be read as けつ, あくび, or かけ, so a reading has to be specified",
        "type": "antonym"
      }
    ]
  },
  1456360,
  "P ichi news6k"
]
```
</details>

Note that the entire glossary is contained within a single `jmdictglossary` typed gloss object. If the different sections are broken up into separate glosses (as in your `Minimized` example), then each section will be displayed in a separate list item. This wouldn't look very good in the Yomichan popup without some other major style modifications, and `compact glossaries` mode would group them together in a big mess.

This hypothetical `jmdictglossary` type isn't modular, is unlikely to be useful to any other dictionary authors, and [is likely to require changes in the future](https://www.edrdg.org/wiki/index.php/JMdict:_Next_Generation). So I think that updating Yomichan to be able to use such a data type would probably end up causing a lot of hassle.

I don't know anything about JSON schema validation, and I haven't looked into how Yomichan does it or what engine it uses. My naive hope was that maybe Yomichan could be updated to use a faster validator ([ajv claims to be the fastest](https://github.com/ajv-validator/ajv)) and perhaps that would be enough to solve the problem.

toasted-nutbread:
Yeah, I agree overall that it's difficult to balance modularity, efficiency, and compatibility, hence why I haven't yet took the initiative to implement something like what I mentioned in https://github.com/FooSoft/yomichan/pull/2129#issuecomment-1126557391.

> I don't know anything about JSON schema validation, and I haven't looked into how Yomichan does it or what engine it uses.

A custom one I wrote, which supports a limited subset. I can look into doing a comparison vs ajv. The other downside of complex structured content vs native definitions is in the database storage overhead, since all of the formatting needs to be stored.

PR #2137, opened 2022-05-17
Dictionary validate updates

toasted-nutbread:
This change makes `dictionary-validate.js` and `schema-validate.js` able to use https://ajv.js.org as a validator for testing.

This was originally mentioned in https://github.com/FooSoft/yomichan/pull/2129#issuecomment-1126785466:

> I don't know anything about JSON schema validation, and I haven't looked into how Yomichan does it or what engine it uses. My naive hope was that maybe Yomichan could be updated to use a faster validator ([ajv claims to be the fastest](https://github.com/ajv-validator/ajv)) and perhaps that would be enough to solve the problem.

### DIFF
--- a/dev/dictionary-validate.js
+++ b/dev/dictionary-validate.js
@@ -19,15 +19,7 @@ const fs = require('fs');
 const path = require('path');
 const {performance} = require('perf_hooks');
 const {JSZip} = require('./util');
-const {VM} = require('./vm');
-
-const vm = new VM();
-vm.execute([
-    'js/core.js',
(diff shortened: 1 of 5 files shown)


stephenmk:
My new JMdict dictionary for Yomichan validates in 3.25 seconds on my PC now -- incredible!

toasted-nutbread:
Yeah, it's because their validator compiles to actual JavaScript code, which presents some potential challenges with integrating into Yomichan. This is a first step towards evaluating what can be done here.

Issue #2070, opened 2022-02-14
[Feature Request] Make “View added note” button open Edit card dialog, not Browse


oakkitten:
I use the button “View added note” a lot to add finishing touches to my cards. This opens the browser. Some parts of the browser, such as the sidebar and the note list, take a lot of space and cannot be collapsed fully. I suggest adding an option to open Edit card dialog instead.

I imagine that the “Edit card” dialog, which doesn't exist at the moment, should look just like the existing “Edit Current” dialog, perhaps sans the “Close” button. This would require changes in AnkiConnect, and perhaps in Anki itself. I'm opening an issue here as if this doesn't get a green light, there's no point in asking for changes elsewhere.

oakkitten:
For fun, I went and made the (proof of concept) Edit card dialog. I removed the close button, and added the Preview button as it appears in the Browser, and the Browse button that opens what Yomichan normally opens.
![image](https://user-images.githubusercontent.com/1710718/154821745-fc9a634b-5158-4ffe-a7bf-4ac6eace700b.png)
Works pretty great!
<details>
  <summary>The code for the dialog</summary>
  
```python
import aqt
from aqt import gui_hooks
from aqt.qt import QDialog, Qt, QDialogButtonBox, QKeySequence, QShortcut
from aqt.utils import disable_help_button, restoreGeom, saveGeom, tr

from aqt.browser import Browser
from aqt.browser.previewer import Previewer, MultiCardPreviewer, BrowserPreviewer
from aqt.utils import tooltip


class SimplePreviewer(MultiCardPreviewer):
    def __init__(self, mw, card_ids):
        super().__init__(parent=None, mw=mw, on_close=lambda: None)
        self.card_ids = card_ids
        self.current = 0
        self.last = 0

    def card(self):
        card_id = self.card_ids[self.current]
        return self.mw.col.get_card(card_id)

    def card_changed(self):
        changed = self.current != self.last
        self.last = self.current
        return changed

    def _on_prev_card(self) -> None:
        if self._should_enable_prev():
            self.current -= 1
        self.render_card()

    def _on_next_card(self) -> None:
        if self._should_enable_next():
            self.current += 1
        self.render_card()

    def _should_enable_prev(self) -> bool:
        return self.current > 0

    def _should_enable_next(self) -> bool:
        return self.current < len(self.card_ids) - 1

    def _render_scheduled(self) -> None:
        super()._render_scheduled()
        self._updateButtons()


[...]

Issue #2011, opened 2021-11-15
"Parse text using installed dictionaries" and "Parse text using MeCab" feature clarification?


kuroahna:
Under the Text Parsing section, it says
> Yomichan is able to scan the sentence surrounding a term and parse individual words of the query text on the search page. This information can be added to Anki cards to provide additional context.

And 
> Parse text using installed dictionaries
>
> Words are scanned by automatically advancing in the sentence after a matching word.

> Parse text using MeCab
>
> Requires a native component to be installed that Yomichan will connect to.
> MeCab is a third-party program which uses its own dictionaries and parsing algorithm to decompose sentences into individual words. In order for Yomichan to use it, both MeCab and a native messaging component must be installed. A setup guide can be found here.

1. Does enabling "Parse text using installed dictionaries" and/or "Parse text using Mecab" affect the parsing algorithm when scanning text? Or does this simply just affect the automatic sentence generation to anki with the `{sentence}` marker in the anki card format settings? 
2. Can both "Parse text using installed dictionaries" and "Parse text using Mecab" be enabled at the same time?
3. What does "Parse text using installed dictionaries" exactly do? The description "words are scanned by automatically advancing in the sentence after a matching word" doesn't exactly clarify it to me. Does this mean it uses entries from your installed dictionaries to find new words?
4. What is the advantages/disadvantages of using "Parse text using Mecab"? Does it provide more accurate parsing compared to Yomichan's built-in parsing algorithm when scanning text? Is it faster?

If possible it would be nice to clarify some of these questions in the main settings page itself

toasted-nutbread:
1. It affects the parsed text at the top of the standalone search page (see reference images below), but it does not affect the `{sentence}` marker. The `{sentence}` marker only uses the internal non-Mecab parser (currently).
2. Yes, if Mecab is installed. There will be a dropdown menu on the search/popup page to change the parse results.
3. It's a somewhat naive algorithm for parsing words in a sentence. It does this by finding the longest matched text using the installed dictionaries and assuming that is a word. It does not do any sort of sophisticated contextual analysis of whether word boundaries actually form grammatically correct sentences.
4. Mecab is probably more sophisticated and accurate, though I cannot speak to the exact implementation details. The internal parser will have some results that aren't always so great (see reference image 2), but in the general case it's acceptable.

See also:
https://en.wikipedia.org/wiki/MeCab

Reference image 1:
[...]

PR #2143, opened 2022-05-17
Add support for guiEditNote to view notes

toasted-nutbread:
See #2070


oakkitten:
Since Browser is the default, perhaps letting the call to `guiEditNote` fail would be better? If I chose Editor in preferences but still got Browser, I'd assume something is broken in Yomichan. Hopefully the error is not too cryptic. If it is, perhaps show a different error with something like “You chose to open Editor but you need Anki-Connect version 22.4.26.0 for that”?

Also—of course this is a matter of taste, but—I am a fan of settings that read like a sentence, so I would say

> View added notes in… {Card browser, Note editor}

or even

> Clicking on “View added note” opens… {Card browser, Note editor}

“Card browser” instead of “Browser” might be a bit less confusing I think? Also, since the user will be unfamiliar with Note editor, `settings-item-description` might say,

> Note editor is more compact than Card browser. It requires Anki-Connect 22.4.26.0.

If `guiEditNote` is allowed to fail, some kind of “it requires” would be a must. However, Anki is not aware of addon versions... Perhaps “It requires a recent version ...”? Oh well.

oakkitten:
I pulled your branch and it works! I still have some misgivings about the way it works though.

* I'm still not fan of the terminology. English is not my mother tongue so take this with a grain of salt. “View note window” reads like a verb phrase and is confusing. Especially as it is sandwiched between “Suspend new cards” and “Show card tags”. And when you finally read it the right way, what *is* the “View note window”? Is it related to what the “View added note” *icon* does? I'm not sure all users will make the connection. I have a strong preference for settings that are un-confusing as possible.

  Same goes for Editor and Browser. (Doesn't help that Anki calls it Brows*e*.) I understand that Browser can also show notes, but it's been a Card browser for a while and most people use it to look at cards so it's just the most clear way to refer to it methinks. As for Editor—I like “Note editor” slightly better than just “Editor”—it needs a tiny bit of explanation, otherwise the user would have little idea what it could be at all.

  And while I prefer settings that read like a sentence they don't have to be sentences, the important part is that it's clear and unambiguous. Or maybe just drop the “...”, this still looks okay to me:

  > Clicking on “View added note” opens {Card browser, Note editor}

  I can think of a few other clear variants but they feel way worse than this. Like, idk, “Window that is opened when clicking on ‘View added note’”? Ugh.

* Trying to open Edit dialog when it errors out opens the Browser. Since the error is shown, perhaps it's better not to open the browser? The browser obscured the error for me, and when I went back to look at it, it disappeared because I moved a mouse and that closed the popup window. Feels a bit janky... The link to the setting is awesome, though!

* The same error (again, it mentions “note window”—granted, this time it's probably be less confusing) is shown when both Edit dialog is not available and when it errors out. The dialog will raise an exception if the note doesn't exist. This is a rare situation but perhaps it would be best to show the default error in this case, since opening the browser would be useless anyway. Better yet, only show this error when API does not exist?

> AnkiConnect doesn't really have versioning, so that does make that a bit difficult. 

Some addons put their version in config.md. This can be done automatically, [like here](https://github.com/oakkitten/anki-wallpaper/blob/59eb2aa0b842403fb46d4ea3d65dff168f48dfaf/.github/workflows/main.yml#L67-L71). I think it would be nice to have this in Anki-Connect as well. (I'll be making a PR with this, as well as with running tests in a matrix and auto-bumping Anki a bit later.)

toasted-nutbread:
> “View note window” reads like a verb phrase and is confusing.

I see what you're getting at now. I was overlooking that interpretation as I was focusing on it more as it relates to the code, and in my mind the _View note_ was an adjective of _window_. Thoughts on some alternatives below.

> And when you finally read it the right way, what _is_ the “View note window”? Is it related to what the “View added note” _icon_ does? I'm not sure all users will make the connection. I have a strong preference for settings that are un-confusing as possible.

I can add a _More&hellip;_ link for this.

> it's been a Card browser for a while and most people use it to look at cards

You know, I've never really thought about this much, but the "view note" button opening the "card browser" is somewhat confusing. It is viewing all _cards_ for that note, not just the note itself. I think you're right, that should be disambiguated in the setting dropdown.

> * I can think of a few other clear variants but they feel way worse than this. Like, idk, “Window that is opened when clicking on ‘View added note’”? Ugh.

Some options I can think of which maintain brevity:

* Note viewer
* Note viewer window
* Preferred note viewer
* Preferred note viewer window
* Preferred _View Note_ window (italics to indicate quoted text)

Again, a _More&hellip;_ link may help for this.

oakkitten:
I see a lot that brevity is preferred in settings and it irks me a lot. Especially on Android—you have to jump through a few hoops to have settings with multiline titles. Most apps use single line titles, and even with the apps by Google these titles often don't fit the screen. Why do I have to rotate the screen to read a setting name?

And then you have setting descriptions, or summaries, as Android calls them. These are necessary! You tried to make the title brief, so it's now confusing. And what do you put there? Well, the same thing you would put into the title if you had more space. So now you have the same thing repeated twice. This is useless! I mean, I don't have to go far for the examples...

> Result grouping mode
Change how related results are grouped.

> Frequency sorting dictionary
Sort results using a frequency dictionary.

> Maximum number of results
Adjust the maximum number of results shown for lookups.

*throws table*

Sorry, rant over. 😅 I guess this is a boulder that I would be willing to ~~die~~ get slightly sick on. But yeah, I prefer settings that don't require descriptions, and if there is a description, it should not be repeating what the title says.

Said that, I can live with “Note viewer”... except the choices would be “(Card) browser“ and... “(Note) editor”? Note viewer: Note editor. I don't quite like this but I find it better than the other variants.

toasted-nutbread:
IMO there's rationale for having a setting title and a summary, since it makes it easier to refer to a specific setting from other contexts. Example: it's easier to say "Check the <em>Note viewer</em> setting." than "Check the <em>Clicking on “View added note” opens</em> setting." So my preference is for stylistic consistency, and generally having the settings titles read as fully descriptive sentences is not the current style.

Point taken, arguably some of the descriptions could use some work or removal, but that's beyond the scope of this PR. In general, I don't believe that it's possible to reduce all (title, summary) pairs to just a single (title-summary) for all settings.

---

> Said that, I can live with “Note viewer”... except the choices would be “(Card) browser“ and... “(Note) editor”? Note viewer: Note editor. I don't quite like this but I find it better than the other variants.

43b15480e7b2944c63da56dd97090a3dcf43abf7 updates the options to "Card browser" and "Note editor".

---

> Trying to open Edit dialog when it errors out opens the Browser. Since the error is shown, perhaps it's better not to open the browser? The browser obscured the error for me, and when I went back to look at it, it disappeared because I moved a mouse and that closed the popup window. Feels a bit janky... The link to the setting is awesome, though!

> The same error (again, it mentions “note window”—granted, this time it's probably be less confusing) is shown when both Edit dialog is not available and when it errors out. The dialog will raise an exception if the note doesn't exist. This is a rare situation but perhaps it would be best to show the default error in this case, since opening the browser would be useless anyway. Better yet, only show this error when API does not exist?

This is updated, now just shows an error without falling back. Also fixed up some error handling to only show the special message when the API is not available.

oakkitten:
> updates the options to "Card browser" and "Note editor".

I only wanted to say that “Note viewer: Note editor” together would sound a bit awkward 🙄

> This is updated, now just shows an error without falling back.

Nice! I would also change the message to reflect the exact error. Something like... 

> Note editor could not be shown as this version of Anki-Connect doesn't support it. You might need to update Anki-Connect in order to use Note editor. Meanwhile, you can keep using Card browser to view added notes by setting _Note viewer_ to _Card browser_.

By the way, you can weave stuff like “Clicking on ‘View added note’ opens” into this if you are careful; like, 

> Meanwhile, you can set _Clicking on ‘View added note’ opens_ to _Card browser_ which is supported by all Anki-Connect versions.

It is a little awkward grammar-wise, yes, but in my opinion the meaning is still very clear; besides, users spend much more time reading settings by themselves than they get referred to by others don't they? 

But yeah this would not fit current style. It just styles can be [a bit hard to change](https://github.com/ankidroid/Anki-Android/issues/8710) 😪

Oh well I'll be shutting up 😅 This is a matter of opinion where I've always been a bit of a minority, and not that big an issue, anyway. This is pretty neat either way. And can always be changed one way or another.

toasted-nutbread:
Updates made:

* Updated setting title (I included "window" since I think it adds a bit more specificity)
* Added a _More&hellip;_ description with test buttons.
* Updated link colors now applies to the settings link for easier readability. (#2151)

Preview of UI and errors:
image

ext/settings.html
```diff
@@ -1764,6 +1764,39 @@ <h1>Yomichan Settings</h1>
...
+                <div class="settings-item-right">
+                    <select data-setting="anki.noteGuiMode">
+                        <option value="browse">Card browser</option>
+                        <option value="edit">Note editor</option>
+                    </select>
+                </div>
+            </div>
+            <div class="settings-item-children more" hidden>
+                <p>
+                    AnkiConnect releases after around 2022-05-20 support a new note editor window
```
toasted-nutbread, line 1785: * [ ] TODO - update this date once this feature is actually deployed.

toasted-nutbread:
> Nice! I would also change the message to reflect the exact error. Something like...

I went with a short error in the popup since the default size of the popup is very small. A longer description is found after expanding _More&hellip;_.

Overall, I appreciate the feedback on this. We don't often get a lot of user feedback on whether or not the settings are adequately clear, so I think it's helpful. Some of the "power user" settings have also been confusing at times also (#2011).

ext/settings.html
toasted-nutbread: Updated date based on merge date of https://github.com/FooSoft/anki-connect/pull/323. Note that this is not necessarily the release date of an AnkiConnect update.

PR #2178, opened 2022-06-18
CSS zoom handling

### LINKED ISSUE #2176: Does not work on a website with Chrome
**Description**
The popup does not show up when trying to scan words on this page: https://ieagent.jp/blog/eria/osusumeeki-11901
This issue only happens on Chrome. It works as expected on Firefox.

**Browser version**
Chrome 102.0.5005.115

**Yomichan version**
22.4.4.0

**Exported settings file**
[yomichan-settings-2022-06-18-06-02-22.txt](https://github.com/FooSoft/yomichan/files/8933030/yomichan-settings-2022-06-18-06-02-22.txt)

toasted-nutbread:
This change handles the non-standard CSS `zoom` property on Chromium-based browsers, which causes scanning to break.

Will resolve #2176.

### DIFF
--- a/ext/js/dom/document-util.js
+++ b/ext/js/dom/document-util.js
@@ -24,6 +24,7 @@
 class DocumentUtil {
     constructor() {
         this._transparentColorPattern = /rgba\s*\([^)]*,\s*0(?:\.0+)?\s*\)/;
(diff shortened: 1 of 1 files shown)


toasted-nutbread:
For simplicity, this will not be put behind an option.

PR #2186, opened 2022-06-24
Do not change selection if user selected text manually

### LINKED ISSUE #2071: [Feature Request] Don't select matched text if user selected some text manually
Unless I am mistaken, when the option “Select matched text” is on, the only way to copy some text from the page without turning off Yomichan is to select text and press `Ctrl-C` while still holding the left mouse button. It would be nice to be also able to select text after releasing the mouse button—and also via the context menu.

I imagine that it works like this; Yomichan selects matched text only when user has no manually selected text on the page, and user must dismiss their selection by clicking on the page to make Yomichan resume selecting matching text.

oakkitten:
While this works for me, please note that I have no idea what I am doing.
image
Particularly, I am not sure about the `setTimeout()` bit, and about all the possible edge cases. This also could be slightly optimized by only hooking `selectionchange` when select text is on, but the extra complexity is probably not worth it.

Closes #2071.

### DIFF
--- a/ext/js/language/text-scanner.js
+++ b/ext/js/language/text-scanner.js
@@ -759,7 +759,7 @@ class TextScanner extends EventDispatcher {
             eventListenerInfos.push(...this._getMouseClickOnlyEventListeners2(capture));
         }
 
-        eventListenerInfos.push(this._getManualSelectionChangeByUserListener(capture));
+        eventListenerInfos.push(this._getManualSelectionChangeByUserListener());
 
         for (const args of eventListenerInfos) {
             this._eventListeners.addEventListener(...args);
@@ -822,13 +822,13 @@ class TextScanner extends EventDispatcher {
         return entries;
     }
 
-    _getManualSelectionChangeByUserListener(capture) {
+    _getManualSelectionChangeByUserListener() {
         const callback = () => {
             if (this._yomichanIsChangingTextSelectionNow) { return; }
             this._userHasNotSelectedAnythingManually = window.getSelection().isCollapsed;
         };
 
-        return [document, 'selectionchange', callback, capture];
+        return [document, 'selectionchange', callback];
     }
 
     _getTouch(touchList, identifier) {
--- a/ext/js/language/text-scanner.js
+++ b/ext/js/language/text-scanner.js
@@ -87,6 +87,9 @@ class TextScanner extends EventDispatcher {
         this._pointerIdTypeMap = new Map();
 
         this._canClearSelection = true;
+
+        this._yomichanIsChangingTextSelectionNow = false;
+        this._userHasNotSelectedAnythingManually = true;
     }
 
     get canClearSelection() {
@@ -259,8 +262,10 @@ class TextScanner extends EventDispatcher {
 
     setCurrentTextSource(textSource) {
         this._textSourceCurrent = textSource;
-        if (this._selectText) {
+        if (this._selectText && this._userHasNotSelectedAnythingManually) {
+            this._yomichanIsChangingTextSelectionNow = true;
             this._textSourceCurrent.select();
+            setTimeout(() => this._yomichanIsChangingTextSelectionNow = false, 0);
             this._textSourceCurrentSelected = true;
         } else {
             this._textSourceCurrentSelected = false;
@@ -754,6 +759,8 @@ class TextScanner extends EventDispatcher {
             eventListenerInfos.push(...this._getMouseClickOnlyEventListeners2(capture));
         }
 
+        eventListenerInfos.push(this._getManualSelectionChangeByUserListener(capture));
+
         for (const args of eventListenerInfos) {
             this._eventListeners.addEventListener(...args);
         }
@@ -813,6 +820,15 @@ class TextScanner extends EventDispatcher {
             }
         }
         return entries;
+    }
+
+    _getManualSelectionChangeByUserListener(capture) {
+        const callback = () => {
+            if (this._yomichanIsChangingTextSelectionNow) { return; }
+            this._userHasNotSelectedAnythingManually = window.getSelection().isCollapsed;
+        };
+
+        return [document, 'selectionchange', callback, capture];
     }
 
     _getTouch(touchList, identifier) {
--- a/ext/js/language/text-scanner.js
+++ b/ext/js/language/text-scanner.js
@@ -144,6 +144,7 @@ class TextScanner extends EventDispatcher {
 
         if (value) {
             this._hookEvents();
+            this._userHasNotSelectedAnythingManually = window.getSelection().isCollapsed;
         }
     }
 
--- a/ext/js/language/text-scanner.js
+++ b/ext/js/language/text-scanner.js
@@ -377,6 +377,11 @@ class TextScanner extends EventDispatcher {
         this._preventNextClickScan = true;
     }
 
+    _onSelectionChangeCheckUserSelection() {
+        if (this._yomichanIsChangingTextSelectionNow) { return; }
+        this._userHasNotSelectedAnythingManually = window.getSelection().isCollapsed;
+    }
+
     _onSearchClickMouseDown(e) {
         if (e.button !== 0) { return; }
         this._resetPreventNextClickScan();
@@ -760,7 +765,7 @@ class TextScanner extends EventDispatcher {
             eventListenerInfos.push(...this._getMouseClickOnlyEventListeners2(capture));
         }
 
-        eventListenerInfos.push(this._getManualSelectionChangeByUserListener());
+        eventListenerInfos.push(this._getSelectionChangeCheckUserSelectionListener());
 
         for (const args of eventListenerInfos) {
             this._eventListeners.addEventListener(...args);
@@ -823,13 +828,8 @@ class TextScanner extends EventDispatcher {
         return entries;
     }
 
-    _getManualSelectionChangeByUserListener() {
-        const callback = () => {
-            if (this._yomichanIsChangingTextSelectionNow) { return; }
-            this._userHasNotSelectedAnythingManually = window.getSelection().isCollapsed;
-        };
-
-        return [document, 'selectionchange', callback];
+    _getSelectionChangeCheckUserSelectionListener() {
+        return [document, 'selectionchange', this._onSelectionChangeCheckUserSelection.bind(this)];
     }
 
     _getTouch(touchList, identifier) {


ext/js/language/text-scanner.js
```diff
@@ -87,6 +87,9 @@ class TextScanner extends EventDispatcher {
         this._pointerIdTypeMap = new Map();
 
         this._canClearSelection = true;
+
+        this._yomichanIsChangingTextSelectionNow = false;
```
toasted-nutbread, line 91: Simplify names; something like `_ignoreSelectionChange` and `_hasManualUserSelection`

ext/js/language/text-scanner.js
```diff
@@ -815,6 +822,15 @@ class TextScanner extends EventDispatcher {
         return entries;
     }
 
+    _getManualSelectionChangeByUserListener(capture) {
```
toasted-nutbread, line 825: Define callback similar to this:
https://github.com/FooSoft/yomichan/blob/4e7762edacb6cfdb49207bbf43102a04706c78a1/ext/js/language/text-scanner.js#L369-L372
https://github.com/FooSoft/yomichan/blob/4e7762edacb6cfdb49207bbf43102a04706c78a1/ext/js/language/text-scanner.js#L806-L808

Probably just put it after that one, can name it just `_onSelectionChange2` for simplicity. You can also omit the `capture` parameter, as it isn't applicable to `selectionchange`.

ext/js/language/text-scanner.js
```diff
@@ -259,8 +262,10 @@ class TextScanner extends EventDispatcher {
 
     setCurrentTextSource(textSource) {
         this._textSourceCurrent = textSource;
-        if (this._selectText) {
+        if (this._selectText && this._userHasNotSelectedAnythingManually) {
+            this._yomichanIsChangingTextSelectionNow = true;
             this._textSourceCurrent.select();
+            setTimeout(() => this._yomichanIsChangingTextSelectionNow = false, 0);
```
toasted-nutbread, line 268: Probably want to have this callback defined in the constructor so it doesn't have to create a new closure every time it's invoked.

ext/js/language/text-scanner.js
```diff
@@ -87,6 +87,9 @@ class TextScanner extends EventDispatcher {
         this._pointerIdTypeMap = new Map();
 
         this._canClearSelection = true;
+
+        this._yomichanIsChangingTextSelectionNow = false;
+        this._userHasNotSelectedAnythingManually = true;
```
toasted-nutbread, line 92: Reset the state of these variables in `setEnabled`; you may also want to assign `_userHasNotSelectedAnythingManually` based on the current selection in there if `value` is true. This would handle the following edge case:

1. User has Yomichan disabled
2. Opens webpage
3. Selects text
4. Enables Yomichan via the browser action popup
5. Scans text
6. User selection is lost

toasted-nutbread:
> This also could be slightly optimized by only hooking `selectionchange` when select text is on, but the extra complexity is probably not worth it.

We will probably want the extra functionality to run only if `_selectText` is true, as I may change that setting later on to support both use cases.

ext/js/language/text-scanner.js
oakkitten: I can do that, but I was assuming that JS VM should be sufficiently optimizing this sort of thing. It seems that closures are extremely common in JS. Do you think that trying to manually optimize this is worthwhile?

ext/js/language/text-scanner.js
oakkitten: Aha! Will do.

oakkitten:
> We will probably want the extra functionality to run only if `_selectText` is true, as I may change that setting later on to support both use cases.

Not sure I should do anything about it? What I meant was, to only add event listener for selection change if `_selectText` is true. Given how it can seemingly change while the scanner is active, this will lead to a bit of spaghetti code. I can do that, it's just that optimizations like that tend to have a negative effect later on.

ext/js/language/text-scanner.js
toasted-nutbread: First part:
* My rationale was that `_yomichan` is somewhat redundant since this class belongs to Yomichan, and I've tried to avoid redundant names on things (though there are [some places](https://github.com/FooSoft/yomichan/blob/4e7762edacb6cfdb49207bbf43102a04706c78a1/ext/js/pages/settings/anki-controller.js#L26-L48) I've not addressed).
* Similarly, the `Now` part is redundant since when you reference the value, it will always be "now".
* `_ignoreSelectionChange` may not be the best name, you're right, `_isChangingTextSelection` would probably have been a better suggestion, but my rationale is the same.
* Ultimately this is a semantics thing where your naming convention is different from how I usually do it, but probably not worth me getting hung up on.

Second part:
> Also, this inverts the logic. If a variable is only ever used with negation, it should not exist, IMO. 

Fair enough, as this would incur an additional negation unary on line 829. Aside from that, my rationale was that I have generally avoided storing negation variables, such as having "not" in the name, or preferring "enabled" rather than "disabled", for reasons of consistency. There is no difference in bytecode for either case.

ext/js/language/text-scanner.js
toasted-nutbread: You can choose a more meaningful name than `_onSelectionChange2`, it was for example and to show that I wouldn't mind if you did name it that. Rationale behind this request is for consistency with the rest of the file (and application in general).

> easier for the JS VM to optimize

I don't think it makes any difference realistically, but a while back I decided on using `.bind` for event handlers rather than arrow functions, so this kind of falls under that decision.

ext/js/language/text-scanner.js
toasted-nutbread: This is something I've done for a few high-traffic code paths, but doing some testing with a similar case doesn't show much of a difference either way. So at worst it would not affect anything, but overall it probably doesn't matter.

toasted-nutbread:
> Not sure I should do anything about it? What I meant was, to only add event listener for selection change if `_selectText` is true. Given how it can seemingly change while the scanner is active, this will lead to a bit of spaghetti code. I can do that, it's just that optimizations like that tend to have a negative effect later on.

True, for some reason I thought that this class updated the events listeners each time the settings were updated, but looks like I misremembered. I wouldn't worry about this then.

toasted-nutbread:
The only other edges case I can think of that might conflict with this and/or potentially confuse the user is if the website has an invisible selection or some external code selects text, which isn't truly a user selection but would be detected as such. I don't think these can be worked around if this behaviour is always active though.

Simple example case, but there are many others:
```html
<span id="select"> </span>
<script>
window.addEventListener('load', () => {
    setTimeout(() => {
        const range = document.createRange();
        range.setStart(document.querySelector('#select'), 0);
        range.setEnd(document.querySelector('#select'), 1);
        const selection = window.getSelection();
        console.log(selection.isCollapsed, {s: selection.toString()});
        selection.removeAllRanges();
        selection.addRange(range);
        console.log(selection.isCollapsed, {s: selection.toString()});
    }, 1000);
});
</script>
```
Logs:
```
true {s: ''}
false {s: ''}
```
Result is that no selection is visible and Yomichan won't update the selection.

ext/js/language/text-scanner.js
oakkitten: I used `yomichan` to signify that the add-on is changing the selection, as apposed to the user, or something else. I like it better than `addon` or `we`. With `selectionIsBeingChangedNow` you can't be sure, and `isChangingTextSelection` strongly suggests that we are *not* the actor, as stuff like `isConnected()` or `isBlankOrNull()` are often used to check the state of a thing that you do not directly set. Or at least I would find that confusing.

In a similar fashion, `yomichanIsChangingTextSelection` without `Now` could be read as a setting, as in the option to change selection is on. I think suffixes such as `Now` or `Flag` (I use this to signify that the thing is reset by the reader) are somewhat useful. (Also you would often use phrases ending with -ing for stuff like `with process_running() as pid: ...`, however, this can be seen as a counterexample maybe?)

It's not that I am a fan of long variable names, but in some cases I like them. Here, they don't hurt so to say. The statements all fit one line, and they only appear in a few places. If these were used extensively close together, it would make code blocks harder to read, but far apart it only helps, I think.

> I have generally avoided storing negation variables

I also avoid things like that! I guess this is a rare exception.

ext/js/language/text-scanner.js
oakkitten: You know, I went to implement that change and I tried to pick a good name and the best thing I could come up with was `onSelectionChangeListenerThatSetsWhetherUserHasSelectedAnythingManually` and this would be ridiculous. The thing about event listeners is that I am not sure if they can ever have a specific name. A selection change listener is a selection change listener! I guess this is one of the reasons they rarely get identifiers, and if they do, they are rarely descriptive of the action.

I guess if you think that consistency is key, I can rename this to `_onSelectionChange2`.

P.S. This shouldn't matter much but apparently [binds can even be slower than closures](https://stackoverflow.com/questions/17638305/why-is-bind-slower-than-a-closure)?

ext/js/language/text-scanner.js
oakkitten: When I learned of the existence of profiling flame charts, I stopped trying to optimize minor things like that. The bottlenecks were never where I thought they would be!

oakkitten:
Even if the user manages to stumble upon such a page, simply clicking in a page should fix that, shouldn't it? Can't say for sure, but I think that I'd click on a page if I saw no selections by Yomichan, if only because I like clicking on things that don't work 😅.

Although, it might be worth checking with a few websites that hijack selections to insert links to their pages. Nasty nasty websites!

ext/js/language/text-scanner.js
toasted-nutbread: > I guess if you think that consistency is key, I can rename this to `_onSelectionChange2`.

Yeah, the rationale I've used for putting event handlers directly on the body is because:

1. Sometimes they are non-trivial. I prefer to not have 10s of lines of callback code embedded into functions.
2. Due to 1, for consistency, even short callbacks are set up the same way.
3. Makes the event handlers a bit easier to find all of them. All of the `_on*` functions are generally located in the same area.

And as far as naming goes, I generally follow just `_on<Object (optional)><EventName>`, without trying to include in the name what the code is doing or why. So following that pattern, adding a `2` is the only meaningful thing. I suppose the handlers could also be merged into a single handler which is always attached.

> P.S. This shouldn't matter much but apparently [binds can even be slower than closures](https://stackoverflow.com/questions/17638305/why-is-bind-slower-than-a-closure)?

The question is from 8 years ago so things have potentially changed; I've seen this be the case for a few old topics I've seen on there. It also seems to be referencing that `bind` itself can be slower than assigning a closure. Invocation of the function later on is basically the same.

This was also a design decision that was made a while back in #359.

toasted-nutbread:
> Even if the user manages to stumble upon such a page, simply clicking in a page should fix that, shouldn't it? Can't say for sure, but I think that I'd click on a page if I saw no selections by Yomichan, if only because I like clicking on things that don't work 😅.

Correct, the user can deselect to fix the issue. It's not something I think should be worried about, just an interesting edge case.

> Nasty nasty websites!

Very true

ext/js/language/text-scanner.js
oakkitten: Perhaps naming a selection change listener a selection change listener is fine if you only have one of these. But there was one already! Eugh. Well, I looked around and found the method `_onTouchMovePreventScroll`, which seems to be a friendly cousin of `_onTouchMove`. Now this is not too bad! `TouchMove` describes when it's called, and `PreventScroll` describes what it does. *And* if you look for `_onTouchMove` you'll find both.

I tried phrasing it this way and came up with `_onSelectionChangeCheckUserSelection`? Yes it is still akward, but at least
  * you won't confuse it with `_onSelectionChange`
  * it still fits all lines
  * consistent with the rest of the code
  * you'll find it if you look for `_onSelectionChange`!

How about this? If you don't like it, `_onSelectionChange2` it is.

oakkitten:
Fwiw I managed to find one website that does copy hijacking and everything worked well with it.

ext/js/language/text-scanner.js
toasted-nutbread: Looks good to me

Issue #2093, opened 2022-03-25
Google Docs HTML Render appears broken


Mageinta:
It's March 25th rn and it was working fine, I think yesterday but March 23rd for sure. I've done some poking around and idk, seems like Google may have decided to change some render rules and stuff. Opening in console an error appears to be thrown where it wants a "trusted" data structure. Anyway, I'm sure you knew it but I wanted to put it up here anyway. So now you cannot Yomichan scan google docs. Really tragic, gonna set my project back until it's fixed. Love the work, cheers.

toasted-nutbread:
Can you screenshot the error you're seeing?

Mageinta:
![image](https://user-images.githubusercontent.com/6201560/160214744-757d78ac-edac-4167-a023-6d9cd042f5c8.png)

toasted-nutbread:
What web browser are you using? Do you have any addons?

Mageinta:
I use Chrome and I've got a few addons. 
![image](https://user-images.githubusercontent.com/6201560/160216140-f7d5d31f-248a-4d99-a31b-7b07e1a158dd.png)
![image](https://user-images.githubusercontent.com/6201560/160216169-2ff14ac0-9d4f-4bfc-846e-4529e2615827.png)
The reason I started poking around is because Dark Mode also ceased working and removing it didn't fix the issue

toasted-nutbread:
The error is probably unrelated to why GDocs is no longer working, as I don't get the error in my testing. However, GDocs still does not work, so I assume Google is removing the HTML rendering fallback.

See also: https://github.com/melink14/rikaikun/issues/881

toasted-nutbread:
@FooSoft Maybe this is something worth looking into, just to see what they say, but Google provides a form that can be filled out to indicate that the canvas changes negatively affect extensions.

[https://workspaceupdates.googleblog.com/2021/05/Google-Docs-Canvas-Based-Rendering-Update.html](https://workspaceupdates.googleblog.com/2021/05/Google-Docs-Canvas-Based-Rendering-Update.html#:~:text=submit%20this%20form%20to%20provide%20feedback%20and%20notify%20our%20team)

> If your company has developed a private Chrome Extension that you believe will be impacted and you are unable to migrate to the Google Workspace Add-ons framework, you can submit this form to provide feedback and notify our team.

Granted this document is pretty old, so who knows if they're still monitoring it.

[...]

Issue #305, opened 2019-12-09
Sometimes scanning doesn't work


seanblue:
This error happened on Bunpro (which happens occasionally, though it usually works), but I've also seen scanning randomly fail on other websites over many months. I wonder if there's something about how their JavaScript is written that causes a conflict or race condition with the extension.

Here is the error for this occurrence:
```
popup.js:291 Uncaught (in promise) TypeError: Cannot read property 'postMessage' of null
    at Popup.invokeApi (popup.js:291)
    at Popup.termsShow (popup.js:276)
invokeApi @ popup.js:291
termsShow @ popup.js:276
async function (async)
termsShow @ popup.js:275
searchTerms @ frontend.js:360
async function (async)
searchTerms @ frontend.js:351
searchAt @ frontend.js:315
async function (async)
searchAt @ frontend.js:301
search @ frontend.js:94
onMouseMove @ frontend.js:103
```

toasted-nutbread:
It looks like the error is coming from `contentWindow` being `null`, which appears to be possible in the [HTML spec](https://html.spec.whatwg.org/multipage/iframe-embed-object.html#dom-iframe-contentwindow). It may be a timing issue where the frame isn't initialized, but that's not necessarily the cause.

A few questions:
* What browser are you seeing the error in?
* Does this error happen only once? Or do repeated scans also trigger a new error?
* Is it consistently reproducible if the page is refreshed, or is it sporadic?
* Is there any example content on the site which triggers the error? Since it appears to be a paid service, I don't have access to anything immediate to test with.

seanblue:
> What browser are you seeing the error in?

Google Chrome on Windows 10.

> Does this error happen only once? Or do repeated scans also trigger a new error?

Whenever this happens it's as if the page gets into a bad state. Once it happens once it happens 100% of the time until I reload the page.

> Is it consistently reproducible if the page is refreshed, or is it sporadic?

It's sporadic. Some websites never exhibit this problem, while others do on occasion. This is why I believe it's some combination of website and extension code causing a race condition.

> Is there any example content on the site which triggers the error?

There was nothing special about the page it happened on. All I can say is that the content was **not** in an iframe. I'll be sure to comment again when I see the issue happen on a free/open website. For now I just wanted to make you aware of the general problem in case the cause was obvious. This isn't a new issue by the way. This has been happening on and off for a long time.

toasted-nutbread:
[...]

PR #2235, opened 2022-09-24
Google Docs accessibility update

### LINKED ISSUE #2231: Is there a way to make Yomichan work on Google Docs again?
Since Google Docs changed from HTML rendering to canvas rendering, Yomichan no longer works and I can't scan Japanese. Is there some workaround I can do, or is there any way the devs can fix it?

toasted-nutbread:
This change updates the implementation of the Google Docs accessibility scripts and restores some support for scanning on Google Docs pages. It currently lacks the ability to select text, since there is no real text, but it otherwise functions similar to how scanning input elements works: by creating a temporary "<span title="ඞ">imposter</span>" element and then scanning it.

Google exposes some structural text information in a hidden `<svg>` element if `window._docs_annotate_canvas_by_ext` is assigned to a whitelisted extension ID. This change registers an extension to `DocumentUtil` which will handle the scanning on Google Docs webpages.

The extension ID whitelist can be found by viewing the source of the main Google Docs script and searching for an ID which is known to be on the list, at least at the time of writing this. There is a big array of extension IDs, and Yomichan's ID `ogmnaimimemjmbakcfefmnahgdfhfami` is on that list, which is nice. (Maybe filling out that form mentioned in [this comment](https://github.com/FooSoft/yomichan/issues/2093#issuecomment-1079732325) actually did something!)

This implementation is still subject to the whims of Google's changes, which are unpredictable and undocumented.

Resolves #2231.
Related: #2093, https://github.com/FooSoft/yomichan/issues/305#issuecomment-1249550618.

### DIFF
--- a/ext/js/accessibility/google-docs-util.js
+++ b/ext/js/accessibility/google-docs-util.js
@@ -40,8 +40,8 @@ class GoogleDocsUtil {
         styleNode.disabled = true;
         for (const element of elements) {
             if (!element.matches(selector)) { continue; }
-            const {ariaLabel} = element;
-            if (typeof ariaLabel !== 'string') { continue; }
+            const ariaLabel = element.getAttribute('aria-label');
+            if (typeof ariaLabel !== 'string' || ariaLabel.length === 0) { continue; }
             return this._createRange(element, ariaLabel, x, y, normalizeCssZoom);
         }
         return null;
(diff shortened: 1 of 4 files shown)


gadicc:
Massive thanks for this.  All my Japanese learning notes are in Google Docs...  I've been getting by with 10ten (Rikaichamp) but have been missing Yomichan so much!

PR #2278, opened 2022-12-08
Add color option for structured content

KiwiCracker:
### Add support for the [color property](https://developer.mozilla.org/en-US/docs/Web/CSS/color)
It's nice to have some more variation aviable when it comes to designing dictionary-entries.
A few basic color names can be used, or any valid RGBA value.


stephenmk:
Since yomichan has light and dark modes, these styles might also make text unreadable in one mode or the other. They could also potentially cause some accessibility issues (e.g. for users with color blindness). Just a couple things to consider.

